PR #31, opened 2020-07-10
Updated WriteByte to new ISComm interface

### LINKED ISSUE #29: GpsParser not updated for new ISComm interface
In `cInertialSenseParser`, `WriteByte` was not updated following the new structure of `is_comm_instance_t`

kiranshila:
Fixes #29


waltjohnson:
Kiran.  It looks like GpsParser.cpp was supposed to be removed from the
SDK.  Where does it get used?  After deleting it all of the example
projects still compile.

>

waltjohnson:
This is no longer necessary as GpsParser.cpp and GpsParser.h are not needed in the SDK.

PR #46, opened 2020-10-27
SN-80 Added gyro PSD plot to LogInspector

waltjohnson:


waltjohnson:
@aabogdanov Notice in their plots their units are dB dps/rtHz verses Hz.  What units are most conventional here?

waltjohnson:
This is the PR for SN-80.

siddirp (approved):

PR #50, opened 2020-11-27
add namespaced target name for the library

ycdtosa:
This makes much easier integrating this project as a subdirectory on other cmake projects.

Just add the dependency:

<pre>
target_link_libraries( yourproject
   PRIVATE
      InertialSense::libinertialsense
      Boost::date_time
)
</pre>

other changes using the target_xxxx cmake functions would be welcome, but this is the minimal change required

feel free to use a alias name, but it needs to be "namespaced"


ycdtosa:
humm, and this would be most welcome on the release branch.

steven-fusco-inertialsense (approved):

PR #52, opened 2020-12-15
add partial support for Rad Studio C++ formerly known as C++ Builder

ycdtosa:
Hi.

I am adding support for the uINS device in an application. The application is running on linux ( gcc and clang ) and 
windows ( RadStudio 10.3 ).  

For the linux application I use the SDK as an external library and everything is fine. 

Meanwhile, on windows I am just using some files from the SDK ( not the full SDK ) until I can work on getting the SDK
compiling in RadStudio.

The tiny changes I propose here are almost enough to have the code ( the .c files ) working under Rad Studio 10.3
I would be nice to have them merged into the official SDK.

### DIFF
--- a/src/ISConstants.h
+++ b/src/ISConstants.h
@@ -195,7 +195,9 @@ extern void vPortFree(void* pv);
 #if PLATFORM_IS_WINDOWS
 
 #include <direct.h>
+#if ! defined ( __BORLANDC__ )
 #include <sys/utime.h>
+#endif
 #define _MKDIR(dir) _mkdir(dir)
 #define _RMDIR(dir) _rmdir(dir)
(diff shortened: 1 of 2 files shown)


ycdtosa:
the "almost enough" in the previous comment is there because even if I have it working, I have tweaked the definition of the STATIC_ASSERT macro and I am not yet comfortable with the fix.

Hopefully I will find a solution that can be committed in the next few days.

ycdtosa:
Finally I did not need any more changes.
Those were enough build our RadStudio C++ solution

waltjohnson:
Thanks for the input, @ycdtosa.  I'm going to close this PR if it is no longer needed.  Let us know if you have any other suggestions or needs.

PR #54, opened 2021-02-02
Feature/luna 775 900 m hz remote kill

waltjohnson:
Refactored SDK EVB-2 project for Luna remote kill support.


steven-fusco-inertialsense:
I just noticed we don't have this wired up into Travis for builds.  That seems like a good idea.

steven-fusco-inertialsense:
I'm guessing this one goes hand in hand with https://github.com/inertialsense/luna-vehicle/pull/2860 ?

PR #63, opened 2021-02-20
Added ix prefix to common variables

waltjohnson:
Renamed common variable type names to prevent customers from having conflicts when using our SDK.


src/ISEarth.h
```diff
@@ -88,8 +88,8 @@ void llaDeg2ned_d(double llaRef[3], double lla[3], Vector3_t result);
  *  lla[1] = longitude (rad)
  *  lla[2] = msl altitude (m)
  */
-void ned2lla( Vector3_t ned, Vector3_t llaRef, Vector3_t result );
-void ned2lla_d( Vector3_t ned, double llaRef[3], double result[3] );     // double precision
+void ned2lla( ixVector3 ned, ixVector3 llaRef, ixVector3 result );
```
aabogdanov, line 91: We replaced both Vector3_t and Vector3 with the same ixVector3. Are they the same? Just making sure...

aabogdanov (approved):

src/ISEarth.h
waltjohnson: Yes, ixVector3 and isVector3_t are both float.  "_t" was intended as an option that we might switch from float to double, but never made use of that feature.  So this is fine.

PR #77, opened 2021-03-17
Python cleanup

aabogdanov:
I added a few functions to pylib.pose that were defined elsewhere and left a couple of them (conversion between LLA and NED) commented out because:
1. We are not using them
2. I don't want to lose them because they provide more accurate conversion than LLA-NED functions currently defined in pylib.pose


waltjohnson (approved):

steven-fusco-inertialsense (approved):
Love those docs!

PR #85, opened 2021-04-22
Add encoder averging when count pulses are faster than wheel encoder …

waltjohnson:
…sample period


EVB-2/IS_EVB-2/src/wheel_encoder.cpp
```diff
@@ -53,14 +45,15 @@ void step_wheel_encoder(is_comm_instance_t &comm)
 	static int encoderSendTimeMs=0;
 	static wheel_encoder_t wheelEncoderLast = {0};
 
-	if(abs(g_comm_time_ms-encoderSendTimeMs)>=20)
+	if(++encoderSendTimeMs >= 20)
```
steven-fusco-inertialsense, line 48: Can we move the ++encoderSendTimeMs to the line above?  It's easy to miss, but it's important particularly because it's static.  If it was on it's own line it would be more obvious.

steven-fusco-inertialsense (approved):
I find this much easier to read now with the new capture_t structure.

PR #86, opened 2021-04-26
Feature/cltool edit did fields

waltjohnson:
Add ability to write as well as read from any DID in CLTool.


steven-fusco-inertialsense (approved):
Looks like this just reorders the code, no logic change?

PR #88, opened 2021-04-27
Luna 1215 add throttle scalar

waltjohnson:
Add support for throttle scalar.  Fix nonblockingkeyboard bug in ISDisplay which prevented keyboard inputs in Luna.


steven-fusco-inertialsense (approved):
Looks like it's always true on Linux and always false on Windows?  I don't understand why we would need to do this, but I'm approving because clearly it fixes something for you

PR #99, opened 2021-06-02
Move luna_data_sets.h into inertial-sense-sdk/src

waltjohnson:
We now include luna_data_sets.h in the inertial-sense-sdk.  It gets include automatically with Luna and we can use a INCLUDE_LUNA_DATA_SETS macro to enable it in the cltool and EvalTool (for Luna development / debug).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,4 +17,3 @@ Debug/
 *.o
 IS_logs/
 *.user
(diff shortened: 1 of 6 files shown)


steven-fusco-inertialsense (approved):
LGTM

PR #102, opened 2021-06-03
SN-453 Fixed bootloader non-standard baudrate issue

waltjohnson:
Previously, enabling the bootloader would only work for supported bootloader baudrates (921600, 230400, 115200, 2000000) but now works for all supported IS baudrates.


rudomorg (approved):
I was able to update EVB2 and uINS at 3M baud. looks good.

PR #118, opened 2021-07-01
Path bug

siddirp:


waltjohnson (approved):
Good catch!  Sorry, this bug was my fault.  It got introduced when I refactored the parameters to have spaces, removing the equals sign.  The reason I made those changes was so that tab completion doesn't work if you don't have a space before the path.

waltjohnson:
@siddirp There was another place that has the same problem with the path in the CLTool log replay.

PR #126, opened 2021-07-12
Luna-1471 Comms Example to print Dev Info

siddirp:
This is the alternate method I mentioned on Slack this morning


waltjohnson:
@siddirp   I'm not sure I want to add more stuff the Communications Example Project.  Most of our customers need brain dead simple projects so that the essential part of the project is easier to find.  What do you think?

siddirp:
> @siddirp I'm not sure I want to add more stuff the Communications Example Project. Most of our customers need brain dead simple projects so that the essential part of the project is easier to find. What do you think?

@waltjohnson That's fine. I just threw this together as an example of a brain dead app to get health info aside from the CLTool reporting. :) 75% of it could be removed. Anyway, I'm fine with this going away.

PR #130, opened 2021-07-24
Sn 541

siddirp:


python/pylib/data_sets.py
```diff
@@ -90,7 +90,7 @@
 DID_DUAL_IMU_RAW_MAG            = 84
 DID_DUAL_IMU_MAG                = 85
 DID_PREINTEGRATED_IMU_MAG       = 86
-DID_WHEEL_CONFIG                = 87
+DID_GROUND_VEHICLE                = 87
```
aabogdanov, line 93: Is wheel_config now a part of ground_vehicle?

src/data_sets.h
```diff
@@ -1985,12 +2008,24 @@ typedef struct PACKED
 
 } wheel_config_t;
```
aabogdanov, line 2009: If ground_vehicle_t replaced wheel_config_t, do we still need wheel_config_t? Does it have any use/purpose anywhere else in the code?

waltjohnson (approved):

python/pylib/data_sets.py
waltjohnson: Yes.

src/data_sets.h
waltjohnson: wheel_config_t is used in the nvm_flash_config_t and also inside of ground_vehicle_t.  So we need both.

PR #135, opened 2021-08-24
Improve GPS velocity validation

waltjohnson:
This addresses an issue we discovered in Fortem's data where GPS2 which didn't have an antenna connected reported a bogus ECEF velocity when the speed accuracy (sAcc) said it was valid yet the GPS position did not have fix and the GPS time was obviously incorrect.

Now we use the same status flags from GPS pos in the GPS vel message so we can use the GPS position status to validate the GPS velocity.

### DIFF
--- a/src/data_sets.h
+++ b/src/data_sets.h
@@ -811,7 +811,7 @@ typedef struct PACKED
 	/** Speed accuracy in meters / second */
 	float					sAcc;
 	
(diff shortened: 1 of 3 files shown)


steven-fusco-inertialsense (approved):

steven-fusco-inertialsense:
related: https://github.com/inertialsense/imx/pull/122/files

PR #136, opened 2021-08-24
Luna 1598 evb pwm motor control

waltjohnson:
Refactored EVB wheel control into dynamic classes.  Added PWM wheel control class for RK2K mower.


src/cltool.cpp
```diff
@@ -294,6 +294,12 @@ bool cltool_parseCommandLine(int argc, char* argv[])
 			g_commandLineOptions.replayDataLog = true;
 			g_commandLineOptions.replaySpeed = (float)atof(&a[4]);
 		}
+        else if (startsWith(a, "-resetEvb") ||
```
steven-fusco-inertialsense, line 297: I wonder if it's worth adding a parameter to startsWith to allow case insensitive matching.  This won't catch reseteVB or resetevB

steven-fusco-inertialsense (approved):

steven-fusco-inertialsense:
This also looks like it includes the date / time stamp parsing fix, I think you are going to end up with some merge conflicts if you aren't careful.

PR #149, opened 2021-10-17
SN-448 SN-507 Fix bad numbers in RTOS task profiling

steezeman:
This should fix both issues, but it needs to be tested with imx on a testbed for at least 5 days before we can merge. Timer should roll over in 4.9 days. Builds on uINS-3 and EVB-2, needs more testing on both platforms.


steezeman:
@jakefawson1 This is ready for another week of testing.

steven-fusco-inertialsense (approved):

waltjohnson (approved):

PR #169, opened 2021-12-01
SN-968 Clean up bootloader

waltjohnson:
- Fixed issues with using cltool to update bootloader firmware.
- Remove old unused code from bootloader.  Updating bootloader firmware must be done along with updating application firmware.
- Consolidated application and bootloader firmware update in cltool.
- Added -fb forceBootloaderUpdate option to cltool.
- Fix failed to update firmware error status.
- Added error checking to catch invalid filename or if force update option is supplied without a file name. 
- Change return state of bootloadFileEx() to 0 on success, -1 on failure.


steven-fusco-inertialsense:
How many devices have we tested this on?

waltjohnson:
> How many devices have we tested this on?

16 simultaneously works great both on the EvalTool and cltool.

steven-fusco-inertialsense (approved):

PR #178, opened 2021-12-21
SN-1004 Update idle task to work on uINS-5

steezeman:
The method to put the Microchip and STM32 chips we have in our products are almost exactly the same. I broke out the calls in the pmc_enable_sleepmode function. @rudomorg, I need your help to test the following:

- uINS-3 power consumption with the sleep lines commented and uncommented
- uINS-5 power consumption with the sleep lines commented and uncommented

Once we can confirm that these lines are indeed putting us into sleep mode, we can go ahead and merge this.

### DIFF
--- a/hw-libs/misc/rtos.c
+++ b/hw-libs/misc/rtos.c
@@ -129,8 +129,11 @@ void vApplicationIdleHook(void)
 {
     // Sleep to reduce power consumption
 #ifndef uINS_5
-    pmc_enable_sleepmode(0);	// TODO: Implement this function for uINS_5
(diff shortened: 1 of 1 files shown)


waltjohnson (approved):

waltjohnson:
@steezeman I'm assuming you've tested this on the uINS-3?

steezeman:
@waltjohnson yes, it runs on the uINS-3. Morgan and I will confirm that the power consumption looks the same next week.

steezeman:
image

Looks like this works on the uINS-5. Already tested on the uINS-3, for basic operation. I have very high confidence this will draw the same power on the uINS-3, the function call's contents were simply broken out into the idle hook.

PR #181, opened 2021-12-22
LUNA-1979 add public python math lib

steven-fusco-inertialsense:
I moved pose.py into it's own module and published it to pip

In order to better support Windows, I've removed the symlink and changed things for relative paths


steven-fusco-inertialsense:
Published here: https://pypi.org/project/inertialsense-math/

chasekunz:
Did you symlink in the file into the old location so that all of the dependent code still works?

steven-fusco-inertialsense:
> so that all of the dependent code still works?

yes, that's the idea, but I still think we should remove the symlink and actually use the new lib

chasekunz:
Agree

chasekunz (approved):

waltjohnson (approved):

waltjohnson (approved):

waltjohnson:
@steven-fusco-inertialsense I ran LogInspector and did the full build in Windows and Linux and it passes.  

We created a ci_hdw python project that uses this library.  It is supposed to run continuously in the background and automatically download and run the latest in the development branch.  It runs several tests in hardware to check that the system didn't break.  However, the ci_hdw project hasn't been used for over a year, since we changed buildings.  We need to resurrect that code at some point.

chasekunz (approved):

PR #184, opened 2021-12-24
bugfix/SN-1012 logger drops data

waltjohnson:
- Fixed significant issue with ISLogger and sdat data logging.
- Fixed ISLogger copy function so that .dat, .sdat, and .csv file formats can be interconverted without dropping data.
- Fixed the way data is serialized when read out of ISLogger .sdat data format.
- Fixed broken unit test for ISLogger and added more robust checks.
- Convert cLogStats objects from dynamic to static allocation inside cISLogger.

### DIFF
--- a/src/DeviceLog.cpp
+++ b/src/DeviceLog.cpp
@@ -78,14 +78,17 @@ void cDeviceLog::InitDeviceForReading()
 
 bool cDeviceLog::CloseAllFiles()
 {
+	if (m_writeMode)
+	{
 #if 1
-    string str = m_directory + "/stats_SN" + to_string(m_devInfo.serialNumber) + ".txt";
-    m_logStats.WriteToFile(str);
+		string str = m_directory + "/stats_SN" + to_string(m_devInfo.serialNumber) + ".txt";
+		m_logStats.WriteToFile(str);
 #else   // stringstream not working on embedded platform
-	ostringstream serialNumString;
-	serialNumString << m_devInfo.serialNumber;
-    m_logStats.WriteToFile(m_directory + "/stats_SN" + serialNumString.str() + ".txt");
+		ostringstream serialNumString;
+		serialNumString << m_devInfo.serialNumber;
+		m_logStats.WriteToFile(m_directory + "/stats_SN" + serialNumString.str() + ".txt");
 #endif
+	}
     return true;
 }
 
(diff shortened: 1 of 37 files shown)


src/DeviceLog.cpp
```diff
@@ -71,20 +71,24 @@ void cDeviceLog::InitDeviceForReading()
...
 bool cDeviceLog::CloseAllFiles()
 {
+	if (m_writeMode)
+	{
 #if 1
-    string str = m_directory + "/stats_SN" + to_string(m_devInfo.serialNumber) + ".txt";
-    m_logStats->WriteToFile(str);
+		string str = m_directory + "/stats_SN" + to_string(m_devInfo.serialNumber) + ".txt";
+		m_logStats.WriteToFile(str);
 #else   // stringstream not working on embedded platform
```
steven-fusco-inertialsense, line 86: We can probably just delete this block.  Why keep it?

src/DeviceLogCSV.cpp
```diff
@@ -230,7 +230,7 @@ bool cDeviceLogCSV::SaveData(p_data_hdr_t* dataHdr, const uint8_t* dataBuf)
 	}
 
 	// Write date to file
-    int nBytes = m_csv.WriteDataToFile(++m_nextId, log.pFile, *dataHdr, dataBuf);
+    int nBytes = m_csv.WriteDataToFile(m_nextId++, log.pFile, *dataHdr, dataBuf);
```
steven-fusco-inertialsense, line 233: Nice catch.  This way it starts at 0?

src/DeviceLogSorted.cpp
```diff
@@ -39,7 +45,7 @@ cDeviceLogSorted::cDeviceLogSorted()
 
 void cDeviceLogSorted::InitDeviceForWriting(int pHandle, std::string timestamp, std::string directory, uint64_t maxDiskSpace, uint32_t maxFileSize)
 {
-    for (uint32_t i = 0; i < DID_COUNT; i++)
+	for (uint32_t i = 0; i < DID_COUNT; i++)
```
steven-fusco-inertialsense, line 48: Tabs or Spaces?  Looks like we have some mixed spacing going on?

steven-fusco-inertialsense (approved):

src/DeviceLogSorted.cpp
waltjohnson: Do you have a good way of preventing this?  I think vscode is supposed to detect the indentation style and use the similar one.  I can spend time arrow keying through the start of each line to determine how it's being indented.

PR #188, opened 2022-01-16
Add higher baudrate support for lower latency data read

waltjohnson:
Per Aim-locks request, this provides a way to read data over standard serial ports (and RS485 and Rugged-2) at 10Mbps and 18.75Mbps.


src/ISComm.c
```diff
@@ -81,7 +81,19 @@ unsigned int getBitsAsUInt32(const unsigned char* buffer, unsigned int pos, unsi
 	return bits;
 }
 
-const unsigned int g_validBaudRates[IS_BAUDRATE_COUNT] = { IS_BAUDRATE_3000000, IS_BAUDRATE_921600, IS_BAUDRATE_460800, IS_BAUDRATE_230400, IS_BAUDRATE_115200, IS_BAUDRATE_57600, IS_BAUDRATE_38400, IS_BAUDRATE_19200, IS_BAUDRATE_9600 };
+const unsigned int g_validBaudRates[IS_BAUDRATE_COUNT] = { 
+	                        // Actual on uINS:
+	IS_BAUDRATE_18750000,   // 18750000 (uINS ser1 only) 
+	IS_BAUDRATE_9375000,    // 9375000
+	IS_BAUDRATE_3125000,    // 3125000
```
steezeman, line 88: Did we ever support 3000000 baud, or was it always 3125000?

steezeman (approved):
Assuming this is tested, looks good to me.

src/ISComm.c
waltjohnson: Yes, it was always 3125000.  All baudrates were tested on the oscope and look good.

waltjohnson:
> Assuming this is tested, looks good to me.

Yes, I tested with oscope.  Baudrates are spot on.  One question I had for you was if we can support these baudrates on the uINS-5.

steezeman:
> One question I had for you was if we can support these baudrates on the uINS-5.

@waltjohnson as far as I know, the USART peripherals we are using on the uINS-5 have high granularity in baud rate selection, so there should be no problem.

waltjohnson:
What I found on the SAME70 is that you have a maximum clock speed upon
which all of the peripherals are based.  And the closer you get to the max
peripheral clock speed, the more granular the data rate gets.

For example, the equation for the UART baud generator in the SAME70 is

baud_rate_generator_register = peripheral_clock_rate (150 MHz) / ( 16 *
desired_baud_rate)

So in the end the baudrate jumps from 9.37 Mbps to 18.75 Mbps.  There's
nothing in between.

We can look into this later.

PR #240, opened 2022-06-09
SN-1819 Remove using namespace std from SDK header files

waltjohnson:
This removes `using namespace std` from SDK header files as they will conflict when included in other files.


siddirp:
Out of our hundreds(?) of customers we've only had one person complain about this. Isn't it more likely that this will break the code of all of our other customers' implementation for this small change that this user can do themselves?

PR #289, opened 2022-09-08
Renaming vars per Alex's recommendation

waltjohnson:


aabogdanov (approved):
Found a few instances of PxyzNED, renamed to StdPosNed

PR #326, opened 2022-11-26
SN-2708 UART-based SAM-BA firmware update fixes

steezeman:
- Put device into 115200 baud mode when using SAM-BA first-stage bootloader
- Fix SAM-BA upload over UART (was using wrong buffer)
- Clean up formatting


steezeman:
- Remove debug code
- Fix case where serial port closes before last transmission is complete

waltjohnson (approved):

PR #335, opened 2022-12-28
SN-2679 consolidate sdk and ROS repos

waltjohnson:
Consolidates the inertial-sense-sdk and inertial_sense_ros repos into one.


steezeman (approved):

ros/.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "lib/inertial-sense-sdk"]
```
kylemallory, line 1: This should be removed, since ros does not have submodules.

ros/.gitignore
```diff
@@ -0,0 +1,6 @@
+*.autosave
```
kylemallory, line 1: Since this is now incorporated into the parent directory, this can go.  We should probably include relevant ignore directives into the parent's .gitignore

kylemallory (changes requested):

ros/.gitmodules
waltjohnson: Done.  Thanks.

ros/.gitignore
waltjohnson: Done.  Thanks.

kylemallory (approved):

PR #337, opened 2022-12-29
Fix compiler alignment warnings for FreeRTOS stream buffer

steezeman:
All of our projects with FreeRTOS were throwing this warning. Only the EVB-2 uses stream buffers, so the fix was two-fold:

1. On the EVB-2, switch to statically allocated stream buffers. This completely avoids the issue with alignment. If you read the FreeRTOS source code, they assume that malloc is aligned, which I'm not a fan of. I put an `#if 0` ... `#endif` around dynamically allocated stream buffers. I had to decrease the heap size to account for the buffers being allocated statically.
1. On our internal projects, don't compile the `stream_buffers.c` file (see PR in internal repo) and disable stream buffers in the `FreeRTOSConfig.h` file.

```
../hw-libs/freertos/stream_buffer.c: In function 'xStreamBufferGenericCreate':
../hw-libs/freertos/stream_buffer.c:288:43: warning: cast increases required alignment of target type [-Wcast-align]
             prvInitialiseNewStreamBuffer( ( StreamBuffer_t * ) pucAllocatedMemory,       /* Structure at the start of the allocated memory. */ /*lint !e9087 Safe cast as allocated memory is aligned. */ /*lint !e826 Area is not too small and alignment is guaranteed provided malloc() behaves as expected and returns aligned buffer. */
                                           ^
../hw-libs/freertos/stream_buffer.c:301:16: warning: cast increases required alignment of target type [-Wcast-align]
         return ( StreamBufferHandle_t ) pucAllocatedMemory; /*lint !e9087 !e826 Safe cast as allocated memory is aligned. */
                ^
```

### DIFF
--- a/EVB-2/IS_EVB-2/src/communications.cpp
+++ b/EVB-2/IS_EVB-2/src/communications.cpp
@@ -50,6 +50,12 @@ is_comm_instance_t 			g_commTx = {};
 StreamBufferHandle_t        g_xStreamBufferUINS;
 StreamBufferHandle_t        g_xStreamBufferWiFiRx;
 StreamBufferHandle_t        g_xStreamBufferWiFiTx;
+static uint8_t s_xStreamBufferUINS_buf[ STREAM_BUFFER_SIZE ] __attribute__((aligned(4)));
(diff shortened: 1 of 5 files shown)


waltjohnson (approved):
@steezeman Please put a note in the change_log.md for this.  Thanks!

steezeman:
Done in internal repo.

PR #339, opened 2022-12-31
SN-2679 ros sdk cleanup

waltjohnson:
General cleanup of ros node.  Consolidate redundant code into function or macros.  Move all topic nh_.advertise() into the constructor so we can see topic names side by side.


waltjohnson:
@kylemallory Here are some items before we merge:

1. Can we delete SDK\ros\launch\test_yaml_config.yaml?  It seems redundant or unused.
2. Did you say we use the hexidecimal representation for number in yamls (i.e. ioConfig: 40673376).  If so we should convert our decimal numbers to hexidecimal.  Or is there a limitation on the ros param server that won't read these in?
3. I think we should either comment out comment out all parameters except a minimal set used in a typical application or have a separate file that contains the minimal set used for a typical application.  We want the example to stream only the typical message/topics and not extras.
4. We need to add a bandwidth limitation / buffer overflow / data drop indication in the inertial sense ROS node so customers realize if they have too many messages enabled not all data is making it over communications port.  This is also part of the reason we should reduce the amount of messages enabled in the previous bullet.

kylemallory:
1) The test_yaml_config.yaml I felt was redundant too, but then I realized that we have a need, specifically for CI tests, to validation different configuration options, which may not be (or should not be) in a valid yaml file.  So, I added it.
2) Yes, Hexadecimal is natively supported, so we just need to review which parameters we want in illustrated in hex, and then change them.
3) Yes, I think a minimal example, and a "complete" yaml, which serves more for documentation purposes makes sense.  I suspect @andrewpriddis may be the best candidate for this, as he's got the a better sense of what customers might like to see in the example config.
4) We can do SN-3081, but I think we need to get this merged in and slate that for another PR (but we can still prioritize it for the release. However, I don't want to delay this PR any further.
5) The disclaimer should not be removed. We can discuss further.

PR #344, opened 2023-01-04
SN-2679 Created c++ class for param helper

waltjohnson:
Get param names to match topic names names.  Converted rs_param_t to ParamHelper class.


siddirp (approved):
It's fine. Honestly, I feel like lots of work went into this to get minimal benefit. I guess from our meetings I still don't fully grasp what in Luna the driver was. Lots of breaking changes for customers without a clear driver.

PR #348, opened 2023-01-11
SN-2901 SN-2611 DFU fixes

steezeman:


src/ISBootloaderBase.cpp
```diff
@@ -451,15 +451,27 @@ is_operation_result cISBootloaderBase::update_device
         (obj)->m_use_progress = true;
         addMutex->lock();
         contexts.push_back(obj);
-        *new_context = obj;
         addMutex->unlock();
-        if((obj)->download_image(filenames.bl_IMX_5.path) != IS_OP_OK)
+        for(size_t i = 0; i < 3; i++)
```
kylemallory, line 455: A comment here, indicating that its a retry mechanism would be helpful.

src/ISBootloaderISB.cpp
```diff
@@ -992,13 +1013,17 @@ is_operation_result cISBootloaderISB::download_image(std::string filename)
 
     status_update("(ISB) Erasing flash...", IS_LOG_LEVEL_INFO);
 
-    if(erase_flash() != IS_OP_OK) return IS_OP_ERROR;
-    if(select_page(0) != IS_OP_OK) return IS_OP_ERROR;
+    result = erase_flash();
+    CHECK_IS_OP_DWNLD()
```
kylemallory, line 1017: I'm not sure I see the value in this macro.
I think just updating the existing condition check to  `if ((result = erase_flash()) != IS_OP_OK) { fclose(firmware_file); return result; }` is more readable, than having to move back up the code the find what CHECK_IS_OP_DWNLD() is doing.  Plus, its clearer when stepping through with the debugger.

Alternatively, in the lower calls to CHECK_IS_OP_DWNLD(), you could still keep the `result = function_op()` and just add the `if (result != IS_OP_OK) { ... }.

Finally, it might get a little "nesty", but also doing this is a nested fashion, might also be appropriate:
`if ( (result=erase_flash()) == IS_OP_OK) then { if (result=select_page()) == IS_OP_OK) { ... } } fclose(firmware_file);` then in reading the nested code, it is straight forward that failure of one, means non-execution of the nested, and we keep a consistent, singular close/return result).

Just suggestions though... as long as it works!

src/ISBootloaderThread.cpp
```diff
@@ -332,6 +332,20 @@ void cISBootloaderThread::update_thread_libusb(void* context)
     m_libusb_thread_mutex.unlock();
 }
 
+#define RETURN_CHECK_IF_CANCELLED   if(m_uploadProgress(NULL, 0.0f) == IS_OP_CANCELLED) { m_continue_update = false; m_update_in_progress = false; m_update_mutex.unlock(); if(m_waitAction) m_waitAction(); return vector<confirm_bootload_t>(); }
+#define RETURN_IF_CANCELLED         if(m_uploadProgress(NULL, 0.0f) == IS_OP_CANCELLED) { m_continue_update = false; m_update_in_progress = false; m_update_mutex.unlock(); if(m_waitAction) m_waitAction(); return IS_OP_CANCELLED; }
```
kylemallory, line 336: Similar here.  I see why you're using them, but they feel a bit hackish.  I'm not entirely sure what is happening here with the result values though, so maybe it makes sense.

kylemallory (approved):
I'm good with this, and approve, but consider my comments.  Readability for others (and yourself in a year from now) is important, and I do think it can be improved.  If you can improve its readability with an extra 10-20 minutes of work, it will save someone else that same amount of time trying to figure it out down the road.

steezeman:
@kylemallory I fixed the readability issues you pointed out

PR #355, opened 2023-02-01
Fix logInspector for PyQt5 5.15.7

AchmadFathoni:
Since the python dependency version is not locked in `setup.py`,  the `pip install logInspector/` will always install the latest version python dependency . This commit will fix logInspector for the current latest PyQt5 5.15.7.

I suggest we lock the python dependency version.


steezeman:
@AchmadFathoni thanks for the PR. We will review soon.

waltjohnson (approved):
This works.

PR #381, opened 2023-04-04
logInspector icon and terminal improvements

steezeman:
- Adds an icon to the taskbar in Windows that was previously missing. The way we were using the QMainWindow class wasn't working with some other parts of Qt related to this. 
- Adds a handler for Ctrl-C in the terminal. Qt doesn't let the terminal do anything after the exec() call, so we set a timer for 5Hz and let the system manage interrupts at that interval.
- Clean up warning in terminal related to improper usage of Qt image classes

### DIFF
--- a/python/logInspector/logInspector.py
+++ b/python/logInspector/logInspector.py
@@ -1,25 +1,19 @@
 #!/usr/bin/python3
 
-import sys, os, shutil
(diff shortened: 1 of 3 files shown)


waltjohnson (approved):
Looks good.

PR #383, opened 2023-04-10
Fix/minor tweaks

kylemallory:


waltjohnson (approved):

siddirp (approved):

ros/src/inertial_sense_ros.cpp
```diff
@@ -33,6 +33,15 @@
 
 #define STREAMING_CHECK(streaming, DID)      if(!streaming){ streaming = true; ROS_INFO("%s response received", cISDataMappings::GetDataSetName(DID)); }
 
+void odometryIdentity(nav_msgs::Odometry& msg_odom) {
```
siddirp, line 36: This function could use a description.

kylemallory: Thanks for the reminder!

PR #387, opened 2023-04-19
SN-3334 yaml-cpp SDK test

siddirp:
Fixes to YAML functions in SDK. Batch script to install yaml-cpp in Windows. Put back the apt install of yaml-cpp

### DIFF
--- a/src/InertialSense.cpp
+++ b/src/InertialSense.cpp
@@ -1114,7 +1114,6 @@ void InertialSense::CloseSerialPorts()
 }
 void InertialSense::SaveFlashConfigFile(std::string path, int pHandle)
 {
-#if 1
 	nvm_flash_cfg_t* outData = &m_comManagerState.devices[pHandle].flashCfg;
 
 	YAML::Node map = YAML::Node(YAML::NodeType::Map);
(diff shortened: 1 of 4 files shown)


src/InertialSense.cpp
```diff
@@ -1237,101 +1237,105 @@ void InertialSense::SaveFlashConfigFile(std::string path, int pHandle)
 
 int InertialSense::LoadFlashConfig(std::string path, int pHandle)
 {
-#if 0
+#if 1
```
kylemallory, line 1240: Remove these (since they are always true, there is no point in them being conditional), or change the 1 to a macro flag (ie, #ifdef USE_YAML_CPP)

kylemallory (changes requested):

src/InertialSense.cpp
```diff
@@ -1114,7 +1114,7 @@ void InertialSense::CloseSerialPorts()
 }
 void InertialSense::SaveFlashConfigFile(std::string path, int pHandle)
 {
-#if 0
+#if 1
```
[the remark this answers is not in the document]
siddirp: Done.

src/InertialSense.cpp
siddirp: Done

kylemallory (approved):
Looks good!

PR #388, opened 2023-04-19
Follow up fixes for debug libraries

jjxtra:
I've made the libraries included in InertialSense.cpp, this allows bringing in the "d" version of the libraries for debug builds, and the non "d" versions for release, at least on Windows. For Linux, they are part of the packages on the system still and seem to build fine in debug and release mode.

Opening in Visual Studio as a folder worked, but I had to delete the .vs folder in the cltool directory first.


jjxtra:
Example projects are building, along with the sdk and able to use the sdk static lib.

@waltjohnson This pr removes the library linking from the source code and puts it all in the cmake files. For the eval tool, we will need to do some refactoring on including the libusb, yamlcpp and winsock lib files.

steezeman (approved):

kylemallory (commented):
I would like to move the yaml-cpp back to a submodule.  I might feel differently if it is the ONLY submodule, but if we are linking libusb-1.0 or other packages also as a submodule, we've already incurred the burden of submodules (1 vs 10 submodules is effectively the same effort), but it also means our code a more properly segmented, and we can update to later versions of yaml-cpp pretty easily, rather than dragging "pinned" code around with ours.  I'm going to leave this without explicit approval, as I'm willing to allow it, but its not my preferences to have it embedded into our code.

kylemallory (approved):

waltjohnson (approved):

PR #411, opened 2023-07-09
SN-3840 add isb support

waltjohnson:
Add Inertial Sense Bootloader (ISB) support back into IMX-5 project.


kylemallory (approved):
Looking over some of the changes, I'm not sure everything that motivated them.  Was this all stuff that was taken out, and you are just putting it back in?  Or did you have to make additional improvements in order to make it work?

PR #418, opened 2023-07-21
Seemingly related to new NMEA messages

kylemallory:
Draft.  If its not necessary, close it and I'll delete it.


waltjohnson:
Dave decided to re-write our NMEA library.  However, the changes add considerable complexity and processing overhead that I don't feel are merited/justified.  It was a nice idea but unnecessary and not worth merging for use in our products.  We may want to archive the new NMEA library files he created but we should not merge them for use in our products at this time.

PR #421, opened 2023-08-03
Sn 3756 update durring ppd

tmcd249:
Fixes issue in https://inertialsense.atlassian.net/browse/SN-3756. Able to start updating regardless of streaming speed. 330Kb/s tested.

### DIFF
--- a/src/ISBootloaderAPP.cpp
+++ b/src/ISBootloaderAPP.cpp
@@ -46,7 +46,7 @@ eImageSignature cISBootloaderAPP::check_is_compatible()
 
     // Get DID_DEV_INFO from the uINS.
     is_comm_instance_t comm;
-    uint8_t buffer[2048];
+    uint8_t buffer[65536];
     is_comm_init(&comm, buffer, sizeof(buffer));
     int messageSize;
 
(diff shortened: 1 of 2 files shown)


src/ISBootloaderAPP.cpp
```diff
@@ -48,14 +48,24 @@ eImageSignature cISBootloaderAPP::check_is_compatible()
     is_comm_instance_t comm;
     uint8_t buffer[2048];
     is_comm_init(&comm, buffer, sizeof(buffer));
-    int messageSize;
+    int messageSize, n, i, m;
+
+    // clear the Rx serial port
+    n = is_comm_free(&comm);
+    for (i = 0; i < 20; i++)
```
kylemallory, line 55: This seems a little non-deterministic, and unclear (but I follow its intent).  It maybe clearer to write this as a while() loop, with a retry-count that breaks to prevent locking.  Also, I generally recommend using an actual timeout, with a timer, rather than an arbitrary retry counter.

src/ISBootloaderAPP.cpp
```diff
@@ -48,14 +48,24 @@ eImageSignature cISBootloaderAPP::check_is_compatible()
     is_comm_instance_t comm;
     uint8_t buffer[2048];
     is_comm_init(&comm, buffer, sizeof(buffer));
-    int messageSize;
+    int messageSize, n, i, m;
```
kylemallory, line 51: Curious, but not a problem, why you moved all the temp variables up here, rather that keeping them in their respective scopes?

kylemallory (approved):

src/ISBootloaderAPP.cpp
tmcd249: Commenting section to clarify code

PR #422, opened 2023-08-06
SN-3585 v2 packet format

waltjohnson:
Revised InertialSense binary (ISB) protocol.  Refactored and reduced packet header/footer size from 20 bytes to 8 byes per packet.  Removed ambiguity of Tx and Rx buffers in ISComm instance.  Optimized sending and reading packets by removing extra memcpy calls.


kylemallory (changes requested):
This is a BIG PR, but looks like its does a lot of good things (like we've discussed).  Overall, the comms code still feels very cobbled together over its life (lots of varying code styles, etc), but I think this is all work in the right direction.  That said, It feels like there is a lot of room for potential to decouple and abstract some of the concepts out, taking advantage of C++.  Maybe there is a reason we are holding on the C patterns here, but with other areas of code in C++, I wonder if we are holding ourselves back by trying to minimize the scale of the refactor, and stay within existing bounds, despite it already being a significant code change.  I guess, if feels like we discussed how the code behind parsing and buffer handling could be more optimal, and refactored based on that, but didn't stop to discuss if/how the code itself could be reworked to make it more decoupled, more efficient, and easier to read.  Feels like a missed opportunity.

kylemallory (approved):

PR #433, opened 2023-08-29
Sn 4090 hard fault from request

tmcd249:
A hard fault was caused by an index out-of-bounce exception when GPX flash config was requested. This is because the did is 121 which tries to access an array of 100 elements. The check in place only checked that the value was non-zero.


tmcd249:
wrong merge point

PR #438, opened 2023-09-04
SN-4098 nmea intel bug

waltjohnson:
Created NMEA $INTEL CI-hdw and SDK unit tests.


src/protocol_nmea.cpp
```diff
@@ -217,7 +217,7 @@ char *ASCII_to_f64(double *vec, char *ptr)
 	return ptr;
 }
 
-char *ASCII_to_vec4u8(uint8_t vec[], char *ptr)
+char *ASCII_to_ver4u8(uint8_t vec[], char *ptr)
```
kylemallory, line 220: Did you mean to rename this, for vec4 to ver4?  I'm getting this is our version info, and not a 4-element vector?

kylemallory (approved):
Other than just confirming the rename on the function, looks good.

src/protocol_nmea.cpp
waltjohnson: Yes.  I figured referring to it as a version rather than a vector is more appropriate as the version is decimal delimited were as a vector is comma delimited.

PR #449, opened 2023-09-17
Development query dev info

waltjohnson:
This switches use of inertial sense binary to NMEA in the SDK, cltool, and EvalTool for the following commands: query device info, software reset, stop messages streaming, and save persistent messages.  Also adds the device hardware type (IMX, GPX, uINS, EVB) to dev_info_t and NMEA $INFO while maintaining backwards compatibility.   These changes only minimal impact on the firmware and facilitate function during firmware upgrade and downgrade.  This allow either v1 or v2 SDK, Cltool, and EvalTool to communicate and upgrade or down firmware despite the breaking v2 protocol changes.

This change was applied to the development branch so that release 1.11 will have ability to see and downgrade release 2.x firmware just the same as release 2.x can upgrade release 1.x.


src/ISBootloaderBase.cpp
tmcd249, line 1: in the statement:
 IS_IMAGE_SIGN_IMX_5: target_signature = bootloaderRequiredSignature_uINS_5
Is this supposed to still say:
     bootloaderRequiredSignature_uINS_5

tmcd249 (approved):

kylemallory (approved):

PR #470, opened 2023-10-13
Sn 4293 iscomm parser refactor

waltjohnson:
Revise v2 protocol packet parser to only parse one packet at a time (no concurrent parsing anymore).  We only care to parse the highest level packets.  Packet parsing is invalidate either by 1.) exceeding max packet length or 2.) incomplete packet received by timeout.  The prior method of allowing concurrent parsing was a bad idea as it allowed encapsulated packets (packets inside packets) to invalidate the outer packet parse.

### DIFF
--- a/src/test/test_ISComm.cpp
+++ b/src/test/test_ISComm.cpp
@@ -43,9 +43,6 @@ extern "C"
 
 typedef struct
 {
-	// com_manager_t			cm;
-	// com_manager_status_t	cmBufStatus[NUM_COM_PORTS] = { 0 };
-	// broadcast_msg_t			cmBufBcastMsg[MAX_NUM_BCAST_MSGS] = { 0 };
 	struct 
 	{
 		dev_info_t			devInfo;
@@ -1192,7 +1189,7 @@ TEST(ISComm, alternating_isb_nmea_parse_error_check)
 
 
 #if TEST_TRUNCATED_PACKETS
-TEST(ISComm, IncompletePackets)
+TEST(ISComm, TruncatedPackets)
 {
 	// Initialize Com Manager
 	init(tcm);
--- a/src/test/test_ISComm.cpp
+++ b/src/test/test_ISComm.cpp
@@ -688,7 +688,6 @@ static void ringBuftoRingBufWrite(ring_buf_t *dst, ring_buf_t *src, int len)
 
 
 #if BASIC_TX_BUFFER_RX_BYTE_TEST
-uint8_t g_buf[1024];
 TEST(ISComm, BasicTxBufferRxByteTest)
 {
 	// Initialize Com Manager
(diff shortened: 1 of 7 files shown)


tmcd249 (approved):

src/test/test_ISComm.cpp
```diff
@@ -1188,3 +1189,147 @@ TEST(ISComm, alternating_isb_nmea_parse_error_check)
...
+TEST(ISComm, IncompletePackets)
+{
+	// Initialize Com Manager
+	init(tcm);
+
+	// Generate and add data to deque
+	generateData(g_testTxDeque);
+
+	int badPktCount = 0, goodPktCount = 0;
+	uint8_t buf[4096] = {0};
```
kylemallory, line 1204: A bigger buffer.

src/test/test_ISComm.cpp
```diff
@@ -1188,3 +1189,147 @@ TEST(ISComm, alternating_isb_nmea_parse_error_check)
...
+			}
+
+			data_holder_t td = g_testTxDeque.front();
+			g_testTxDeque.pop_front();
+
+			switch (ptype)
+			{
+			case _PTYPE_INERTIAL_SENSE_DATA:
+				// Found data
+				DEBUG_PRINTF("[%d] (%d) Good %d: ISB %3d\n", found, comm.rxPkt.data.size, g_comm.rxPktCount, comm.rxPkt.hdr.id);
```
kylemallory, line 1282: I don't think I've seen DEBUG_PRINTF before... is this new?  I'm assuming this is only used if its not-embedded?

kylemallory (approved):

src/test/test_ISComm.cpp
```diff
@@ -706,8 +706,9 @@ TEST(ISComm, BasicTxBufferRxByteTest)
 		switch (td.ptype)
 		{
 		default:	// IS binary
-			n = is_comm_data_to_buf(g_buf, sizeof(g_buf), &g_comm, td.did, td.size, 0, td.data.buf);
-			portWrite(0, g_buf, n);
+			uint8_t buf[1024];
```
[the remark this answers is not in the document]
waltjohnson: These are all test buffers (used only in test code).  But I've increased the buffer size to be consistent with the serial port buffer size.  It only needs to handle the max packet size but it doesn't hurt when running on a PC.

src/test/test_ISComm.cpp
waltjohnson: Same as above.

src/test/test_ISComm.cpp
waltjohnson: Yes, non-embedded only for testing.  This is a simple way to enable print statements when I need to debug that I don't want to delete.

PR #490, opened 2023-11-03
General EvalTool GPX update polish items.

tmcd249:
General EvalTool GPX update polish items.

### DIFF
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -27,7 +27,7 @@ fwUpdate::update_status_e ISFirmwareUpdater::initializeUpdate(fwUpdate::target_t
     updateStartTime = current_timeMs();
     nextStartAttempt = current_timeMs() + attemptInterval;
     fwUpdate::update_status_e result = (fwUpdate_requestUpdate(_target, slot, chunkSize, fileSize, session_md5, progressRate) ? fwUpdate::NOT_STARTED : fwUpdate::ERR_UNKNOWN);
-    printf("Requested Firmware Update to device '%s' with Image '%s', md5: %08x-%08x-%08x-%08x\n", fwUpdate_getSessionTargetName(), filename.c_str(), session_md5[0], session_md5[1], session_md5[2], session_md5[3]);
+    printf("Requested Firmware Update to device '%s' with Image '%s', md5: %08x-%08x-%08x-%08x\n%s\n", fwUpdate_getSessionTargetName(), filename.c_str(), session_md5[0], session_md5[1], session_md5[2], session_md5[3], fwUpdate_getNiceStatusName(result));
     return result;
 }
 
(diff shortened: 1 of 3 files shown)


src/ISFirmwareUpdater.cpp
```diff
@@ -27,7 +27,7 @@ fwUpdate::update_status_e ISFirmwareUpdater::initializeUpdate(fwUpdate::target_t
     updateStartTime = current_timeMs();
     nextStartAttempt = current_timeMs() + attemptInterval;
     fwUpdate::update_status_e result = (fwUpdate_requestUpdate(_target, slot, chunkSize, fileSize, session_md5, progressRate) ? fwUpdate::NOT_STARTED : fwUpdate::ERR_UNKNOWN);
-    printf("Requested Firmware Update to device '%s' with Image '%s', md5: %08x-%08x-%08x-%08x\n", fwUpdate_getSessionTargetName(), filename.c_str(), session_md5[0], session_md5[1], session_md5[2], session_md5[3]);
+    printf("Requested Firmware Update to device '%s' with Image '%s', md5: %08x-%08x-%08x-%08x\n%s\n", fwUpdate_getSessionTargetName(), filename.c_str(), session_md5[0], session_md5[1], session_md5[2], session_md5[3], fwUpdate_getNiceStatusName(result));
```
kylemallory, line 30: We should revert this...  It doesn't gain us anything

src/ISFirmwareUpdater.cpp
```diff
@@ -105,7 +105,20 @@ bool ISFirmwareUpdater::fwUpdate_handleDone(const fwUpdate::payload_t &msg) {
     return true;
 }
 
-fwUpdate::msg_types_e ISFirmwareUpdater::fwUpdate_step() {
+bool ISFirmwareUpdater::fwUpdate_isDone()
+{
+    if (hasPendingCommands())
```
kylemallory, line 110: While this is easier to read, it should probably just be a series of if ( () || () || () || () ) return false;

src/InertialSense.cpp
```diff
@@ -1025,7 +1025,7 @@ bool InertialSense::isFirmwareUpdateFinished() {
     for (int i = 0; i < (int) m_comManagerState.devices.size(); i++) {
         ISFirmwareUpdater *fwUpdater = m_comManagerState.devices[i].fwUpdater;
         if (fwUpdater != nullptr) {
-            if (fwUpdater->hasPendingCommands() || (fwUpdater->fwUpdate_getSessionStatus() < fwUpdate::NOT_STARTED) || (fwUpdater->fwUpdate_getSessionStatus() >= fwUpdate::FINISHED))
+            if (!fwUpdater->fwUpdate_isDone())
```
kylemallory, line 1028: This two ifs should be combined: `if (fwUpdater != null && fwUpdate->fwUpdate_isDone()) return false;`

src/InertialSense.cpp
```diff
@@ -1039,10 +1039,8 @@ bool InertialSense::isFirmwareUpdateSuccessful() {
     for (int i = 0; i < (int) m_comManagerState.devices.size(); i++) {
         ISFirmwareUpdater *fwUpdater = m_comManagerState.devices[i].fwUpdater;
         if (fwUpdater != nullptr) {
-            if (fwUpdater->hasPendingCommands() || (fwUpdater->fwUpdate_getSessionStatus() < fwUpdate::NOT_STARTED))
+            if (fwUpdater->hasPendingCommands() || ((fwUpdater->fwUpdate_getSessionStatus() >= fwUpdate::NOT_STARTED) && (fwUpdater->fwUpdate_getSessionStatus() < fwUpdate::FINISHED)))
```
kylemallory, line 1042: This should also be updated to use `if (fwUpdater != null && !fwUpdate->fwUpdate_isDone()) return false;` and ensure we are using the same logic in all places.  Note this one is a NOT fwUpdate_isDone().

kylemallory (changes requested):

kylemallory (approved):

waltjohnson (approved):

PR #500, opened 2023-11-15
Fix comManagerSend() parameter order and add comments

waltjohnson:
Fixed parameter order in comManagerSend() and added some comments.


kylemallory (approved):
Love the doxygen!  Thank you!

tmcd249 (approved):

PR #513, opened 2023-12-18
Sn 4463 separate period multiples

tmcd249:
(SDK) Separated the NMEA period multiple for RMC period multiple. 
(GPX) Fixed bugs where some GPX_STATUS and DEV_INFO values were being misreported.


src/protocol_nmea.cpp
```diff
@@ -2137,7 +2095,7 @@ uint32_t nmea_parse_ascb(int pHandle, const char msg[], int msgSize, rmci_t rmci
 	{
 		return 0;
 	}
-	
+	//
```
kylemallory, line 2098: Let's remove this...

src/protocol_nmea.cpp
```diff
@@ -2190,6 +2148,50 @@ uint32_t nmea_parse_ascb(int pHandle, const char msg[], int msgSize, rmci_t rmci
 uint32_t nmea_parse_asce(int pHandle, const char msg[], int msgSize, rmci_t rmci[NUM_COM_PORTS])
 {
 	(void)msgSize;
+
+	if(pHandle >= NUM_COM_PORTS)
+	{
+		return 0;
+	}
+	char *ptr = (char *)&msg[6];				// $ASCE
```
kylemallory, line 2156: I'm getting a little knit-picky here, but this is a bit messy with variable definitions mixed with conditions, etc mixed with curious white spacing choices, etc.  I think this, and the following lines of the rest of this function could probably benefit from some inline comments as well, to better communicate what its happening...

With the apparently non-trivial reworking of this (and the subsequent) functions, let's go ahead and add a doxygen block to these two functions as well.

src/test/test_protocol_nmea.cpp
```diff
@@ -25,22 +25,22 @@ TEST(protocol_nmea, nmea_parse_ascb)
     rmci_t rmci[NUM_COM_PORTS] = {};
     int port = 1;
     rmci_t &r = rmci[port];
-    r.periodMultiple[DID_INS_2] = 2;
-    r.periodMultiple[DID_PIMU] = 1;
-    r.periodMultiple[DID_GPS1_POS] = 1;
-    r.nmeaBits = 
+    r.rmcNmea.nmeaPeriod[NMEA_MSG_ID_PINS2] = '1';
```
kylemallory, line 28: This is curious.... setting the period to the character of the period multiple instead of the value itself?  I'm curious what warranted this change?

kylemallory (changes requested):

src/test/test_protocol_nmea.cpp
tmcd249: Easier to debug on failure since the ascii will always be readable in a terminal. No other reason than that. left it because there is no real draw back besides looking weird.

waltjohnson (approved):

kylemallory (approved):

PR #518, opened 2024-01-08
SN-4561 switch ci firmware check to md5

waltjohnson:
Revised md5 algorithm to not use dynamic memory allocation.  Added unit tests for md5.


src/test/test_md5.cpp
```diff
@@ -0,0 +1,111 @@
...
+{
+
+#if 1   // Repeat string with index count
+    string filename = "md5_test.txt";
+    // Create test file
+    ofstream myfile(filename);
+    for (int i=0; i<1024; i++)
+    {
+        myfile << "The quick brown fox jumped over the lazy dogs.\n";
+        myfile << "Hope Kyle doesn't see this! " << i << "\n";
```
kylemallory, line 74: I see it!!

src/util/md5.cpp
```diff
@@ -1,14 +1,263 @@
...
+ (a) += (b); \
+  }
+#define II(a, b, c, d, x, s, ac) { \
+ (a) += I ((b), (c), (d)) + (x) + (uint32_t)(ac); \
+ (a) = ROTATE_LEFT ((a), (s)); \
+ (a) += (b); \
+  }
+
+// MD5 initialization
+void MD5Init(MD5_CTX_t *context) {
```
kylemallory, line 67: We should be consistent in name style/convention; probably avoid function names that start with upper-case, and be consistent in use of underscores?

Since this is all new code, we should also use Doxygen blocks for each function.

kylemallory (changes requested):
This looks good.  Will there be a matching PR on the is-common repo, that uses this implemenation in the various device firmware?  Similarly, it looks like some of the function calls have changed, and the usage of the library is different.  This will need to be updated.

There was, somewhere, a union type to represent the MD5 hash in different ways that allows for simple comparisons, etc.  Did this get removed?  It was used in various places as well.

waltjohnson:
> This looks good. Will there be a matching PR on the is-common repo, that uses this implemenation in the various device firmware? Similarly, it looks like some of the function calls have changed, and the usage of the library is different. This will need to be updated.
> 
> There was, somewhere, a union type to represent the MD5 hash in different ways that allows for simple comparisons, etc. Did this get removed? It was used in various places as well.

The union is defined here:
https://github.com/inertialsense/inertial-sense-sdk/blob/SN-4561_switch_ci_firmware_check_to_md5/src/util/md5.h#L12-L17

We still have `md5_matches()` here:
https://github.com/inertialsense/inertial-sense-sdk/blob/SN-4561_switch_ci_firmware_check_to_md5/src/util/md5.h#L36C2-L36C2

tmcd249 (approved):

src/util/md5.cpp
```diff
@@ -1,131 +1,399 @@
...
+    unsigned int i, j;
+
+    for (i = 0, j = 0; j < len; i++, j += 4) {
+        output[i] = ((uint32_t)input[j]) | (((uint32_t)input[j+1]) << 8) |
+                    (((uint32_t)input[j+2]) << 16) | (((uint32_t)input[j+3]) << 24);
+    }
+}
+
+// Copies memory
+void MD5_memcpy(unsigned char *output, const unsigned char *input, unsigned int len) {
```
kylemallory, line 245: My bad for not catching this the first time around... do we need an MD5_memcpy? and and MD5_memset?  Aren't regular memset/memcpy sufficient?

src/util/md5.cpp
```diff
@@ -1,131 +1,399 @@
...
+    state[0] += a;
+    state[1] += b;
+    state[2] += c;
+    state[3] += d;
+
+    // Zeroize sensitive information
+    MD5_memset((unsigned char *)x, 0, sizeof(x));
+}
+
+// Encodes input (uint32_t) into output (unsigned char)
```
kylemallory, line 222: Since I'm requesting changes for the MD5_memset/MD5_memcpy stuff below, I might as well make the request for proper Doxygen blocks for all of these functions.  I wonder if we can find a VScode plugin to automatically create them, anytime a new function is written?  :thinking:

kylemallory (changes requested):
Sorry @waltjohnson I know you're tired of this.  Just a couple more requests! I promise!

PR #528, opened 2024-01-18
SN-4644 fix IMX 1x firmware update development

waltjohnson:
This fixes ability to update IMX-5 firmware from release r1.10.x and before.


kylemallory (approved):
Remember to look at the cpp-linter output, and see if any of the warnings/errors can be addressed; A clean code-base is a happy code-base.  :grinning:

PR #544, opened 2024-02-09
DEVELOPMENT NMEA asce update

tmcd249:
- (SDK) DID_NMEA_BCAST_PERIOD matches the ASCII message $ASCE instead of the deprecated $ASCB.


waltjohnson (approved):

kylemallory (approved):

kylemallory (changes requested):
Looks good, but it also looks like it broke the unit-tests.  Definitely need to address that.

kylemallory (approved):

PR #579, opened 2024-03-24
Restore EVB-2 project for SPI testing

waltjohnson:
This adds the EVB-2 project back so we have a way to test SPI with the v2 protocol.


kylemallory (changes requested):
I think I misunderstood our discussion about this yesterday.   While adding this all to the SDK seems... straight-forward?  I don't think we want this in the SDK.  Being that this is the public repo, I don't think we want this visible.  Maybe, but I'd probably need to be convinced.

I'm more keen to move all this into a new EVB-2 repo, that builds off of is-common and SDK, like GPX does.  Is that possible?

tmcd249 (approved):

waltjohnson:
> I think I misunderstood our discussion about this yesterday. While adding this all to the SDK seems... straight-forward? I don't think we want this in the SDK. Being that this is the public repo, I don't think we want this visible. Maybe, but I'd probably need to be convinced.
> 
> I'm more keen to move all this into a new EVB-2 repo, that builds off of is-common and SDK, like GPX does. Is that possible?

I think it is alright having it in the SDK for a few reasons:

1. Customers need hardware in order to use the EVB-2 project.  We've never had customers asking how to use this project.  Although it is not ideal to have in our public SDK, I don't see it cause any problems for us or customers.
2. There may be some corner cases for customers using SPI where we want to loan them an EVB-2 as a SPI dev kit.
3. It's less work for us.

src/ISFileManager.cpp
```diff
@@ -462,7 +462,7 @@ namespace ISFileManager {
     std::string CurrentWorkingDirectory()
     {
         char curdir[256] = { 0 };
-        if (_GETCWD(curdir, sizeof(curdir)) != nullptr)
+        if (_GETCWD(curdir, sizeof(curdir)) != 0)
```
kylemallory, line 465: Was there a reason to change this from nullptr?  Technically, nullptr is correct, since this may result in warnings about type casting.

src/ISUtilities.cpp
```diff
@@ -26,7 +26,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #endif
 
 #if PLATFORM_IS_EMBEDDED
-    #include "drivers/d_time.h"
+    #include "d_time.h"
```
kylemallory, line 29: Is this using a different d_time.h?  We should be including "drivers/d_time.h", since this removes ambiguity in the source file, and simplified the CMakeLists.txt.  Changing this back to just "d_time.h" may actually break builds elsewhere.

src/rtk_defines.h
```diff
@@ -183,7 +183,7 @@
 typedef uint32_t pthread_t;
 typedef uint32_t pthread_mutex_t;
 typedef int lock_t;
-typedef int DIR;
+// typedef int DIR;
```
kylemallory, line 186: Not sure the importance of DIR, but why would bringing back in the EVB source would need this removed?

PR #580, opened 2024-03-26
Provide a convenience function CurrentWorkingDirectory()

waltjohnson:
Add convenience function CurrentWorkingDirectory() that returns std::string.

### DIFF
--- a/src/ISFileManager.cpp
+++ b/src/ISFileManager.cpp
@@ -493,11 +493,7 @@ namespace ISFileManager {
         std::string realPath = path;
         // first check if path is absolute
         if (!isPathAbsolute(path)) {
-            char curdir[256];
-            if (_GETCWD(curdir, sizeof(curdir)) == nullptr)
-            {   // Handle error
-            }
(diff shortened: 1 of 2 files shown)


tmcd249 (approved):

src/ISFileManager.cpp
```diff
@@ -482,7 +494,7 @@ namespace ISFileManager {
         // first check if path is absolute
         if (!isPathAbsolute(path)) {
             char curdir[256];
-            if (_GETCWD(curdir, sizeof(curdir)))
+            if (_GETCWD(curdir, sizeof(curdir)) == nullptr)
```
kylemallory, line 497: We should just update this to use CurrentWorkingDirector() instead.

src/ISFileManager.cpp
```diff
@@ -455,6 +455,18 @@ namespace ISFileManager {
         return path;
     }
 
+    std::string CurrentWorkingDirectory()
```
kylemallory, line 458: Doxygen needed

kylemallory (changes requested):

PR #600, opened 2024-04-08
SN-5019 runtime tests

waltjohnson:
- Fix time stamp wrapping for NMEA GGA message.
- Prevent truncation errors when printing NMEA Degrees Minute GGA position format.
- Added unit tests to sweep $GnGGA and $PGPS parameters over functional ranges for more complete testing.
- Started runtime testing framework.


kylemallory (approved):

tmcd249 (approved):

cltool/src/cltool_main.cpp
```diff
@@ -544,6 +544,114 @@ static int cltool_createHost()
     return 0;
 }
 
+static int cltool_dataStreaming()
```
kylemallory, line 547: Doxygen block

src/protocol_nmea.cpp
```diff
@@ -332,13 +332,17 @@ char *ASCII_to_hours_minutes_seconds(int *hours, int *minutes, float *seconds, c
 	return ptr;
 }
 
-char *ASCII_to_TimeOfDayMs(uint32_t *timeOfWeekMs, char *ptr)
+char *ASCII_UtcTimeToGpsTowMs(uint32_t *gpsTimeOfWeekMs, utc_time_t *utcTime, char *ptr, int utcWeekday, uint32_t leapS)
```
kylemallory, line 335: Doxygen block

tests/CMakeLists.txt
kylemallory, line 1: Looks like this was moved from ./test to ./tests.  Should it have been?  Should the other files in ./test/* be moved as well?

tests/runtime/device_runtime_tests.cpp
```diff
@@ -0,0 +1,312 @@
...
+#include <sstream>
+#include <stdarg.h>
+#include "protocol_nmea.h"
+#include "device_runtime_tests.h"
+#include "ISFileManager.h"
+
+
+#define LOG_DIRECTORY   "realtime_logs"
+
+DeviceRuntimeTests::DeviceRuntimeTests()
```
kylemallory, line 14: While I generally don't feel as strongly about Doxygen stuff in our tests as I do in the mainline code, I do think its a good practice all the way around... So, I won't reject it, but I'd like to encourage all of us to add Doxygen for all code (since eventually, someone else will have to access/modify it, etc).

tests/test_data_utils.cpp
```diff
@@ -149,122 +253,177 @@ void GenerateGpsVel(test_message_t &msg, gps_vel_t &gps, int i, float f, bool in
...
     }
+
+    return false;
 }
 
-void GenerateUblox(test_message_t &msg, int i, float f)
+bool GenerateNMEA(test_message_t &msg, int i, float f)
 {
-    (void)f;
+    // if (timeIsSameAndSet(s_msgTimeMs.nmeaPImu, s_msgTimeMs.pimu))
```
kylemallory, line 299: Do we need this dead code?   If it needs keeping, it should include a comment explaining why its still here.  Otherwise, if it wasn't good enough to keep around the first time... let's kill it (again)!

tests/test_data_utils.cpp
```diff
@@ -149,122 +253,177 @@ void GenerateGpsVel(test_message_t &msg, gps_vel_t &gps, int i, float f, bool in
...
+
+    if (timeIsSameAndSet(s_msgTimeMs.nmeaGga, s_msgTimeMs.gpsPos))
+    {   
+        msg.pktSize = nmea_gga((char*)msg.comm.rxBuf.start, msg.comm.rxBuf.size, s_gpsPos);
+        msg.ptype = _PTYPE_NMEA;
+        return true;
+    }
 
-    msg.ptype = _PTYPE_UBLOX;
+    // if (timeIsSameAndSet(s_msgTimeMs.nmeaGpsPos, s_msgTimeMs.gpsPos))
```
kylemallory, line 321: Same dead code.

tests/test_protocol_nmea.cpp
```diff
@@ -18,6 +20,65 @@ using namespace std;
...
+    EXPECT_NEAR(g1.timeOfWeekMs, g2.timeOfWeekMs, 1);
+    for (int i=0; i<3; i++)
+    {
+        EXPECT_NEAR(g1.vel[i], g2.vel[i], 0.01);        // 10 cm
+    }
+    EXPECT_NEAR(g1.sAcc, g2.sAcc, 0.01);                // 10 cm/s
+    EXPECT_EQ(g1.status, g2.status);
+}
+
+// TEST(protocol_nmea, gps_to_utc_time_conversion)
```
kylemallory, line 61: Dead code?  Or test that wouldn't pass but eventually needs to be fixed?   If the later, a FIXME should be added with a comment; otherwise, let's kill the dead code.

tests/test_runtime_tests.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+	DeviceRuntimeTests RuntimeTest;
+	uint8_t stream[BUFFER_SIZE] = {0};
+	int streamSize = GenerateDataStream(stream, sizeof(stream));
+	run_realtime_test(RuntimeTest, stream, streamSize);
+	ASSERT_EQ( RuntimeTest.ErrorCount(), 0 );
+}
+#endif
+
+#if 1
```
kylemallory, line 78: Lose the "if #1"; either put it behind a real macro/define (like "RUN_EXTENDED_REALTIME_TESTS") or just remove it entirely.

tests/test_time_conversion.cpp
```diff
@@ -1,19 +1,71 @@
...
+    int leapS = C_GPS_LEAP_SECONDS;
+
+    SetUtcTimeZone();
+
+    // Cycle through entire range of time of week
+    for (gpsTowMs = 0; gpsTowMs < C_MILLISECONDS_PER_WEEK; gpsTowMs += 200)
+    {
+        std::tm utcTime = stdGpsTimeToUtcDateTime(gpsTowMs/1000, gpsWeek, leapS);
+        uint32_t msec = gpsTowMs%1000;
+#if 0
```
kylemallory, line 22: Dead code

kylemallory (changes requested):

tests/test_data_utils.cpp
waltjohnson: Yes, we do want that code.  Thanks.

PR #602, opened 2024-04-09
NMEA GSV sentence now uses i32 for Ele/Azi values.

kylemallory:
NMEA GSV sentence exludes sats < 0 elevation.
NMEA GGA 'fixQuality' reinforced to reduce likelihood of errant values.

### DIFF
--- a/src/protocol_nmea.cpp
+++ b/src/protocol_nmea.cpp
@@ -75,6 +75,23 @@ void nmea_print_u32(char buf[], int bufSize, int &offset, int precision, uint32_
 	}
 }
 
+void nmea_print_i32(char buf[], int bufSize, int &offset, int precision, uint32_t value)
+{
+    bufSize -= offset;
+    if (bufSize<=0) return;		// Prevent snprintf w/ invalid size
+    buf += offset;
+
+    if (value)
+    {	// Non-zero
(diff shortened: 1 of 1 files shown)


tmcd249 (approved):

src/protocol_nmea.cpp
```diff
@@ -1386,17 +1403,17 @@ int nmea_gsv_group(char a[], int aSize, int &offset, gps_sat_t &gsat, gps_sig_t
 				for (uint32_t j=0; j<=gsat.numSats; j++)
 				{
 					gps_sat_sv_t &sat = gsat.sat[j];
-					if (sat.gnssId == sig.gnssId &&
-						sat.svId == sig.svId)
+					if ((sat.gnssId == sig.gnssId) && (sat.svId == sig.svId)
+                        && (sat.elev >= 0)) // TODO: We should probably remove this, implement a parameter in the GPX flash config, and configure on the CXD5610.
```
waltjohnson, line 1407: You switched the elevation print below to use nmea_print_i32() which will correctly display negative elevation angles.   But then you limit the elevation angles to only non-negative angles.  Are we sure we don't want to print negative angles?

waltjohnson (approved):

PR #609, opened 2024-04-17
SN-4452 evaltool set otp manf info

waltjohnson:
Add hardware detection check to BIT basic test to ensure correct firmware is loaded and used.

### DIFF
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -115,7 +115,7 @@ bool ISFirmwareUpdater::fwUpdate_handleVersionResponse(const fwUpdate::payload_t
     }
 
     remoteDevInfo.serialNumber = msg.data.version_resp.serialNumber;
-    remoteDevInfo.hardware = msg.data.version_resp.hardwareId;
+    remoteDevInfo.hardwareType = msg.data.version_resp.hardwareType;
     memcpy(remoteDevInfo.hardwareVer, msg.data.version_resp.hardwareVer, 4);
     memcpy(remoteDevInfo.firmwareVer, msg.data.version_resp.firmwareVer, 4);
     remoteDevInfo.buildYear = msg.data.version_resp.buildYear;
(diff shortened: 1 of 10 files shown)


src/ISFirmwareUpdater.cpp
```diff
@@ -129,9 +129,9 @@ bool ISFirmwareUpdater::fwUpdate_handleVersionResponse(const fwUpdate::payload_t
     target_devInfo = &remoteDevInfo;
 
     if(pfnInfoProgress_cb != nullptr) {
-        if ((remoteDevInfo.hardware >= HDW_TYPE__UINS) && (remoteDevInfo.hardware <= HDW_TYPE__GPX)) {
+        if ((remoteDevInfo.hardwareType >= DEV_INFO_HARDWARE_TYPE_UINS) && (remoteDevInfo.hardwareType <= DEV_INFO_HARDWARE_TYPE_GPX)) {
             const char *hdw_names[5] = {"UNKNOWN", "UINS", "EVB", "IMX", "GPX"};
```
kylemallory, line 133: You didn't modify this line... but while we are in here, doing this hardwareType, we should move this line into global const in data_sets.h that defines the string constants for each hardware type (IMX, GPX, etc).  We have these string duplicated lots of places, and could simplify and be consistent elsewhere.  ie, in ISDisplay.cpp above, for example, by replacing the switch with a lookup of "g_constHdwTypeNames[info.hardwareType]".

src/ISFirmwareUpdater.cpp
```diff
@@ -519,20 +519,20 @@ void ISFirmwareUpdater::runCommand(std::string cmd) {
             // any target which doesn't report version info will also expect the old MD5 digest
             if (!target_devInfo) {
                 // TODO: We should be able to remove most of this after 2.1.0 has been released
-                if ( ((target & fwUpdate::TARGET_IMX5) && (devInfo->hardware == HDW_TYPE__IMX)) ||
-                     ((target & fwUpdate::TARGET_GPX1) && (devInfo->hardware == HDW_TYPE__GPX)) ) {
+                if ( ((target & fwUpdate::TARGET_IMX5) && (devInfo->hardwareType == DEV_INFO_HARDWARE_TYPE_IMX)) ||
```
kylemallory, line 522: Since we are consolidating these, is there still a reason to denote "DEV_INFO_HARDWARE_TYPE_IMX" as opposed to just "HARDWARE_TYPE_IMX"?    Personally, I'm a fan of the more abbreviated HDW_TYPE__GPX, but don't think we need the "DEV_INFO" since its not specific do DEV_INFO anymore.

kylemallory (approved):
No deal blockers, but I do think my few comments are worth addressing if you don't mind taking the time to do them (should be minor).

kylemallory (changes requested):
I had some comments for suggested changes, but still approved.  After reviewing the is-common PR, I think these should be changed to match the effort there, so I've requested changes here too.

tmcd249 (approved):

kylemallory (approved):

PR #625, opened 2024-05-23
Sn 5247 rtcm in event

tmcd249:
- (SDK) RTMC3 protocol added to DID_EVENT. 
- Sony Binary protocol added to DID_EVENT.


kylemallory (changes requested):
For the sake of consistency, would you change this (and the subsequent DID_EventPriority enums) to be ALL_CAPS and consistent in style (ie DID_EVENT_PROTOCOL_RAW), to all the other Enums that are defined in data_sets.h?

You should be able to perform a "Refactor" operation to rename it here, and have all the usages of that throughout the code automatically renamed.

kylemallory (approved):

waltjohnson (approved):

PR #637, opened 2024-06-03
Sn 5230 connect to invalid port

andrewpriddis:
Added WaitForSync to make sure code not advancing with nothing communicating. Changed order of stream config to after advertisers so callback and publishers not called before publishers created.


kylemallory (approved):

waltjohnson (approved):

andrewpriddis:
Added some wait for sync code. Rearranged configuration to start streaming data after the publishers were created.

PR #644, opened 2024-06-07
Sn 5299 ephemeris counter

aabogdanov:
Added ephemeris counter, observation available from rover and base, unhealthy satellite counter


tmcd249 (approved):

waltjohnson (approved):

src/data_sets.h
```diff
@@ -3481,6 +3481,13 @@ typedef struct PACKED
     uint8_t obs_pairs_used;
     uint8_t raw_ptr_queue_overrun;
     uint8_t raw_dat_queue_overrun;
+
+    uint8_t obs_rover_avail;
```
kylemallory, line 3485: Would you mind adding descriptions `//! this variable represents...`-type comments that explaind what these mean?  Its not clear to me what "available" means, vs unhealthy, and also `obs_eph_avail` vs base/rover.

kylemallory (changes requested):

kylemallory (approved):
Thanks Alex, those comments are very helpful!

PR #648, opened 2024-06-12
SN-5332 Automatic Python Virtual Enviroments

FiriusFoxx:
This should make it so that when building Log Inspector or installing Log Inspector's dependencies it will automatically locate or create a virtual environment if one can't be found

It also adds .venv to .gitignore so no one accidentally uploads a Python virtual environment to GitHub

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,7 @@ CMakeFiles/
 *.user
 .vs
 .vscode
-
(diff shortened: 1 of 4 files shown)


kylemallory (approved):
Nice!

tmcd249 (approved):

PR #675, opened 2024-07-10
SN-5441 flash config sync platform

waltjohnson:
Fixed issue where flash config sync would not update checksum correctly and perform unnecessary flash sync if platformConfig type was unchanged.  Ignore bit PLATFORM_CFG_UPDATE_IO_CONFIG when checking whether to upload flash config platformConfig from InertialSense class.

### DIFF
--- a/src/imx_defaults.c
+++ b/src/imx_defaults.c
@@ -4,13 +4,13 @@
 
 
 // Return 0 if not valid.  Zero is not a valid platform type.
-int platformConfigTypeValid(uint32_t platformConfig)
+int imxPlatformConfigTypeValid(uint32_t platformConfig)
 {
     uint32_t platformType = platformConfig&PLATFORM_CFG_TYPE_MASK;
     return (platformType > 0) && (platformType < PLATFORM_CFG_TYPE_COUNT);
 }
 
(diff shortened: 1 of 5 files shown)


src/imx_defaults.c
```diff
@@ -0,0 +1,329 @@
+#include "imx_defaults.h"
+
+
+
+
+// Return 0 if not valid.  Zero is not a valid platform type.
+int imxPlatformConfigTypeValid(uint32_t platformConfig)
+{
+    uint32_t platformType = platformConfig&PLATFORM_CFG_TYPE_MASK;
+    return (platformType > 0) && (platformType < PLATFORM_CFG_TYPE_COUNT);
```
kylemallory, line 10: I can't see these macros in the PR, but is this right?  the platformType being < PLATFORM_CFG_TYPE_COUNT only applies it the TYPE MASK is in the lowest order bits (0-n), otherwise it may never be true.

kylemallory (approved):

src/imx_defaults.c
waltjohnson: This code just ensures that the platformType which is a subset in the platformConfig is greater than 0 and less than the max value for platformType.  Looks correct to me.

PR #694, opened 2024-08-18
Switch evaltool to use ISDataMappings

waltjohnson:
Move data set definitions used in EvalTool from ParcelHelper to ISDataMappings.


tmcd249 (approved):

kylemallory (changes requested):
Minor stuff...

src/ISDataMappings.cpp
```diff
@@ -230,407 +90,396 @@ static void PopulateTimestampField(uint32_t id, const data_info_t** timestamps,
...
-    ADD_MAP(m, totalSize, "command", command, 0, DataTypeUInt8, uint8_t, 0);
-    ADD_MAP(m, totalSize, "port", port, 0, DataTypeUInt8, uint8_t, 0);
-    ADD_MAP(m, totalSize, "testMode", testMode, 0, DataTypeUInt8, uint8_t, 0);
-    ADD_MAP(m, totalSize, "state", state, 0, DataTypeUInt8, uint8_t, 0);
-    ADD_MAP(m, totalSize, "detectedHardwareId", detectedHardwareId, 0, DataTypeUInt16, uint16_t, 0);
-    ADD_MAP(m, totalSize, "reserved[0]", reserved[0], 0, DataTypeUInt8, uint8_t&, 0);
-    ADD_MAP(m, totalSize, "reserved[1]", reserved[1], 0, DataTypeUInt8, uint8_t&, 0);
-
-    ASSERT_SIZE(totalSize);
+#if 0
```
kylemallory, line 119: This is a big block of dead code.  Do we need it?

src/ISDataMappings.cpp
```diff
@@ -230,407 +90,396 @@ static void PopulateTimestampField(uint32_t id, const data_info_t** timestamps,
...
+    mapper.AddMember("g4", &system_fault_t::g4, DATA_TYPE_UINT32, "", "value at fault", flags);
+    mapper.AddMember("g5Lr", &system_fault_t::g5Lr, DATA_TYPE_UINT32, "", "Load register at fault", flags);
+    mapper.AddMember("pc", &system_fault_t::pc, DATA_TYPE_UINT32, "", "program counter at fault", flags);
+    mapper.AddMember("psr", &system_fault_t::psr, DATA_TYPE_UINT32, "", "program status register at fault", flags);
+}
+
+void PopulatePortMonitorMappings(map_name_to_info_t mappings[DID_COUNT], uint32_t lookupSize[DID_COUNT], map_index_to_info_t indices[DID_COUNT], uint32_t did)
+{
+    INIT_MAP(port_monitor_t, did);
+    ADD_MAP_7("activePorts", activePorts, DATA_TYPE_UINT8, uint8_t, "", "Number of active ports", DATA_FLAGS_READ_ONLY);
```
kylemallory, line 220: I thought you have changed these to the mapper.AddMember() template.  These appear to be arrays.  Are arrays not supported in the template?  I think I remember you saying something about that.

kylemallory (changes requested):
I think we should allocate some time to fix this right.  I know its not a "big priority" but in its current state, this appears only half-baked.  I think between you, Tony and I we could probably get this finished in about 30m to an hour, unless I'm missing some bigger issue.

PR #697, opened 2024-08-23
Add option to cltool to disable device verification

waltjohnson:
- Fixes cltool -sysCmd send reliability issue
- Adds disable device verification option -vd.


tmcd249 (approved):

kylemallory (changes requested):
Looks like this is already included in the other PR.  Should be fine, but we can probably close this PR. :man_shrugging:

PR #698, opened 2024-08-24
Code migration from ROS 1 to ROS 2.

sbogdanov2:
Converted ROS 1 syntax to ROS 2, including messages, console logs, and parameters. Also worked on ROS 2 bridge tests.


.gitignore
```diff
@@ -31,4 +31,8 @@ libInertialSenseSDK.a
 scripts/.venv/
 python/venv/
 python/logInspector/.eggs/
-
+.cmake/
+/install/
+/log/
+/ros2/install/
+/ros2/log/
```
kylemallory, line 38: Should also add .ninja to the .gitnore, and while we're at it, remove the two .ninja files below.  They do not belong in the PR/repo.

.ninja_deps
kylemallory, line 1: Remove this.

.ninja_log
kylemallory, line 1: Remove this.

Testing/Temporary/LastTest.log
kylemallory, line 1: These should be removed as well.

build.ninja
kylemallory, line 1: Do we need this file too? This says its generated, in which case we should not need to keep it.  Ie, ANY generated files should not be committed (should be added to .gitignore), since they will just be regenerated with each build.

Is ninja a requirement for ROS only, or did these come from somewhere else, like the IDE?  If so required for ROS, we should probably move it into the ROS2 folder.  Otherwise, let's discuss.

ros2/File.txt
kylemallory, line 1: Remove

ros2/File_errors1.txt
kylemallory, line 1: Remove all of these File_errors*.txt

ros2/include/ParamHelper.h
kylemallory, line 1: A lot of these files are copied from the ROS1 code.  How much has changed between them, for the conversion to ROS2?  If its significant, we can keep them.  If its minor, or not at all, we should look at organizing them so they can be shared between both versions.

The thing we ultimately want to avoid is having to make identical changes in both, whenever we fix something in the future that applies to both.  So, let's do an IDE comparison between this new file and the original ROS1 version, and see what the correct course of action here should be.

kylemallory (changes requested):
Savva, this looks pretty good, but there are a lot of files that need to be removed from the PR, and deleted from the Git repo.

Also, we should look to see how we can consolidate the code that is shared/common/similar between ROS1 and ROS2.  Ideally, only changes that are unique to ROS2 should be here.

If necessary, we may want to try and back-port some of the ROS2 changes into ROS1, where possible, to make sure they play nice with both teams.  For example, I know there were some case-issues early on; can we use the ROS2 case-specific names, etc in ROS1?  There is probably some level of cross compatibility. If not, that is fine, but let's explore it.

ROS/include/RtkBase.h
```diff
@@ -0,0 +1,144 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ *
+ ***************************************************************************************/
+
+#ifndef INERTIAL_SENSE_IMX_RTKBASE_H
+#define INERTIAL_SENSE_IMX_RTKBASE_H
+
+#ifdef ROS2
+#include "rclcpp/rclcpp/rclcpp.hpp"
+#endif
```
kylemallory, line 24: Use an #elif defined(ROS1) or just an #else to catch the "ROS1" version below.

ROS/include/RtkBase.h
```diff
@@ -0,0 +1,144 @@
...
+#ifdef ROS1
+#include "ros/ros.h"
+#endif
+#include "ParamHelper.h"
+
+
+class RtkBaseCorrectionProvider {
+protected:
+    ParamHelper ph_;
+#ifdef ROS2
```
kylemallory, line 35: Is there no ROS1 equivelant nh_?

.cmake/api/v1/query/cache-v2
kylemallory, line 1: I see that .cmake is already in the .gitignore, but we still have these files being added in the PR.  They need to be removed/deleted from your branch and pushed up.  You will probably need to remove `.cmake/` from the .gitignore, then delete and push the deletes up, then add `.cmake/` back into .gitignore, and then finally commit/push .gitignore to the repo.

ROS/ros/include/TopicHelper.h
kylemallory, line 1: Looks like this is included twice still, but identical in both locations.

ROS/ros/include/ParamHelper.h
kylemallory, line 1: Looks like this should be in ROS/include, rather than ROS/ros/include.  Also, if possible, lets rename ROS/ros to ROS/ros1 so that its clear this is the ROS1-specific code.

kylemallory (commented):
This is a good start to consolidating common code elements, by moving shared code out into ROS/include.  We should also have a ROS/src, where common source is included (things like ParamHelper.cpp, etc).

waltjohnson (approved):

kylemallory (approved):

tmcd249 (approved):

PR #702, opened 2024-08-25
SN-5601 - cltool "-did <did>=0" stalls occasionally

kylemallory:
Specifically, this changes internal cltool logic to keep track of requested DIDs and re-request them if they are not received within an expected timeframe (based on the requested periodMultiple).

- (cltool) Fixed stalls when querying single DID with no response (implement default 30-second timeout, overridden using -dur option).
- (cltool) -stats cmd argument now correctly displays DID stats
- (cltool) Fixed errant display when "onceDid" (-did <did>=0) is set, for both PRINT_PRETTY and PRINT_SCROLL.
- (cltool) Miscellaneous, minor improvements to cltool run/display logic
- (SDK) implements "serialPortDrain()" function to block until pending serial-port writes have been transmitted.


cltool/src/cltool_main.cpp
```diff
@@ -664,8 +700,9 @@ static int cltool_createHost()
     cout << "Shutting down..." << endl;
 
     // close the interface cleanly, this ensures serial port and any logging are shutdown properly
-    inertialSenseInterface.Close();
-    inertialSenseInterface.CloseServerConnection();
+    // Note the following two lines are called by the InertialSense destructor, which will be called when this function exits
+    // inertialSenseInterface.Close();
+    // inertialSenseInterface.CloseServerConnection();
```
waltjohnson, line 705: @kylemallory Should we just remove this commented out code?  We can still add a comment to the following:

We don't need to close the InertialSense class intererface here because it the InertialSense desctructor.

waltjohnson (approved):

tmcd249 (approved):

PR #705, opened 2024-08-27
SN-5612 Fixed multiple cltool issues

waltjohnson:
- Fix `-did x` option stops streaming when message was previously streaming. 
  - Prevent reception of stray data by disabling cltool_dataCallback() until stop broadcast in cltool_setupCommunications() is called. This prevents the cltool from processing any data received prior to cltool_setupCommunications().
  - Simplify data request to be more robust, re-requesting data once a second. 
- Fix `-edit` option not displaying.  Copy data into m_editData.pData.pDataBuf to ensure large data sets at high rates do not get mangled when print to display. 
- Fix `-edit` option variable selection.  
- Reformat print of DID_GPX_STATUS state variables for readability.

### DIFF
--- a/src/ISComm.c
+++ b/src/ISComm.c
@@ -1152,6 +1152,20 @@ char copyDataBufPToStructP(void *sptr, const p_data_buf_t *data, const unsigned 
     }
 }
 
+char copyDataPToDataP(p_data_t *dst, const p_data_t *src, const unsigned int maxsize)
+{
+    if (src->hdr.size <= maxsize)
+    {
+        dst->hdr = src->hdr;
+        memcpy(dst->ptr, src->ptr, src->hdr.size);
+        return 0;
+    }
+    else
+    {
+        return -1;
+    }
+}
+
 /** Copies packet data into a data structure.  Returns 0 on success, -1 on failure. */
 char copyDataPToStructP2(void *sptr, const p_data_hdr_t *dataHdr, const uint8_t *dataBuf, const unsigned int maxsize)
 {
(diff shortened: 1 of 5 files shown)


tmcd249 (approved):

src/ISComm.c
```diff
@@ -1152,6 +1152,20 @@ char copyDataBufPToStructP(void *sptr, const p_data_buf_t *data, const unsigned
     }
 }
 
+char copyDataPToDataP(p_data_t *dst, const p_data_t *src, const unsigned int maxsize)
```
kylemallory, line 1155: Needs a doxygen.  Also, doesn't this function exist elsewhere?  Or has it been moved here?

src/ISDisplay.cpp
```diff
@@ -653,7 +657,8 @@ string cInertialSenseDisplay::DataToString(const p_data_t* data)
             str = DataToStringRawHex((const char *)data->ptr, data->hdr, 32);
 		else if (m_editData.did == data->hdr.id)	
 		{	// Default view
-			str = DatasetToString(m_editData.pData = data);
+			copyDataPToDataP(&m_editData.pData, data, MAX_DATASET_SIZE);
```
kylemallory, line 660: We discussed this, but I didn't see the code above on line 477.  Why are we doing the call to copyDataPToDataP() twice?  Can we modify 477 to be true, if either m_editData.did OR m_outputOnceDid matches the hdr.id?  Or does that not jive with the PRETTY_PRINT on line 526?   It seems like in EVERY CASE, we could/should copy the data into m_editData.pData, and then display based on that.

kylemallory (changes requested):

src/ISDisplay.cpp
waltjohnson: copyDataPToDataP() is new helper function because we've never copied from p_data_t to p_data_t.

You are correct in that the second instance of copyDataPToDataP() was extra and not needed.  I removed the extra data copy and retested the different display modes.

PR #707, opened 2024-08-28
SN-5612 cltool debug

waltjohnson:
Add runtime and count for tx packet, rx packet, rx byte to cltool.


cltool/setup_debug_acu.sh
kylemallory, line 1: This shouldn't be in the public Repo, since its customer-specific

cltool/src/cltool_main.cpp
```diff
@@ -264,7 +264,7 @@ static void cltool_dataCallback(InertialSense* i, p_data_t* data, int pHandle)
         cout.flush();
     else if (!g_inertialSenseDisplay.ExitProgram()) // don't process any additional data once we've been told to exit
         g_inertialSenseDisplay.ProcessData(data);
-}
+    }
```
kylemallory, line 267: Errant whitespace?

cltool/src/cltool_main.cpp
```diff
@@ -714,6 +714,12 @@ static int cltool_dataStreaming()
         cout << "Failed to open serial port at " << g_commandLineOptions.comPort.c_str() << endl;
         return -1;    // Failed to open serial port
     }
+    g_inertialSenseDisplay.SetSerialPort(inertialSenseInterface.SerialPort());
```
kylemallory, line 717: This seems like a strange place (`cltool_requestDataSets()`) to be doing ISDisplay setup.  Should be be moved into `cltool_streamData()` or `cltool_setupCommunications()`?

src/ISDataMappings.cpp
```diff
@@ -2975,6 +2975,14 @@ uint32_t cISDataMappings::DefaultPeriodMultiple(uint32_t dataId)
     case DID_INL2_STATES:
     case DID_ROS_COVARIANCE_POSE_TWIST:
     case DID_INL2_MAG_OBS_INFO:
+    case DID_GPX_DEV_INFO:
```
kylemallory, line 2978: Nice!

src/ISDisplay.cpp
```diff
@@ -225,19 +225,41 @@ void cInertialSenseDisplay::GoToColumnAndRow(int x, int y)
...
+	{
+		m_startMs = current_timeMs();
+	}
+
+	// cltool runtime
+	double runtime = 0.001 * (current_timeMs() - m_startMs);
+
+	std::ostringstream stream;
+
+	stream << "$ Inertial Sense.  Connected.  CTRL-C to terminate.  ";
```
kylemallory, line 243: Its not a big deal, and it was this way before your work, but we should try and consolidate these messages/string.  Would a `generateHeader(bool connected)` make more sense, and then have the resulting string be a concatenation of these fields, if we're connected (or not)?

src/ISDisplay.h
```diff
@@ -138,6 +139,8 @@ class cInertialSenseDisplay
 	bool UploadNeeded() { bool uploadNeeded = m_editData.uploadNeeded; m_editData.uploadNeeded = false; return uploadNeeded; };
 	edit_data_t *EditData() { return &m_editData; }
 	void setOutputOnceDid(int did) { m_outputOnceDid = did; m_interactiveMode = m_outputOnceDid == 0; }
+	void SetSerialPort(serial_port_t* port) { m_port = port; }
```
kylemallory, line 142: How does this work when connected to multiple devices (ie `-c *`)?  How does normal/current display stuff work in that situation?

kylemallory (changes requested):

tmcd249 (approved):

cltool/src/cltool_main.cpp
waltjohnson: Moved into cltool_setupCommunications()

src/ISDisplay.cpp
waltjohnson: Done

src/ISDisplay.h
waltjohnson: It only shows stats for the first device.

PR #708, opened 2024-08-30
Fixes to Pass ROS bridge tests

andrewpriddis:
- **_Disabled_** the gps time sync test for now because of non-determinist characteristics. This will be fixed in a future issue.
- Modified the workflow so tests are called in standard ROS way.
- Added protection for a false connection to unit. 
- Fixed a vector out of bounds access (.end() used in place of .back())
- Added protection around calling non-defined pusblishers.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -169,4 +169,4 @@ jobs:
           source devel/setup.bash
           roscore &
           catkin_make run_tests_inertial_sense_ros_gtest_test_ros_bridge
-          catkin_test_results || (rostest --text inertial_sense_ros test_ros_bridge.launch && catkin_test_results && false)
+          catkin_test_results
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -102,7 +102,7 @@ jobs:
   ros-bridge-tests:
     name: "ROS Bridge Tests"
     runs-on: [self-hosted, Linux, X64, docker, focal]
-    if: false  # disable this job
+    # if: false  # disable this job
 
     env:
       ROS_CI_DESKTOP: "`lsb_release -cs`"  # e.g. [trusty|xenial|...]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -167,6 +167,6 @@ jobs:
       - name: Run Functional/Integration Tests (via hardware)
         run: |
           source devel/setup.bash
-          rostest inertial_sense_ros test_ros_bridge.launch
+          roscore &
+          catkin_make run_tests_inertial_sense_ros_gtest_test_ros_bridge
           catkin_test_results || (rostest --text inertial_sense_ros test_ros_bridge.launch && catkin_test_results && false)
-
--- a/ros/src/inertial_sense_ros.cpp
+++ b/ros/src/inertial_sense_ros.cpp
@@ -1685,12 +1685,14 @@ void InertialSenseROS::RTK_Misc_callback(eDataIDs DID, const gps_rtk_misc_t *con
     {
     case DID_GPS1_RTK_POS_MISC:
         rs_.rtk_pos.streamingCheck(DID);
-        rs_.rtk_pos.pubInfo.publish(rtk_info);
+        if (rs_.rtk_pos.pubInfo.getNumSubscribers() > 0)
+            rs_.rtk_pos.pubInfo.publish(rtk_info);
         break;
 
     case DID_GPS2_RTK_CMP_MISC:
         rs_.rtk_cmp.streamingCheck(DID);
-        rs_.rtk_cmp.pubInfo.publish(rtk_info);
+        if (rs_.rtk_cmp.pubInfo.getNumSubscribers() > 0)
+            rs_.rtk_cmp.pubInfo.publish(rtk_info);
         break;
     }
 }
--- a/ros/src/inertial_sense_ros.cpp
+++ b/ros/src/inertial_sense_ros.cpp
@@ -552,7 +552,7 @@ bool InertialSenseROS::connect(float timeout)
             ROS_ERROR("InertialSenseROS: Unable to open serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
             usleep(500000); // is this a good idea?
         }
-        else if (serialNum == "0")
+        else if (serialNum.length() < 5)
         {
             ROS_ERROR("InertialSenseROS: Unable to open serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
             sdk_connected_ = false;
--- a/ros/src/inertial_sense_ros.cpp
+++ b/ros/src/inertial_sense_ros.cpp
@@ -552,7 +552,7 @@ bool InertialSenseROS::connect(float timeout)
             ROS_ERROR("InertialSenseROS: Unable to open serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
             usleep(500000); // is this a good idea?
         }
-        else if (serialNum.length() < 5)
+        else if (serialNum == "0")
         {
             ROS_ERROR("InertialSenseROS: Unable to open serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
             sdk_connected_ = false;
--- a/ros/src/inertial_sense_ros.cpp
+++ b/ros/src/inertial_sense_ros.cpp
@@ -557,7 +557,6 @@ bool InertialSenseROS::connect(float timeout)
             ROS_ERROR("InertialSenseROS: False connection \"%s\", at %d baud", cur_port.c_str(), baudrate_);
             sdk_connected_ = false;
         }
-         
         else 
         {
             ROS_INFO("InertialSenseROS: Connected to IMX SN%d on \"%s\", at %d baud", IS_.DeviceInfo().serialNumber, cur_port.c_str(), baudrate_);
@@ -1206,8 +1205,6 @@ void InertialSenseROS::INS4_callback(eDataIDs DID, const ins_4_t *const msg)
                     rs_.odom_ins_ned.pub.publish(msg_odom_ned);
 
                 }
-                
-                
                 if (publishTf_)
                 {
                     // Calculate the TF from the pose...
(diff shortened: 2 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -167,6 +167,6 @@ jobs:
       - name: Run Functional/Integration Tests (via hardware)
         run: |
           source devel/setup.bash
-          rostest inertial_sense_ros test_ros_bridge.launch
-          catkin_test_results || (rostest --text inertial_sense_ros test_ros_bridge.launch && catkin_test_results && false)
-
+          roscore &
```
kylemallory, line 170: I don't know that this is safe... rostest is supposed to launch the roscore, and then terminate it when its finished.  Was there a specific reason for doing this way?

ros/src/inertial_sense_ros.cpp
```diff
@@ -535,21 +535,36 @@ bool InertialSenseROS::connect(float timeout)
...
+        {
+            serialNum = IS_.DeviceInfo().serialNumber;
+        }
+        
+        if (!sdk_connected_) 
+        {
             ROS_ERROR("InertialSenseROS: Unable to open serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
-            sleep(1); // is this a good idea?
-        } else {
+            usleep(500000); // is this a good idea?
```
kylemallory, line 553: Is it a good idea?  You're sleeping for... 0.5ms?  It looks like it is scanning through available serial ports, so probably not the worst.  Ideally though, you wouldn't wait, you'd just jump directly to the next port.

It feels like there maybe a better way to write this, where you can iterate through each port, and on failure call `continue;` to jump directly to the next port.   Probably makes the most sense to do a `while()` for the timeout, and then to a `for (auto cur_port : ports_) { ... }` to iterate through all the ports.

This will give the benefit of ensuring that all ports get tried at least once, before it gives up.  As its written, it could timeout before it checks all the ports in the `ports_` list.

ros/src/inertial_sense_ros.cpp
```diff
@@ -535,21 +535,36 @@ bool InertialSenseROS::connect(float timeout)
...
         /// Connect to the IMX
         ROS_INFO("InertialSenseROS: Connecting to serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
         sdk_connected_ = IS_.Open(cur_port.c_str(), baudrate_);
-        if (!sdk_connected_) {
+        if (sdk_connected_)
+        {
+            serialNum = IS_.DeviceInfo().serialNumber;
+        }
+        
+        if (!sdk_connected_)
```
kylemallory, line 550: This logic feels fishy.  This should probably be an `} else {` of the previous if, since its the only other possible result of `sdk_connected_`).
Then the following `if (serialNum == 0)` should be inside the 'if connected" stanza.  If you are NOT connected, because `IS_.Open()` failed, the only thing to do is print the error, and move on.

ros/src/inertial_sense_ros.cpp
```diff
@@ -535,21 +535,36 @@ bool InertialSenseROS::connect(float timeout)
...
-            sleep(1); // is this a good idea?
-        } else {
+            usleep(500000); // is this a good idea?
+        }
+        else if (serialNum == 0)
+        {
+            ROS_ERROR("InertialSenseROS: False connection \"%s\", at %d baud", cur_port.c_str(), baudrate_);
+            sdk_connected_ = false;
+        }
+
```
kylemallory, line 560: Remove empty line (we're trying to be better about excessive whitespace)

ros/src/inertial_sense_ros.cpp
```diff
@@ -1168,8 +1201,13 @@ void InertialSenseROS::INS4_callback(eDataIDs DID, const ins_4_t *const msg)
...
                 msg_odom_ned.twist.twist.angular.y = result[1];
                 msg_odom_ned.twist.twist.angular.z = result[2];
-                rs_.odom_ins_ned.pub.publish(msg_odom_ned);
+                if (rs_.odom_ins_ned.pub)
+                {
+                    rs_.odom_ins_ned.pub.publish(msg_odom_ned);
 
+                }
+                
+
```
kylemallory, line 1210: Let's keep for formatting consistent.  if its one empty line after a closing brace everywhere else in the source file, we should keep it to one empty line after a closing brace here too.

ros/src/inertial_sense_ros.cpp
```diff
@@ -1607,7 +1682,12 @@ void InertialSenseROS::baro_callback(eDataIDs DID, const barometer_t *const msg)
     baro_msg.header.frame_id = frame_id_;
     baro_msg.fluid_pressure = msg->bar;
     baro_msg.variance = msg->barTemp;
-    rs_.barometer.pub.publish(baro_msg);
+    if (rs_.barometer.pub)
+    {
+        rs_.barometer.pub.publish(baro_msg);
+    }
+
```
kylemallory, line 1689: Remove these lines as well

ros/src/inertial_sense_ros.cpp
```diff
@@ -1865,7 +1966,16 @@ void InertialSenseROS::GPS_obs_bundle_timer_callback(const ros::TimerEvent &e)
...
-            rs_.gps2_raw.pubObs.publish(gps2_obs_Vec_);
+            if (rs_.gps2_raw.pubObs)
+            {
+                rs_.gps2_raw.pubObs.publish(gps2_obs_Vec_);
+            }
+            if (rs_.gps2_raw.pubObs)
+            {
+                rs_.gps2_raw.pubObs.publish(gps2_obs_Vec_);
+            }
+
```
kylemallory, line 1977: More lines to remove

ros/src/inertial_sense_ros.cpp
```diff
@@ -1959,9 +2089,24 @@ void InertialSenseROS::GPS_geph_callback(eDataIDs DID, const geph_t *const msg)
     geph.dtaun = msg->dtaun;
     switch (DID)
     { 
-    case DID_GPS1_RAW:      rs_.gps1_raw.pubGEp.publish(geph);       break;
-    case DID_GPS2_RAW:      rs_.gps2_raw.pubGEp.publish(geph);       break;
-    case DID_GPS_BASE_RAW:  rs_.gpsbase_raw.pubGEp.publish(geph);   break;
+    case DID_GPS1_RAW:
+    if (rs_.gps1_raw.pubGEp)
```
kylemallory, line 2093: These "if" should be indented under the case, to match the indent level of the following "break" (indented 1 depth further than the enclosing braces of the 'switch' statement.

ros/test/test_main.cpp
```diff
@@ -60,86 +61,87 @@ TEST(test_main, basic)
...
-    TEST_COUT << "Timestamp Deviation (GPS <> INS):   [" << testNode.get_min_deviation(testNode.gps_ts, testNode.ins_ts)  << " <= " <<  testNode.get_avg_deviation(testNode.gps_ts, testNode.ins_ts) << " <= "  << testNode.get_max_deviation(testNode.gps_ts, testNode.ins_ts) << "]" << :: std::endl;
-    TEST_COUT << "Timestamp Deviation (INS <> IMU):   [" << testNode.get_min_deviation(testNode.ins_ts, testNode.imu_ts)  << " <= " <<  testNode.get_avg_deviation(testNode.ins_ts, testNode.imu_ts) << " <= "  << testNode.get_max_deviation(testNode.ins_ts, testNode.imu_ts) << "]" << :: std::endl;
-
-    EXPECT_GE( 0.05,  testNode.get_avg_deviation(testNode.gps_ts, testNode.pimu_ts));
-    EXPECT_GE( 0.05,  testNode.get_avg_deviation(testNode.gps_ts, testNode.imu_ts));
-    EXPECT_GE( 0.05,  testNode.get_avg_deviation(testNode.gps_ts, testNode.ins_ts));
-    EXPECT_GE( 0.005, testNode.get_avg_deviation(testNode.ins_ts, testNode.imu_ts));
-    isROS.terminate();
-}
+// TEST(test_main, gps_ins_time_sync)
```
kylemallory, line 64: Can we put this in a "#if 0  ...  #endif" block instead of as a giant block of single-line comments?
Also, be sure to add a comment after the `#if 0` that explains why its been turned off.  Ie, "This test has been disabled for reason X..."

kylemallory (changes requested):

.github/workflows/main.yml
andrewpriddis: Whoever wrote the test above this did it that way so when I was adjusting the calls below I just matched what was above. I know that the way of calling the tests in the upper test was ROS more normal so I matched the other pieces as well.

andrewpriddis: Also, if you do it the "new" method then the output is more verbose. That was one big benefit.

waltjohnson (approved):

ros/src/inertial_sense_ros.cpp
andrewpriddis: I added this piece which, yes does seem odd, but it helps identify a false connection which is indicated by a serial number of 0.

ros/src/inertial_sense_ros.cpp
andrewpriddis: I'd just like to say that the comment was made by Walt when he wrote this last year. 😄  I just halved the sleep time since it took a while. Unless it's really necessary I don't want to chop of this code right now. 🤷

.github/workflows/main.yml
kylemallory: How do you ensure the roscore is stopped at the end correctly?   I guess I'm curious what the point of `rostest` is, if the "more ROS way" to run tests is to not use rostest?

ros/src/inertial_sense_ros.cpp
kylemallory: I think this probably poor design then (somewhere, if not here).  While not "normal", its possible to have a 0 serial number.  If IS_.Open() is returning true/success, but serialNum == 0, then that implies some other problem (open port but unresponsive device, or failed to query devInfo, etc).  I think, within the `develop` branch, there are probably better ways to check for a valid device.  Check out the ISDevice class; there is a hdwRunState, that maybe more reliable.

In any case, this is probably better than nothing at all.

kylemallory (approved):

.github/workflows/main.yml
andrewpriddis: I understand it as catkin_make run_test_x plus catkin_test_results is the way ROS handles gtest, etc better.

PR #711, opened 2024-09-08
re-init Ser0 support

waltjohnson:
Add sys command to re-initialize ser0.

### DIFF
--- a/src/InertialSense.cpp
+++ b/src/InertialSense.cpp
@@ -781,7 +781,7 @@ void InertialSense::SetSysCmd(const uint32_t command, int pHandle)
  *       pHandle: Send in target COM port. 
  *                If arg is < 0 default port will be used 
 */
-void InertialSense::SetEventFilter(int target, uint32_t msgTypeIdMask, uint8_t portMask, uint8_t priorityLevel, int pHandle)
+void InertialSense::SetEventFilter(int target, uint32_t msgTypeIdMask, uint8_t portMask, int8_t priorityLevel, int pHandle)
 {
     #define EVENT_MAX_SIZE (1024 + DID_EVENT_HEADER_SIZE)
     uint8_t data[EVENT_MAX_SIZE] = {0};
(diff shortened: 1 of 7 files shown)


src/InertialSense.cpp
```diff
@@ -781,7 +781,7 @@ void InertialSense::SetSysCmd(const uint32_t command, int pHandle)
  *       pHandle: Send in target COM port. 
  *                If arg is < 0 default port will be used 
 */
-void InertialSense::SetEventFilter(int target, uint32_t msgTypeIdMask, uint8_t portMask, uint8_t priorityLevel, int pHandle)
+void InertialSense::SetEventFilter(int target, uint32_t msgTypeIdMask, uint8_t portMask, int8_t priorityLevel, int pHandle)
```
kylemallory, line 784: I am not sure what is right here, but I feel like this data type has changed a number of times in the last few months...   Either its been reverted accidentally, or you changed this back?

Perhaps we should make this an enum, so that it doesn't keep changing, and we KNOW what its types/values are?  @tmcd249 thoughts?

kylemallory (approved):

src/InertialSense.cpp
waltjohnson: It is an enum where one of the values is -1.  I had switched these to unsigned one PR and then Tony corrected me.

```
enum eEventPriority
{
    EVENT_PRIORITY_FAULT            = 0,
    EVENT_PRIORITY_ERR              = 1,
    EVENT_PRIORITY_WARNING          = 2,
    EVENT_PRIORITY_INFO             = 3,
    EVENT_PRIORITY_INFO_VERBOSE     = 4,
    EVENT_PRIORITY_DBG              = 5,
    EVENT_PRIORITY_DBG_VERBOSE      = 6, 
    EVENT_PRIORITY_TRIVIAL          = 7,
    
    // None should be used on all messages that should only be broadcast based on ID
    EVENT_PRIORITY_NONE             = -1,
};
```

PR #714, opened 2024-09-13
save gsv filter

tmcd249:
(IMX) Added support for save persistent of the GSV filter.

### DIFF
--- a/src/protocol_nmea.cpp
+++ b/src/protocol_nmea.cpp
@@ -22,10 +22,6 @@ static struct
 uint8_t nmea2p3_svid_to_sigId(uint8_t gnssId, uint16_t svId);
 bool gsv_freq_ena(gps_sig_sv_t* sig);
 
-typedef struct {
-    uint8_t constMask[SAT_SV_GNSS_ID_COUNT]; /* Constilation mask (see eGnGSVIndex)*/
-} gsvMask_t;
-
 static gsvMask_t s_gsvMask = {0};
 
 //////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


andrewpriddis (approved):

src/protocol_nmea.cpp
```diff
@@ -54,6 +50,32 @@ int ssnprintf(char buf[], int bufSize, const char *fmt, ...)
     return l;
 }
 
+/**
+ * Sets Gsv filter. 
+ * 
+ * @param seedArr - filters for each constellation (uint8_t seedArr[SAT_SV_GNSS_ID_COUNT])
+ */
+void nmea_setGsvFilter(uint8_t* seed)
```
kylemallory, line 58: Should be renamed to "nmea_harvestGsvFilter()".  :grinning: :seedling:

src/protocol_nmea.cpp
```diff
@@ -54,6 +50,32 @@ int ssnprintf(char buf[], int bufSize, const char *fmt, ...)
...
+        s_gsvMask.constMask[i] = seed[i];
+    }
+}
+
+/**
+ * Gets Gsv filter. 
+ * 
+ * @param seedArr - filters for each constellation (uint8_t seedArr[SAT_SV_GNSS_ID_COUNT])
+ */
+void nmea_getGsvFilter(uint8_t* seed)
```
kylemallory, line 71: Should be renamed to "nmea_plantGsvFilter()".  :grinning: :seedling:

kylemallory (approved):

src/protocol_nmea.cpp
tmcd249: if it wasn't in the SDK in half tempted to do it

PR #716, opened 2024-09-16
Add -nmea option to cltool

waltjohnson:
- Renamed `asciiMessages` to `nmeaMessage`.
- Added `-nmea` option to cltool help menu.
- Print to screen sent NMEA message with checksum footer.

### DIFF
--- a/cltool/src/cltool_main.cpp
+++ b/cltool/src/cltool_main.cpp
@@ -942,6 +942,9 @@ static int inertialSenseMain()
                 printf("%s", (char*)asciiData);
                 printf("\r\n");
             }
+
+            // Scan for "q" press to exit program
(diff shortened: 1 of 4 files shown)


cltool/src/cltool_main.cpp
```diff
@@ -878,6 +878,22 @@ static void sigint_cb(int sig)
     signal(SIGINT, SIG_DFL);
 }
 
+// Create and send full NMEA message with terminator w/ checksum trailer
+static void sendNmea(serial_port_t &port, string nmeaMsg)
```
kylemallory, line 882: This should be a full doxygen block.

kylemallory (approved):
Still have heartburn over this going into a 2.1.5-rc...

PR #729, opened 2024-10-08
Add parameter for rtk snrmin

waltjohnson:
Add flash config parameter to set RTK snrmin.

### DIFF
--- a/src/data_sets.h
+++ b/src/data_sets.h
@@ -3320,6 +3320,9 @@ typedef struct PACKED
     /** Reserved */
     uint8_t                 reserved[3];
 
+    /** Reserved */
+    uint32_t                reserved2[2];
+
 } nvm_flash_cfg_t;
 
 /** (DID_INL2_NED_SIGMA) Standard deviation of INL2 EKF estimates in the NED frame. */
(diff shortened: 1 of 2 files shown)


aabogdanov (approved):
Just a suggestion. Maybe we should start using GNSS instead of GPS in variable names. Just in case we need to use constellation-specific parameters in the future.

tmcd249 (approved):

src/data_sets.h
```diff
@@ -3314,6 +3314,12 @@ typedef struct PACKED
 	/** Magnetometer calibration quality sensitivity threshold. Typical range is 10-20 (10 default) and 1000 to disable mag calibration quality check, forcing it to be always good. */
 	float                   magCalibrationQualityThreshold;
 
+    /** Minimum SNR (4x CN0) used in GPS RTK solution */
+    uint8_t                 gpsMinSnr;
+
+    /** Reserved */
+    uint8_t                 reserved[3];
```
kylemallory, line 3321: I think if we are going to extend this, we should extend it by more than 4 bytes.  Maybe a uint8_t[4], a uint16_t[2] and a uint32_t[2].  This would give us a fair amount of breathing room to accommodate different types for the foreseeable future.

kylemallory (changes requested):
This looks good, and no issue, but I do think if we are expanding flashCfg, let's expand it by more than 4 bytes.

src/data_sets.h
waltjohnson: Sounds good

kylemallory (approved):

PR #758, opened 2024-11-12
SN-5909 map error

waltjohnson:


tmcd249 (approved):

src/ISLogger.cpp
```diff
@@ -678,57 +677,23 @@ float cISLogger::LogSizeAllMB()
...
+    return (DeviceFind(devSerialNo) ? m_devices[devSerialNo]->FileCount() : 0);
 }
 
 std::string cISLogger::GetNewFileName(uint32_t devSerialNo, uint32_t fileCount, const char *suffix)
 {
-    return (m_devices.find(devSerialNo) == m_devices.end()) ? m_devices[devSerialNo]->GetNewFileName(devSerialNo, fileCount, suffix) : std::string("");
-}
-
-/*
-bool cISLogger::SetDevice(const ISDevice *info)
```
kylemallory, line 700: I am not entirely sure why these were commented out.  I thought they were used elsewhere in the code; but clearly not.  Maybe I commented them out, thinking they were, but clearly aren't, so... yeah, let's delete the code.

src/ISLogger.cpp
```diff
@@ -319,6 +319,11 @@ std::shared_ptr<cDeviceLog> cISLogger::registerDevice(uint16_t hdwId, uint32_t s
     return deviceLog;
 }
 
+bool cISLogger::DeviceFind(int serialNum)
```
kylemallory, line 322: I am not a fan of this function (or at least its name).  This implies that it finds the associated device, by serial number, but it doesn't.  It should probably be renamed to `hasSerialNum(...)` or `hasDeviceBySerialNum(...)`  as it more clearly indicates that its a test of whether a DeviceLog exists associated with a particular serial number.

However, I thought there was already to function to return a DeviceLog* by a serial number, or return NULL if its not found.  If that does exist, maybe we should just use that instead, checking for NULL.  If it doesn't, then perhaps it makes more sense to implement that capability instead of this call.

kylemallory: The more I think about it, I think the correct way forward is to implement a `DeviceLogBySerialNum(int serialNum)` which returns the pointer.  The subsequent calls this this function should call to get the DeviceLog(), and if not NULL, then return the relevant data.

kylemallory (changes requested):

src/ISLogger.cpp
waltjohnson: Ok.  Renamed to DeviceLogBySerialNum()

PR #770, opened 2024-11-23
Ros2 CMakeLists Fix and Readme Suggestions

bboyack:
The CMakeLists.txt in the ros2 directory was missing the install command for the actual inertial sense node and as such it was not available when trying to do ros2 run. It only had the test node available before. 
I also adjusted the Readme to recursively clone the git repo so that there aren't submodule issues. 
I also changed the name of the node in the readme to match the node name that comes out of running colcon build (this makes more sense to me but it's only a suggestion)
Oh and I was taught that colcon build is always supposed to be run in the ros2_ws directory instead of the src directory. I changed this in the readme but this is just how I do it.


kylemallory:
@bboyack can you redirect this PR to the `develop` branch, rather than `main`, and we can get your contribution merged. They are much appreciated!

kylemallory (approved):

PR #776, opened 2024-12-03
Sn 5933 npp init bug

aabogdanov:
SDK changes to go along with the NPP init PR in imx


src/DeviceLog.cpp
```diff
@@ -88,6 +88,21 @@ bool cDeviceLog::CloseAllFiles()
     return true;
 }
 
+void cDeviceLog::ResetToStart()
```
kylemallory, line 91: Can you put in a doxygen block that describes the purpose of this function?

kylemallory (approved):

waltjohnson (approved):

PR #784, opened 2024-12-11
Refactors ISConstants.h::EPS to ISEarth.c::C_ECEF_EPS.

kylemallory:
Pretty basic refactor; @aabogdanov we should probably consider refactoring a number of the constant variables in ISEarth.c to be more meaningful names, but totally a suggestion, so long as they are aren't included in the header file.

### DIFF
--- a/src/ISEarth.c
+++ b/src/ISEarth.c
@@ -25,6 +25,11 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #define C_NEG_MG_DIV_KT0_F			-1.18558314779367E-04f		// - (M * g)  / (K * T0)
 #define C_NEG_KT0_DIV_MG_F			-8.43466779922578000E+03f	// - (K * T0) / (M * g)
 
+#ifndef C_ECEF_EPS
+#define C_ECEF_EPS (1.0E-10f)
+#endif
+
+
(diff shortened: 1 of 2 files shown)


src/ISEarth.c
```diff
@@ -25,6 +25,11 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #define C_NEG_MG_DIV_KT0_F			-1.18558314779367E-04f		// - (M * g)  / (K * T0)
 #define C_NEG_KT0_DIV_MG_F			-8.43466779922578000E+03f	// - (K * T0) / (M * g)
 
+#ifndef C_ECEF_EPS
```
aabogdanov, line 28: Not a big deal, but it really has nothing to do with ECEF, more like for safe division and such. Maybe keep it generic, like IS_EPS or whatever?

aabogdanov (approved):
Kyle, see if it makes sense to give that EPS a different name?

src/ISEarth.c
kylemallory: I renamed to C_ECEF_EPSILON to clarify the name.  I'm going to keep the ECEF in the name, since its only usage is in ECEF calculations.  If we end up using this in a broader context, where it justifying putting it back into ISConstants.h, then we can put it behind an IS namespace, and probably define it as a C++ constexpr, so it honors scope.  For now, this will work given its limited usage.

PR #794, opened 2024-12-27
SN-6066 Add GPX TOD debug

waltjohnson:
- Created GPS timeOfWeek filter to prevent time from jumping on the GPX.  
- Added debug information and error reporting via GPX and IMX status flags.  
- Added test to SDK Runtime Tests shown in EvalTool to verify time does not jump.


src/protocol_nmea.cpp
```diff
@@ -758,11 +758,11 @@ void nmea_GPSTimeToUTCTimeMsPrecision_ZDA_debug(char* a, int aSize, int &offset,
 
     // Check for irregular update timing
     int32_t cpuDtMs = cpuMs - lastCpuMs;
-    bool cpuDtMsGood = _ABS(cpuDtMs) < 30000;
+    bool cpuDtMsGood = _ABS(cpuDtMs) < 5000;
```
kylemallory, line 761: Should these be attached to a constant/macro, so they aren't so magic?  is 5sec appropriate?  Or should an even smaller number be considered?  2.5s?  2? 1.5?

src/data_sets.h
```diff
@@ -1467,6 +1467,11 @@ enum eGenFaultCodes
     GFC_CHIP_ERASE_INVALID				= 0x01000000,
     /*! Fault: GPS time fault */
     GFC_GNSS_TIME_FAULT                 = 0x02000000,
+    /*! Fault: GPS receiver time fault */
+    GFC_GNSS_RECEIVER_TIME              = 0x04000000,
```
kylemallory, line 1471: This feels a little redundant.  It probably worth clarifying the differences in GNSS_RECEIVER_TIME and GNSS_TIME_FAULT in the comments.  Each are GNSS, TIME, and FAULTs.  The assumption is that one it not a receiver fault, but the other is; but what does that really mean?

src/time_conversion.cpp
```diff
@@ -203,13 +203,20 @@ double timeToGpst(gtime_t t, int *week)
...
-void UtcDateTimeToGpsTime(const int dateTime[6], int leapSeconds, uint32_t &gpsTowMs, uint32_t &gpsWeek)
+void UtcDateTimeToGpsTime(const int dateTime[6], int leapSeconds, uint32_t &gpsTowMs, uint32_t &gpsWeek, double *debugGtm)
 {
     gtime_t gtm = epochToTime(dateTime);
     int week;
     gtm.time += leapSeconds;
     double iTOWd = timeToGpst(gtm, &week);
-    gpsTowMs = (uint32_t)((iTOWd * 1000.0) + 0.5);  // Round properly to nearest millisecond
+
+    // TODO: Remove debug later (WHJ)
```
kylemallory, line 213: Not an issue, just curious;  Seems like a curious place to add this code; it doesn't seem to be dependent on anything above or below.  Ie, this doesn't use iTOWd, but the line below the debug code does, so the debug code breaks up the context of the usage of iTOWd.

tests/runtime/device_runtime_tests.cpp
```diff
@@ -113,19 +109,22 @@ void DeviceRuntimeTests::ProcessISB(const p_data_hdr_t &dataHdr, const uint8_t *
     }
 }
 
-void DeviceRuntimeTests::TestIsbGps(const p_data_hdr_t &dataHdr, const uint8_t *dataBuf)
+std::deque<DeviceRuntimeTests::msg_history_t>& DeviceRuntimeTests::AddMsgHistory(std::deque<DeviceRuntimeTests::msg_history_t> &hist, DeviceRuntimeTests::msg_history_t msgHist)
```
kylemallory, line 112: We really need to get you out of the habit of duplicating work that you know I'm already working on.  :grin:

tests/runtime/device_runtime_tests.cpp
```diff
@@ -113,19 +109,22 @@ void DeviceRuntimeTests::ProcessISB(const p_data_hdr_t &dataHdr, const uint8_t *
...
-    msg_history_t &hist = m_hist.isb.gps1Pos;
-
-    copyDataPToStructP2(&m_gps1Pos, &dataHdr, dataBuf, sizeof(gps_pos_t));
+    hist.push_front(msgHist);
+    if (hist.size() > 3) { hist.pop_back(); }    // Keep 2, newest at front
+    return hist;
+}
 
-    // printf("ISB GpsPos1 (%d ms, %d wkday): %.*s", m_gps1Pos.timeOfWeekMs, utcWeekday, msgSize, msg);
+void DeviceRuntimeTests::TestIsbGps(const p_data_hdr_t &dataHdr, const uint8_t *dataBuf)
```
kylemallory, line 119: Even though these are internal functions, having doxygen or at least high-level comments that describe the purpose of the test really make it easier for other to follow in your footsteps.  Its not really clear what these next few tests are doing, and why they are important.

kylemallory (changes requested):

src/protocol_nmea.cpp
waltjohnson: This code is intended to be temporary and removed so I didn't invest as much time into it.

kylemallory (approved):

PR #815, opened 2025-01-23
Sn 6191 add event mem request

tmcd249:
Add event IMX mem request


src/protocol_nmea.cpp
```diff
@@ -771,7 +771,7 @@ void nmea_GPSTimeToUTCTimeMsPrecision_ZDA_debug(char* a, int aSize, int &offset,
     if (cpuDtMsGood && utcDtMsGood)
     {   // No time wrap
         g_debug.i[3] = utcDtMs;
-        g_debug.i[4] = cpuDtMs;
+        // g_debug.i[4] = cpuDtMs;
```
kylemallory, line 774: Do these need to be updated in the data_sets.h? (which debugs are used by what?)

kylemallory (approved):

waltjohnson (approved):

PR #829, opened 2025-02-05
Merge main into develop

waltjohnson:


aabogdanov (approved):

src/filters.cpp
```diff
@@ -482,36 +485,36 @@ float integrateDeltaThetaVelRoscoe(
     ixVector3 delta_alpha;
     ixVector3 delta_veloc;
 
-    //__________________________________________________________________________________________________________________
```
kylemallory, line 485: This appears to be entirely whitespace changes, and should be reverted.

kylemallory (commented):
A bit of whitespace...

tmcd249 (approved):

PR #830, opened 2025-02-07
RTK EKF refactor

aabogdanov:
Added velocity constraint parameter


src/data_sets.h
```diff
@@ -3756,6 +3756,8 @@ typedef struct
     double baseline[3];
     double max_baseline_error;
     double reset_baseline_error;
+    /** velocity constraint in compassing mode {var before fix, var after fix} (m^2/s^2) **/
```
kylemallory, line 3759: This should probably be added at the end, so it doesn't break compatibility with different versions of the SDK/firmware.

kylemallory (changes requested):
Minor fix, but important.

waltjohnson (approved):

src/data_sets.h
aabogdanov: moved it

waltjohnson (approved):

kylemallory (approved):

PR #841, opened 2025-03-11
Comment out serCheckInit() code

waltjohnson:
Comment out the serInitCheck() code to reduce the firmware image size.


kylemallory (approved):
I think I probably would have preferred to remove this (since it never actually proved valuable), but this works too.

PR #844, opened 2025-03-16
SN-6377 developer mode bit

waltjohnson:
Add developer mode bit


waltjohnson:
Kyle, Tony, and Walt discussed this and determined that we don’t want to bother with merging this and the hassle of having to ensure the bit is cleared before shipping.  We do set a warning in the EvalTool notifying us that the developer bit is set.
Regardless, we have larger fish to fry right now.

PR #850, opened 2025-03-22
Change logger time format from seconds to h:mm:ss

waltjohnson:
Display elapsed runtime in cltool as `h:mm:ss` rather than `ss.s`.


kylemallory (approved):
I like this change. I don't think it will impact Intellian in any way, but should we check before merging this?

tmcd249 (approved):

PR #886, opened 2025-05-07
SN-6599: ISComm reset empty buffer

tmcd249:
- (SDK) (SN-6599) Resolved issue where an edge case in is_comm_free() would cause ISComm buffer to fill up even when data in the buffer has been assessed.

### DIFF
--- a/src/ISComm.c
+++ b/src/ISComm.c
@@ -169,16 +169,32 @@ int validateBaudRate(unsigned int baudRate)
     return -1;    
 }
 
+
+/**
+ * @brief Sets buffer to inital state
+ * 
+ * @param c 
+ * @param buffer 
+ * @param bufferSize 
+ * @return int 
+ */
+int resetBuffer(is_comm_instance_t* c, uint8_t *buffer, int bufferSize)
+{
+    c->rxBuf.size = bufferSize;
+    c->rxBuf.start = buffer;
+    c->rxBuf.end = buffer + bufferSize;
+    c->rxBuf.head = c->rxBuf.tail = c->rxBuf.scan = c->rxBuf.scanPrior = buffer;
+
+    return bufferSize;
+}
+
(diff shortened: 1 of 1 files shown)


src/ISComm.c
```diff
@@ -169,16 +169,32 @@ int validateBaudRate(unsigned int baudRate)
...
 }
 
+
+/**
+ * @brief Sets buffer to inital state
+ * 
+ * @param c 
+ * @param buffer 
+ * @param bufferSize 
+ * @return int
```
kylemallory, line 179: This should describe what the parameters are, specifically the return value.

src/ISComm.c
```diff
@@ -169,16 +169,32 @@ int validateBaudRate(unsigned int baudRate)
...
+
+/**
+ * @brief Sets buffer to inital state
+ * 
+ * @param c 
+ * @param buffer 
+ * @param bufferSize 
+ * @return int 
+ */
+int resetBuffer(is_comm_instance_t* c, uint8_t *buffer, int bufferSize)
```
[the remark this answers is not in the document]
tmcd249: walt requested it to be title this since this is just and internal helper not for external consumption.

kylemallory (approved):

waltjohnson (approved):

src/ISComm.c
```diff
@@ -883,20 +899,24 @@ int is_comm_free(is_comm_instance_t* c)
...
             buf->scan -= shift;
-        }
 
-        // re-calculate free byte count
-        bytesFree = (int)(buf->end - buf->tail);
+            // re-calculate free byte count
+            bytesFree = (int)(buf->end - buf->tail);
+        }
+    }
+    else if (c->processPkt == NULL && buf->scan == buf->tail)
```
waltjohnson, line 915: Can we change this to `else if (c->processPkt == NULL && buf->scan >= buf->tail)`

waltjohnson (changes requested):

PR #891, opened 2025-05-13
SN-6607: Add unit test for is_comm_buffer_parse_messages()

tmcd249:
- (SDK) Implimented unit test for is_comm_buffer_parse_messages().
- (SDK) Fixes edge case found while implementing unit test.

### DIFF
--- a/tests/test_ISComm.cpp
+++ b/tests/test_ISComm.cpp
@@ -1322,7 +1322,7 @@ TEST(ISComm, TruncatedPackets)
 
 #if TEST_BUFF_PARSE_MSG
 
-#define BUFF_PARSE_PASSES		50000
+#define BUFF_PARSE_PASSES		1000000
 #define BUFF_PARSE_OUT_BUF_SIZE 600  
 #define BUFF_PARSE_DEV          0  
 #define BUFF_PARSE_DEV_NMEA     1  
--- a/tests/test_ISComm.cpp
+++ b/tests/test_ISComm.cpp
@@ -821,40 +821,6 @@ static int generate_ISBPkt_imu(is_comm_instance_t* comm, uint8_t* buf, int buffS
 	return is_comm_write_to_buf(buf, buffSize, comm, PKT_TYPE_DATA, DID_IMU, sizeof(imu_t), 0, &imu);
 }
 
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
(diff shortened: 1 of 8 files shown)


tests/test_ISComm.cpp
```diff
@@ -1319,3 +1318,255 @@ TEST(ISComm, TruncatedPackets)
...
+#define BUFF_PARSE_INS          4  
+
+static uint32_t s_buffParseMsgInCnt[5] = { 0 };
+
+/**
+ * @brief ISB callback for testing the function is_comm_buffer_parse_messages()
+ */
+int BufferParse_isb(unsigned int port, p_data_t* data)
+{
+	switch (data->hdr.id)
```
kylemallory, line 1340: TABS!!!

tests/test_ISComm.cpp
```diff
@@ -1319,3 +1318,255 @@ TEST(ISComm, TruncatedPackets)
...
+TEST(ISComm, BufferParse)
+{
+    is_comm_instance_t comm;
+
+    is_comm_callbacks_t callbacks = {};
+	callbacks.isbData = BufferParse_isb;
+	callbacks.nmea = BufferParse_nmea;
+
+    uint32_t msgOutCnt[5] = {0};
+    uint8_t randomBuf[16] = {0xe0,0x63,0xa4,0x17,0x95,0xb1,0x26,0xd2,0xe0,0x63,0xa4,0x17,0x95,0xb1,0x26,0xd2};
```
kylemallory, line 1375: Is this random data?  If so, it should probably be randomly generated (dynamically).  Also, a lot of the following code appears to be "test data".  I think we have some routines that are responsible for actually generating test data, with some random parameters, etc.

At best, you should be trying to use those functions. Otherwise, we should try and make this reuseable, so that if we need custom data like this elsewhere, it can be reused.

kylemallory (changes requested):
There are a lot of mixed whitespace in the new code.  Let's clean that up.

Also, there are a number of existing mechanisms in test_ISComm that are supposed to test other aspects of ISComm.  I guess I'm curious what these new tests are testing that are not covered by the other tests?  It would probably be helpful to have a more detailed description at part of the doxygen on TEST(ISComm, BufferParse) that provides a more detailed, but still high-level explanation of what this test does and why its necessary.

kylemallory (approved):

waltjohnson (approved):

PR #894, opened 2025-05-19
SN-6232: $GLL conform with NMEA standard

tmcd249:
- (SDK) (SN-6232, PR-894) Update $GxGLL to conform with NMEA standard.


src/protocol_nmea.cpp
```diff
@@ -928,16 +928,30 @@ int nmea_gll(char a[], const int aSize, gps_pos_t &pos)
...
+    
     int n = nmea_talker(a, aSize);
     nmea_sprint(a, aSize, n, "GLL");
-    nmea_latToDegMin(a, aSize, n, pos.lla[0]);                      // 1,2
-    nmea_lonToDegMin(a, aSize, n, pos.lla[1]);                      // 3,4
-    nmea_GPSTimeToUTCTimeMsPrecision(a, aSize, n, pos);                        // 5
-    nmea_sprint(a, aSize, n, ",A");                                 // 6
+
+    if (((pos.lla[0] >= 0.000000000001) || (pos.lla[0] <= -0.000000000001)) &&
+        ((pos.lla[1] >= 0.000000000001) || (pos.lla[1] <= -0.000000000001))
```
waltjohnson, line 939: Is there an actual case you've run into or we know of that merits checking the LLA against a near zero value?  If not, I say we keep this simple and just use that status bits.  That's what they are there for and I can't think of a reason they wouldn't be valid.  Keep it simple, less code...

waltjohnson (changes requested):

kylemallory (approved):

waltjohnson (approved):

PR #907, opened 2025-06-20
Update test_flash_config.cpp

tmcd249:


src/ISDevice.cpp
```diff
@@ -808,6 +808,15 @@ bool ISDevice::WaitForFlashSynced(uint32_t timeout)
     unsigned int startMs = now;
 
     static unsigned int lastRequest = startMs;
+
+    if (forceSync)
+    {
+        sysParams.flashCfgChecksum = 0;
+        flashCfgUpload.checksum = 0;
+        flashCfgUploadTimeMs = 0;
+        printf("Forcing flash config sync.\n");
```
kylemallory, line 817: SDK should avoid printing.  We don't know when or where this will be called, or how often.

kylemallory (changes requested):

kylemallory (approved):

PR #928, opened 2025-07-28
Merge 2.6.0-rc into develop

waltjohnson:


tmcd249 (approved):

src/ISDevice.cpp
```diff
@@ -894,28 +936,132 @@ bool ISDevice::hasPendingFlashWrites(uint32_t& ageSinceLastPendingWrite) {
...
+        [this](gpx_flash_cfg_t& cfg) { return GpxFlashConfig(cfg); });
+}
+
+bool ISDevice::LoadImxFlashConfigFromFile(std::string path)
+{
+    return LoadFlashConfigFromFile<nvm_flash_cfg_t>(path, DID_FLASH_CONFIG,
+        [this](nvm_flash_cfg_t& cfg) { return SetImxFlashConfig(cfg); });
+}
+
+bool ISDevice::LoadGpxFlashConfigFromFile(std::string path)
```
kylemallory, line 1059: Doxygen

src/ISDevice.cpp
```diff
@@ -994,23 +1148,53 @@ int ISDevice::onIsbDataHandler(p_data_t* data, port_handle_t port)
...
     std::lock_guard<std::recursive_mutex> lock(portMutex);
 
     switch (getNmeaMsgId(msg, msgSize))
     {
         case NMEA_MSG_ID_INFO:
-            nmea_parse_info(devInfo, (const char*)msg, msgSize);
+        {	// IMX device Info
+            dev_info_t info;
+            nmea_parse_info(info, (const char*)msg, msgSize);
             hdwId = ENCODE_DEV_INFO_TO_HDW_ID(devInfo);
```
kylemallory, line 1166: This should ENCODE_DEV_INFO_TO_HDW_ID(info) --- not devInfo
The hdwRunState should also be made to info...

Or, these two lines need to be moved below the switch on the hardware type, otherwise, they will be overwritten by info.

src/InertialSense.cpp
```diff
@@ -947,45 +930,61 @@ void InertialSense::ProcessRxData(port_handle_t port, p_data_t* data)
...
-            break;
-        case DID_FIRMWARE_UPDATE:
-            // we don't respond to messages if we don't already have an active Updater
-            if (device->fwUpdater) {
-                device->fwUpdater->fwUpdate_processMessage(data->ptr, data->hdr.size);
-                device->fwUpdate();
-            }
-            break;
-    }
+    // THIS IS MOVED TO ISDevice::ProcessRxData()
```
kylemallory, line 933: If there is no need to keep this code, we should just remove it.  Git can serve as our history book.

src/InertialSense.cpp
```diff
@@ -1355,7 +1354,7 @@ bool InertialSense::OpenSerialPorts(const char* portPattern, int baudRate)
         if (device->hasDeviceInfo()) {
             device->GetData(DID_SYS_PARAMS);
             device->GetData(DID_FLASH_CONFIG);
-            // device->GetData(DID_EVB_FLASH_CFG);
+            device->GetData(DID_GPX_FLASH_CFG);
```
kylemallory, line 1357: I believe there is a flag on ISDevice that indicates if it or has a GPX.
This slaos doesn't request DID_GPX_STATUS, which it should, but ideally, only if its is actually an IMX or the IMX has a GPX attached.

src/InertialSense.cpp
```diff
@@ -1380,239 +1379,6 @@ void InertialSense::CloseSerialPorts(bool drainBeforeClose)
 }
 
 #if 0
```
kylemallory, line 1381: Should these SaveFlashConfig be another wrapper into WithDevice() and call the ISDevice version?

src/TcpPortFactory.cpp
```diff
@@ -32,16 +32,16 @@ port_handle_t TcpPortFactory::bindPort(const std::string& pName, uint16_t pType)
         return nullptr;
     }
 
-    sockaddr addr = {};
+    sockaddr_storage addr = {};
```
kylemallory, line 35: What was the reason for these changes??

tests/CMakeLists.txt
```diff
@@ -29,12 +29,23 @@ file(GLOB TESTS_SOURCES "${CMAKE_CURRENT_LIST_DIR}/test_*.cpp")
...
     list(FILTER TESTS_SOURCES EXCLUDE REGEX "test_time_conversion.cpp")
+    # yamlcpp needs to know it's a static lib
+    add_definitions(-DYAML_CPP_STATIC_DEFINE)
 endif()
 
 # Define the executable
 add_executable(${PROJECT_NAME} 
     ${TESTS_SOURCES}
     runtime/device_runtime_tests.cpp
+
```
kylemallory, line 40: Added but commented out?  SHould these be removed?

PR #932, opened 2025-08-06
SN-6787 Implement InertialSense Manufacturing Port Factory

FiriusFoxx:
This PR implements a port factory that automatically detects and configures TCP ports exposed over the LAN using a MDNS service called _inertialsense-manufacturing.

Connecting to a is-manu port with cltool isn't possible because cltool doesn't discover devices in a loop preventing mdns discoveries from completing although EvalTool which discovers devices in a loop is capable.

This PR also finishes #911 by adding support for hostnames and proper URL parsing support.


src/mdns.cpp
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    freeifaddrs(ifaddr);
+
+#endif
+    socketsOpened = num_sockets;
+    return num_sockets;
+}
+
+/**
```
kylemallory, line 301: Oooof! This is a brutal call back signature.  Is there a way we can reduce this?  At the very least, we should probably define a function typedef. Alternatively, maybe a struct or class that holds all this data, and then we can just pass a copy/reference/pointer of the struct.

Additionally, I think this function need more explanation about what its doing when its actually called.  This looks pretty dense without much inline documentation.

kylemallory (changes requested):
All in all, looks really good, but a few areas that need tweaking.

CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ if (WIN32)
 	add_dependencies(${PROJECT_NAME} gen_files)
 
     # We need to ensure windows static libs are found
-    target_link_libraries(${PROJECT_NAME} Ws2_32.lib)
+    target_link_libraries(${PROJECT_NAME} Ws2_32.lib Iphlpapi.lib)
```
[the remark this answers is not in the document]
FiriusFoxx: Iphlpapi is a Windows system library used to get information about network interfaces in mdns.cpp, so that we can start a listener on every interface.

src/mdns.cpp
FiriusFoxx: This is a callback from mdns.h and therefore can't be modified without changing mdns.h, which would prevent updating it.

PR #947, opened 2025-08-22
SN-7018 bootloader v6g fail w two imxs

waltjohnson:
Fixes an issue where IMX bootloader updates fail when starting with bootloader v6g and updating multiple devices. The fix adds a delay in the SDK firmware updater to ensure devices enter ROM bootloader mode before the update proceeds.


kylemallory (approved):
Wish there was a better way... but easy, if not cheap.

tmcd249 (approved):

PR #966, opened 2025-10-01
faster ecef-to-ned computation

aabogdanov:
Faster euler-to-quaternion for ECEF-to-NED by skipping sin, cos and multiplications for zero


kylemallory (approved):
We should have this tied to Jira ticket that is motivating the work.  Still, if its faster, its faster!

PR #987, opened 2025-10-25
SN-6900 windows log inspector

waltjohnson:
- Fixed ability to run LogInspector in Windows by moving the run action inside of build_log_inspector.py and importing the package as done in _probe_pyd_load().
- Removed unused duplicate python\LogInspector\log_reader.cpp
- Fixed activate_python_venv.bat in Windows, which was not activating the virtual environment due to `setlocal` and `endlocal` calls.
- Switch build option --nobuild to --no-build.

### DIFF
--- a/scripts/build_log_inspector.py
+++ b/scripts/build_log_inspector.py
@@ -16,7 +16,7 @@ IS_WINDOWS = (os.name == "nt") or (platform.system() == "Windows")
 PY = sys.executable  # venv python when activated
 
 def in_venv() -> bool:
-    return hasattr(sys, "real_prefix") or (hasattr(sys, "base_prefix") and sys.base_prefix != sys.prefix)
+    return sys.prefix != getattr(sys, "base_prefix", sys.prefix)
 
 _DLL_DIR_HANDLES: list = []
 _DLLS_ADDED = False


python/setup.py
```diff
@@ -3,15 +3,19 @@
 import sys
 import glob
 import setuptools
-from setuptools import setup, Extension, find_packages
+import distutils.command.build
```
kylemallory: I've got a few other changes here too.  distutils is apparently deprecated.

python/setup.py
```diff
@@ -136,7 +140,7 @@ def build_extensions(self):
     long_description_content_type='text/markdown',
     classifiers=[],
     package_dir={'': '.'},
-    packages=find_packages(where='../inertialsense'),
+    packages=find_namespace_packages(include=["inertialsense", "inertialsense.*"]),
```
kylemallory: this was an area I was looking into as well, but couldn't make it work.  👍

scripts/windows/build_log_inspector.bat
```diff
@@ -1,17 +1,24 @@
...
+setlocal
 
 echo Build Log Inspector
 echo.
 
-:: Set SDK_DIR as directory path
-for %%i in (%~dp0..\..) do SET SDK_DIR=%%~fi
-call %SDK_DIR%\scripts\lib\activate_python_venv.bat
+:: Set SDK_DIR to the repo root (two levels up from this script)
+for %%i in ("%~dp0..\..") do set "SDK_DIR=%%~fi"
```
kylemallory: I have a bunch of fixes for wrapping the `~-dp0..\..` in double quotes, as well as the SET, though specifically this should be `set SDK_DIR="%%!fi"`.  Doing this should eliminate the need to double quote the "%SDK_DIR%" uses below.

PR #1004, opened 2025-11-14
Sn-7273: IMX-6 ST IMU work

tmcd249:
impliments driver for 
defines device-specific imu number


src/data_sets.h
```diff
@@ -1782,6 +1782,7 @@ typedef struct PACKED
 typedef struct PACKED
 {                                                   // Sensor temperature compensation
     uint32_t                timeMs;                 // (ms) Time since boot up.
+    int                     imuCnt;
```
kylemallory: So we aren't here again, we should also include  magCnt.

kylemallory (changes requested):

kylemallory (approved):

PR #1012, opened 2025-11-25
Changes for HDT NMEA support

andrewpriddis:
Work in Progress. *** Testing Needed*** NMEA parser etc for HDT NMEA sentence.


kylemallory (changes requested):
A few things:

 - This PR does not have a JIRA card.  ALL branches and PRs should be associated with a JIRA card.  Its fine if you want to take on the work, and all, but it needs to be tracked properly in JIRA, with the key drivers (customer request, etc), story/feature description, etc.

 - Please rename the branch to inclue the SN-* from Jira.

 - Also, it looks like this parses HDT, but not does generate/render HDT.

kylemallory:
@andrewpriddis Do you want to keep this PR open?  If so, can you update it and address the issues/concerns?  Otherwise, I think we should close it out.

PR #1090, opened 2026-03-16
fix null buffer crash in CM

tmcd249:
Dont allow null buffers to be broadcast on embedded platforms


src/ISComm.c
```diff
@@ -1375,6 +1375,11 @@ int is_comm_write_isb_precomp_to_port(port_handle_t port, packet_t *pkt)
         return -1;
     }
 
+    if (pkt->data.ptr == NULL)
+    {   // Invalid payload pointer
+        return -1;
+    }
```
waltjohnson: Instead of this change, let's add a check if `pkt->data.ptr != NULL` to the if case where 'pkt->data.ptr` gets used.  That way we can allow packets of payload size zero and data.ptr == NULL to still be sent.

waltjohnson (changes requested):
Instead of this change, let's add a check if `pkt->data.ptr != NULL` to the if case where 'pkt->data.ptr` gets used.  That way we can allow packets of payload size zero and data.ptr == NULL to still be sent.

kylemallory (approved):

waltjohnson (approved):

PR #1123, opened 2026-04-29
SN-7879 (D-01): .idx v2 format + writer path

kylemallory:
## Summary

Replace the coarse v1 `.idx` (host-uptime delta + record counter, 16 bytes per record, no header) with a versioned **v2** format that carries the actual DID and a payload-derived timestamp per record, plus a fixed-size 64-byte header with magic / version / SDK producer version / finalize-time aggregates. Per [D0017 / D0023 / D0043 / D0049](https://github.com/inertialsense/logalyzer/blob/develop/docs/architecture/decisions.md).

## Why this PR is stacked on `SN-7880_timestamp`

D-01 doesn't strictly depend on D-06 (`TimeStamp`) but is third in a stack with [#1121](https://github.com/inertialsense/inertial-sense-sdk/pull/1121) (D-10) and [#1122](https://github.com/inertialsense/inertial-sense-sdk/pull/1122) (D-06). It uses `ISExpected<T>` from D-10 directly. Once #1121 and #1122 merge, GitHub auto-rebases this PR's base to `logalyzer-develop` and the diff narrows.

## On-disk layout (immutable once shipped — future changes go to v3)

**Header at offset 0, 64 bytes:**

| Offset | Field | Size | Notes |
|---|---|---|---|
| 0 | `magic` | 4 | `"ISIX"` |
| 4 | `version` | 2 | 2 = this format |
| 6 | `header_size` | 2 | 64 (lets v3+ extend without breaking v2 readers) |
| 8 | `producer_version` | 4 | packed `PROTOCOL_VERSION_CHAR0..3` |
| 12 | `total_records` | 8 | set on finalize |
| 20 | `first_timestamp_ms` | 8 | set on finalize |
| 28 | `last_timestamp_ms` | 8 | set on finalize |
| 36 | `sync_point_count` | 4 | for D-07 |
| 40 | `ts_units` | 1 | `TimestampUnits` enum |
| 41 | `ts_source` | 1 | `HeaderTimeSource` enum |
| 42 | `flags` | 1 | bit 0 = `FINALIZED` |
| 43..63 | reserved | 21 | zeroed |

**Records following the header, 24 bytes each:**

| Offset | Field | Size | Notes |
|---|---|---|---|
| 0 | `timestamp` | 8 | payload-derived ms (units per header `ts_units`) |
| 8 | `offset` | 8 | byte offset into the `.raw` segment |
| 16 | `did` | 4 | data ID (0 = no associated DID, streaming path) |
| 20 | `flags` | 2 | bit 0 = real ToW, bits 1-15 reserved |
| 22 | `reserved` | 2 |  |

All multi-byte fields **little-endian** on disk. Pure shift-and-mask serialize/parse — we never `memcpy` a struct directly to disk. The SDK targets ARM embedded boards where host endianness is a live question.

## Files

- **`src/ISLogIndex.{h,cpp}`** *(new)* — format constants, packed structs (size-asserted), `TimestampUnits` / `HeaderTimeSource` enums, record + header flag constants, pure serialize/parse helpers, `cISLogFileBase`-based wrappers, `makeDefaultHeader` convenience. All under `inertial_sense::idx::`.
- **`src/ISError.h`** *(modified)* — added `ISErrorCode::LegacyFormat`. D-10 explicitly anticipated this kind of growth (`More added as needed by downstream stories`).
- **`src/DeviceLog.h`** *(modified)* — rename current `index_record_s` → `index_record_v1_s` (kept for legacy reader path), alias public spelling to `is_log_idx_record_v2_t`. Widen `m_lastIndexOffset` to `uint64_t`. New `finalizeIndex()` declaration; `addIndexRecord` takes optional `(p_data_hdr_t*, const uint8_t*)`. New bookkeeping members.
- **`src/DeviceLog.cpp`** *(modified)* — `addIndexRecord` populates v2 fields (DID from header, ts from `cISDataMappings::Timestamp` if non-zero else host-uptime fallback, ToW flag from same signal). Drops the v1 1ms-dedup since v2 emits one record per data record. `writeIndexChunk` writes header on first call + appends records; `finalizeIndex` seeks(0) and rewrites the header with totals + FINALIZED flag; hooked into `CloseAllFiles`.
- **`tests/test_log_index.cpp`** *(new)* — gtest cases for static layout sizes, header round-trip, record round-trip, header+N-records contiguous buffer, v1 detection (`LegacyFormat`), wrong-magic / unsupported-version / impossible-header_size each with a distinct error code, explicit little-endian byte-order assertion.

## Deviations from spec, called out for review

1. **Path layout** — `src/...` at repo root, not `SDK/src/...` (same drift as D-10 / D-06 / D-20).
2. **Error type.** Story spec proposed a story-specific `ISLogIndexError`; used `inertial_sense::ISError` (D-10's canonical) instead, with a new `ISErrorCode::LegacyFormat` for the v1-detection case. Keeps the SDK error story coherent rather than fragmented across types.
3. **`header_size` field added** to the header layout (not literally in the AC). Lets future v3+ readers ignore unknown trailing fields gracefully, and v2 readers stop at byte 64 instead of misinterpreting v3 additions as records.
4. **v1 dedup dropped.** Old code skipped emitting a record if `current_uptimeMs() <= m_indexChunks.back().time` (1ms throttle). v2's AC "per-record" wording requires one index entry per data record so per-DID time-range queries can binary-search on `(did, timestamp)`.
5. **`Mixed` value in both `TimestampUnits` and `HeaderTimeSource`.** Lets the writer flag "this file mixes ToW and host-uptime records — readers must check per-record `flags`" without forcing a per-record-units field. Currently used by the SDK writer's default header since each record carries its own timestamp source.

## Test plan

- [x] `static_assert(sizeof(is_log_idx_header_t) == 64)` and `sizeof(is_log_idx_record_v2_t) == 24)` hold (compile-time).
- [x] **Standalone smoke** (`g++ -std=c++17 -I src -I src/tl-expected /tmp/smoke.cpp src/ISLogIndex.cpp`) green: layout, header round-trip, record round-trip, legacy detection, malformed cases, byte-order discipline.
- [x] **Logalyzer build links against rebuilt SDK** cleanly (`DeviceLog.cpp.o` + `ISLogIndex.cpp.o` present in build tree); ctest 2/2 still pass.
- [ ] **`ctest -R test_log_index` (SDK CI)** — not run on this VM (libgtest-dev missing); the committed gtest test will be exercised by SDK CI.
- [ ] **Real `.idx` spot-check** — write a representative `.raw`+`.idx` from a fixture and dump header + first 10 records. Pending availability of a fixture log on this VM; called out as a manual verification step before merge.
- [ ] **Size growth check** — verify `.idx` grows by approximately 2× v1 (24 vs 16 bytes per record). Trivially follows from the layout but worth confirming with a real run.


kylemallory:
## Fixup commits + perf analysis

Pushed two commits on top of the original PR:

- `c07e1a3c` — **per-packet .idx records + multi-segment finalize fix** (writer-side)
- `b264a56e` — **expand .idx test coverage + perf benchmark**

### Why two writer-side fixes

**Per-packet records:** `cDeviceLog::SaveData(int, ...)` was emitting a single chunk-level `addIndexRecord(did=0)` *before* the parser loop ran — so v2 was indexing chunks, not packets. Moved the call inside `cDeviceLogRaw::SaveData(uint8_t*, int)`'s parser loop, so each `_PTYPE_INERTIAL_SENSE_DATA / _CMD` packet emits its own record with the real DID + payload-derived timestamp + ToW flag.

**Multi-segment rotation:** when `.raw` exceeds `maxFileSize` the framework rotates segments via `OpenNewSaveFile()`, which reset `m_lastIndexOffset` but **not** `m_idxHeaderWritten`, `m_idxTotalRecords`, or the first/last timestamp counters. Consequence: segment 2's `writeIndexChunk` skipped the v2 header, and `finalizeIndex` later overwrote that segment's first ~3 records with a stale-totaled header at offset 0. Any log over the default 5 MB segment limit produced a poisoned `.idx` for every segment past the first. Fix: reset all four counters per-segment.

### Test coverage added

Beyond the original layout / round-trip / malformed-file unit tests:

| test | what it proves |
|---|---|
| `IdxFileIO.*` (3) | header + record write/read via `cISLogFile`; truncated-header and truncated-record paths |
| `IdxIntegration.EndToEndViaDeviceLogApi` | `cDeviceLog`-derived stub drives the index API; reads back via `ISLogIndex` helpers |
| `IdxIntegration.FinalizeWithoutAnyRecordsIsIdempotent` | empty-log close path doesn't crash, doesn't produce a stray `.idx` |
| `IdxIntegration.ISLoggerEndToEndProducesViableIdx` | full `cISLogger` framework with simulated telemetry from `GenerateRawLogData()`; asserts `.idx` is FINALIZED, has records, sees `HAS_TOW` flag, and **its record count + DID order match the ISB packets `cISLogger::ReadNextPacket` extracts when re-walking `.raw`** |
| `IdxIntegration.ISLoggerMultiSegmentRotationProducesValidIdxPerSegment` | forces segment rotation; every per-segment `.idx` is independently valid + FINALIZED, header `total_records` matches on-disk record count |
| `IdxRecordRange.LargeOffsetRoundTrip` | `>4 GiB` offsets round-trip cleanly (v2's u64-widened offset field) |
| `IdxFinalize.FinalizedFlagSemantics` | default header has FINALIZED *clear*, set on finalize |
| `IdxLittleEndian.ByteOrderIsExplicit` | known field bytes regardless of host endianness |

Suite: **36/36** Idx + ISTimeStamp + ISExpected + ISLogger.\* tests green.

### Performance benchmark (DISABLED, opt-in)

`IdxBenchmark.DISABLED_PerformanceProfile` sweeps log size and emits CSV between `#CSV-START / #CSV-END` sentinels for downstream plotting. Run with:

```
./IS-SDK_unit-tests --gtest_also_run_disabled_tests \
                    --gtest_filter='IdxBenchmark.*'
```

Findings:

| log size | .raw | .idx | overhead | write time | search via .raw | search via .idx | speedup |
|---|---|---|---|---|---|---|---|
| 0.25 MB | 258 KB | 31 KB | 11.8% | 18 ms | 13.83 ms | 0.005 ms | **2659×** |
| 0.50 MB | 512 KB | 61 KB | 11.9% | 21 ms | 27.91 ms | 0.007 ms | **3876×** |
| 1.00 MB | 1.0 MB | 121 KB | 11.8% | 42 ms | 54.63 ms | 0.014 ms | **4017×** |
| 2.00 MB | 2.0 MB | 242 KB | 11.8% | 108 ms | 109.98 ms | 0.029 ms | **3859×** |
| 4.00 MB | 4.0 MB | 481 KB | 11.7% | 162 ms | 220.78 ms | 0.056 ms | **3949×** |
| 8.00 MB | 8.0 MB | 965 KB | 11.8% | 359 ms | 437.40 ms | 0.114 ms | **3847×** |

- **Storage:** flat ~11.8% of `.raw`, regardless of log size. Driven by the average ISB packet size in the stream — high-rate IMU/GPS logs with bigger payloads will see lower percentages.
- **Searchability:** 3000–4000× faster for "find all packets with DID==X" — linear scan over 24-byte records vs full ISB-parser walk through chunked `.raw`.
- **Write-side:** ~22–25 MB/s effective throughput; per-packet `addIndexRecord` cost scales linearly with packet count, no nonlinear surprises.

### Caught along the way

`ISFirmwareUpdater.h:32-38` wraps `#include "com_manager.h"` in `extern "C" { ... }`, but `com_manager.h` declares C++ overloads. Anything that pulls in `ISFirmwareUpdater.h` before `com_manager.h` fails to compile. Worked around in the new tests by including `com_manager.h` first; underlying header bug is real but out of scope for D-01.

kylemallory:
### Perf-impact graph

v2 .idx impact analysis

Source data + plot snapshot are committed at `docs/perf/idx-v2-baseline.{png,csv}` (commit `81bc72cf`).

Four panels, one per finding from the analysis above:

1. **File size on disk** — bar chart of `.raw` vs `.idx` across the 0.25 – 8 MB sweep.
2. **Storage overhead** — `.idx` as % of `.raw`; flat at ~11.8% across all sizes.
3. **Query time, log scale** — "find all packets with DID==X", `.raw` scan via `cISLogger::ReadNextPacket` vs `.idx` record scan.
4. **Search speedup factor** — same workload expressed as ×; **2659× → 4017×** across the size range.

kylemallory:
### Extended sweep — 16 data points, up to 128 MB

Per follow-up request, re-ran the benchmark with 16 size points spanning 0.25 – 128 MB and converted all panels to line plots (speedup is now a line, not bars).

v2 .idx impact analysis — extended

**Findings hold across the extended range:**

- **Storage overhead:** mean 11.48% of `.raw`, no drift with log size (std-dev <0.5 percentage points).
- **Query speedup:** 2950× – 4017× across the sweep, mean ~3400×. Both `.raw`-scan and `.idx`-scan times grow linearly with log size, so the ratio is essentially flat — exactly what you'd expect when both methods do O(N_packets) work but with very different per-packet constants.
- **Multi-segment correctness implicitly verified:** at 128 MB with the default 5 MB segment cap, the writer rotates ~25 times. Every segment's `.idx` parses cleanly, every record reads back, and the search-via-idx path correctly aggregates across all segments — the multi-segment finalize fix from `c07e1a3c` is doing its job.

Source data is committed at `docs/perf/idx-v2-baseline.csv` (commit `f947f8db`).

kylemallory:
### Benchmark promoted to a standalone tool

Per follow-up: the perf harness is now a real utility under `tools/idx_benchmark/` (sibling of `src/`, mirroring `cltool/`'s pattern), not a DISABLED gtest test. The disabled test version has been removed from `tests/test_log_index.cpp`.

**Layout:**

```
tools/idx_benchmark/
  src/idx_benchmark.cpp    CLI entry point
  CMakeLists.txt           Standalone build
  plot.py                  4-panel plotting script
  README.md                Build + run + interpret docs
  .gitignore               build/
```

**Build:**

```sh
cd tools/idx_benchmark && mkdir -p build && cd build
cmake .. && cmake --build . -j
```

**Run:**

```sh
./idx_benchmark                       # default sweep: 16 sizes 0.25→128 MB
./idx_benchmark --sizes 1,4,16,64     # subset
./idx_benchmark --out-csv run.csv     # also write CSV file
./idx_benchmark --quiet               # silence per-row progress on stderr
```

CSV schema (11 cols): `size_mb, num_messages, raw_bytes, idx_bytes, write_ms, total_records, idx_full_read_ms, target_did, match_count, search_via_idx_ms, search_via_raw_ms`. Sentinels (`#CSV-START / #CSV-END`) on stdout for easy piping.

**Plot:**

```sh
./idx_benchmark | python3 ../plot.py - out.png
# or:
python3 ../plot.py /path/to/results.csv out.png
```

Requires `matplotlib` + `pandas`.

**Refreshed baseline** at commit `594f7a99` (regenerated using the new tool, which validates that the standalone path produces equivalent numbers):

v2 .idx impact analysis — refreshed via standalone tool

| metric | value |
|---|---|
| storage overhead (mean) | 11.70% of `.raw`, std-dev <0.5pp |
| query speedup (mean) | 3683× across 16 sizes |
| query speedup (range) | 2412× (96 MB) – 4108× (4 MB) |

Source CSV is committed at `docs/perf/idx-v2-baseline.csv`. See `tools/idx_benchmark/README.md` for full reproduction instructions and how to compare a new run against the committed baseline.

aabogdanov (approved):

PR #1124, opened 2026-04-29
fix: python build — add tl-expected to include path (followup to D-10)

kylemallory:
## Summary

Followup to PR #1121 (D-10 / SN-7881). When D-10 lands, the pybind11 wrapper at `python/` no longer builds — `ISError.h:17` includes `"tl/expected.hpp"` from the new vendored copy at `src/tl-expected/`, but that directory isn't on `setup.py`'s include search path.

```
fatal error: tl/expected.hpp: No such file or directory
   17 | #include "tl/expected.hpp"
```

This breaks `./scripts/build_log_inspector.sh` (which runs `pip install <SDK>/python` against the C++ source tree).

## Fix

Adds `<sdk>/src/tl-expected` to `python/setup.py`'s `include_dirs` list. One change, four lines including comment.

## How I caught it

While verifying that the legacy LogInspector still parses logs produced by the new D-01 (SN-7879) v2 `.idx` writer. Building from a recent SN-7879_idx-v2 (which stacks on top of D-10) reproduced the failure. With this fix in place, `build_log_inspector.sh` succeeds and the resulting `inertialsense.logs.LogReader` cleanly loads both single-segment and multi-segment v2 logs.

## Test plan

- [x] `./scripts/build_log_inspector.sh` succeeds on a branch that includes D-10
- [x] `inertialsense.logs.logReader.Log().load(<path>)` parses synthetic v2 logs (single-segment 1 MB and multi-segment 16 MB / 4 segments)
- [x] No regressions in the `pip install <SDK>/python` step

## Targeting

PR base is SN-7881_tl-expected (D-10's branch) so the fix lands as part of D-10. If preferred, easy to retarget to `logalyzer-develop` instead and merge as a sequenced followup.


kylemallory:
Adding the same one-line include-path fix for `cltool/CMakeLists.txt`. Same root cause as the python binding: `src/ISError.h` includes `tl/expected.hpp` and downstream consumers need `src/tl-expected/` on their include path.

Verified by rebuilding cltool with `bash scripts/build_cltool.sh` (clean, [BUILD: cltool - Passed]) and confirming the relay-discovery code path runs end-to-end via `./cltool/build/cltool -use-relay-list` (sends mDNS PTR for `_inertialsense-discovery._tcp.local`, returns cleanly). No IS devices currently advertising on the LAN here, but the discovery wiring is intact.

aabogdanov (approved):

waltjohnson (approved):

shefft:
I read through this and inspected Python and CMakeLists change, but appears I was too late to complete a review submission.

PR #1135, opened 2026-05-01
D-07 (SN-7897): ISTimeResolver — piecewise-linear wall-clock reconstruction

kylemallory:
## Summary

Lands the per-device resolver D0024 specified. Walks an `ISDeviceLog`, identifies ToW-bearing records as sync anchors, classifies any record timestamp into one of five confidence tiers, and surfaces clock-correction discontinuities for D-58 (chart-side time-quality display).

### API

```cpp
class ISTimeResolver {
public:
    static ISExpected<ISTimeResolver> build(const ISDeviceLog&);
    static ISExpected<ISTimeResolver> build(const ISDeviceLog&, double threshold);

    TimeStamp resolve(uint64_t hostTimeMs, uint64_t deviceId) const;

    static std::vector<ISSyncPoint> detectSyncPoints(const ISDeviceLog&);

    const std::vector<ISSyncPoint>&    syncPoints()      const noexcept;
    const std::vector<Discontinuity>&  discontinuities() const noexcept;
    Stats                              computeStats(const ISDeviceLog&) const;
    static constexpr double kDefaultDiscontinuityThreshold = 1.0e-3; // 1000 ppm
};
```

`resolve()` is O(log N) via binary-search on the sorted sync-point vector. `build()` is O(M) (linear scan of all segments via `is_comm_parse_byte`) plus O(N log N) sync-point sort.

### Detection

Walks each segment's raw bytes via `is_comm_parse_byte`, picking up ISB packets whose DID is in `kToWBearingDids[]` (`DID_INS_1..4`, `DID_GPS1_POS`, `DID_GPS2_POS`) AND whose payload yields a non-zero ToW via `cISDataMappings::Timestamp(...)`. This is the same condition the writer uses to set `IS_LOG_IDX_REC_FLAG_HAS_TOW` — so detection here recovers the same anchors regardless of how the source `.idx` was produced (or whether it exists at all; D-04 scan-rebuild paths still work).

### v2 .idx schema constraint (documented)

The current writer overwrites `.idx.timestamp` with the payload ToW when `HAS_TOW` is set — host-side capture time isn't preserved alongside ToW. So for v2 sync points, `hostTimeMs == payloadToWMs` by construction, and the slope between them is identity. Pre-fix records (host-uptime-delta in their `.idx.timestamp`) resolve as `ExtrapolatedBackward` with best-effort numeric values; consumers (D-58) treat that confidence tier with dashed-line decoration. A future v3 `.idx` schema with explicit host-time field lets the slope-fit math become non-trivial; the API shape doesn't change. The constraint is captured loud and clear in `ISTimeResolver.h`'s file-level doc + the `ISSyncPoint` struct comment.

### Side-quest: `ISLogReader::rawBytes()`

Same accessor D-08 (#1132) added — already on `logalyzer-develop` post-merge. The rebase merged D-08's version cleanly; this PR's contribution there is just the comment that mentions D-07 as another consumer of the hatch alongside D-03's templated sugar.

### Tests (8 new gtests, all green locally)

| Test | What it covers |
|---|---|
| `EmptyLogResolvesAsSessionOnly` | No sync points → `SessionOnly`/`Unknown` for any query |
| `SyncPointsFromInsRecords` | Detection + adjacent-duplicate filtering (5 records, 1 dup → 4 syncs) |
| `ResolveExactAtSyncPoint` | Query at sync timestamp → `Exact`/`PayloadToW` |
| `ResolveInterpolated` | Query between sync points → `Interpolated` |
| `ResolveExtrapolated` | Forward + backward extrapolation tiers |
| `DiscontinuityFromUnevenGaps` | v2 schema's identity-slope path produces no discontinuities (no false positives) |
| `StatsAddUpToRecordCount` | Histogram-sum invariant |
| `SingleSyncPointDegenerateSlope` | Resolver works with a single anchor |

15 reader regression tests + 6 DIDTraits + 7 writer tests still green.

### What's not done here

- **Real fixture file** (`sample_prefix_lock.raw` per the AC) — tests synthesize fixtures programmatically via `cISLogger`, which exercises the same code paths. The committed `.raw` file would be regenerated each CI run anyway (host-time-dependent), so a synthetic fixture is preferred per `test_log_reader`'s rationale.
- **Per-tile context** — AC mentions both per-sample and per-tile evaluation contexts. v1 is per-sample only; per-tile follows once D-32 lands the LOD pyramid.
- **Full ExampleProject** — stub committed per AC; full example with D-11.

## Test plan

- [x] `ctest -R TimeResolverTest` green (8/8) on Linux Ubuntu 22.04
- [x] Reader / Writer / DIDTraits regression tests green (28/28)
- [ ] Same on Windows runner via Logalyzer's build-Logalyzer workflow once submodule lands

### DIFF
--- a/ExampleProjects/ISTimeResolver/CMakeLists.txt
+++ b/ExampleProjects/ISTimeResolver/CMakeLists.txt
@@ -1,0 +1,7 @@
+# ExampleProjects/ISTimeResolver/CMakeLists.txt — D-07 / SN-7897 stub.
+# Full example lands with D-11 / SN-7900.
+
+cmake_minimum_required(VERSION 3.16)
+project(ISTimeResolver_example LANGUAGES CXX)
+
+message(STATUS "ExampleProjects/ISTimeResolver: stub (D-07); full example lands with D-11.")
--- a/ExampleProjects/ISTimeResolver/README.md
+++ b/ExampleProjects/ISTimeResolver/README.md
@@ -1,0 +1,51 @@
+# `ISTimeResolver` — piecewise-linear wall-clock reconstruction *(stub — D-11 fills out the full example)*
+
+Placeholder slot reserved by **D-07 / SN-7897**. Full code drops in
+with **D-11 / SN-7900**.
+
+## Quickstart
+
+```cpp
+#include "ISDeviceLog.h"
+#include "ISTimeResolver.h"
+
+#include <iostream>
+
+int main(int argc, char** argv) {
+    if (argc < 2) {
+        std::cerr << "Usage: " << argv[0] << " <segment.raw>\n";
+        return 1;
+    }
+    using namespace inertial_sense;
+
+    auto log = ISDeviceLog::fromSegments({ argv[1] });
+    if (!log) {
+        std::cerr << "open failed: " << log.error().message << "\n";
+        return 2;
+    }
+
+    auto resolver = ISTimeResolver::build(*log);
+    if (!resolver) {
+        std::cerr << "build failed: " << resolver.error().message << "\n";
+        return 3;
+    }
+
+    std::cout << "sync points:    " << resolver->syncPoints().size() << '\n'
+              << "discontinuities:" << resolver->discontinuities().size() << '\n';
+
+    auto stats = resolver->computeStats(*log);
+    std::cout << "stats: exact="   << stats.exact
+              << " interp="        << stats.interpolated
+              << " extrap_fwd="    << stats.extrapFwd
+              << " extrap_back="   << stats.extrapBack
+              << " unknown="       << stats.unknown << '\n';
+    return 0;
+}
+```
+
+## Related
+
+- API: [`SDK/src/ISTimeResolver.h`](../../src/ISTimeResolver.h),
+       [`SDK/src/ISSyncPoint.h`](../../src/ISSyncPoint.h)
+- Story: [SN-7897](https://inertialsense.atlassian.net/browse/SN-7897)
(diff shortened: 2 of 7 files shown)


kylemallory:
**Update — validated against real cltool capture**

Per Kyle's pointer: added a `TimeResolverLiveFixture.RealCltoolCaptureSmoke` test that exercises the resolver against the 120 s PPD capture from SN519465 in `inertialsense/sample_logs/cltool_imx5_120s_ppd/...`. Data isn't committed; the test gates on file presence (`IS_SDK_LIVE_FIXTURE_RAW` env override), `GTEST_SKIP` when absent.

Results on the reference capture:

| metric | value |
|---|---|
| records iterated | 27,480 |
| sync points (post-dedup) | 1,812 |
| discontinuities | 0 |
| stats: exact | 2,829 |
| stats: interpolated | 2,755 |
| stats: extrapolatedForward | 1 |
| stats: extrapolatedBackward | 21,895 |
| stats: unknown | 0 |
| **sum** | **27,480** ✓ |

The large `extrapBack=21,895` is exactly the v2 `.idx` schema-collapse behavior the resolver header documents: non-HAS_TOW records carry host-uptime-delta in their `.idx.timestamp` (small ms values), which falls before the GPS-ToW-anchored sync points. The resolver classifies them as `ExtrapolatedBackward` — honest behavior with the right confidence tier. UI consumers (D-58) handle that tier with dashed-line decoration. A future `.idx` v3 with explicit host-time field lets the slope fit become non-trivial without changing this API.

The "Real fixture file" item under "What's not done here" in the PR description is now closed — synthetic gtests + real-capture smoke test cover both paths.

PR #1165, opened 2026-05-19
SN-7659 RTOS cleanup

waltjohnson:
This pull request introduces several improvements and updates related to device firmware information reporting, protocol versioning, and platform-specific timing defaults. The main changes include adding a debug flag indicator to firmware version strings, updating protocol version macros, clarifying build flag naming, and refining navigation period definitions for different hardware platforms.

**Firmware information and build flags:**

* Appends a "-debug" suffix to the firmware version string when the device is built with debug flags, improving visibility of debug builds in firmware info (`ISDevice::getFirmwareInfo`, `BUILD_FLAGS_DEBUG` rename). [[1]](diffhunk://#diff-ebc1babca6ae696cf06797e0953cd073c03f5d637fbc74a09b23d9f67ab25d3fR713-R714) [[2]](diffhunk://#diff-a58fce04e6c6d3ce510e53e485a7dec47f3f8542be0cd23f2527a2254e87813aL634-R635)
* Renames `BUILD_FLAG_DEBUG` and `BUILD_FLAG_DIRTY` to `BUILD_FLAGS_DEBUG` and `BUILD_FLAGS_DIRTY` for consistency.

**Protocol versioning:**

* Increments the `PROTOCOL_VERSION_CHAR1` macro from 1 to 2 to indicate a breaking change in payload structure, ensuring proper version tracking.

**Platform-specific timing defaults:**

* Adds minimum navigation period macros for IMX6 hardware and updates the logic to select the correct timing defaults based on platform (`imx_defaults.h`). This improves clarity and maintainability for platform-dependent timing.

### DIFF
--- a/src/ISDevice.cpp
+++ b/src/ISDevice.cpp
@@ -710,6 +710,8 @@ std::string ISDevice::getFirmwareInfo(const dev_info_t &devInfo, int flags) {
         }
         if (devInfo.firmwareVer[3] != 0)
             out += utils::string_format(".%u", devInfo.firmwareVer[3]);
+        if (devInfo.buildFlags & BUILD_FLAGS_DEBUG) 
+            out += "-debug";
 
         if (devInfo.repoRevision && !(flags & OMIT_COMMIT_HASH)) {
             out += utils::string_format(" %08x", devInfo.repoRevision);
--- a/python/inertialsense/logInspector/logPlotter.py
+++ b/python/inertialsense/logInspector/logPlotter.py
@@ -3981,8 +3981,8 @@ class logPlot:
             else:
                 towOffset = 0
 
-            deltaTimestamp = 0
-            timeImu  = 0
+            deltaTimestamp = np.array([])
+            timeImu  = np.array([])
             timePimu = self.getData(d, DID_PIMU, 'time')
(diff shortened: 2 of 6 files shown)


src/ISDevice.cpp
```diff
@@ -710,6 +710,8 @@ std::string ISDevice::getFirmwareInfo(const dev_info_t &devInfo, int flags) {
         }
         if (devInfo.firmwareVer[3] != 0)
             out += utils::string_format(".%u", devInfo.firmwareVer[3]);
+        if (devInfo.buildFlags & BUILD_FLAGS_DEBUG)
```
kylemallory, line 713: You should be careful about this... I thought I had actually removed this (or refactored it) so that it would use the utils::* functions like utils::devInfoToString().  I think there are a few exceptions, and this could be one of them, but *generally* we want to have consistent *formatting* of Device Info, and consistent *parsing* of Device Info, and the utils:: functions try to standardize that.

Its probably fine, unless there is anything that is trying to parse the output from this ISDevice::getFirmwareInfo(), because it looks like it will get a funky "fw5.6.0-snap.1-debug 5023af57 bef94a.123 ..." type output.

kylemallory (approved):

tmcd249 (approved):

PR #1177, opened 2026-05-29
SN-7877: u5 SPI work

tmcd249:
SN-7877: u5 SPI work


waltjohnson (approved):

cltool/src/cltool.cpp
```diff
@@ -1252,6 +1260,10 @@ void cltool_outputUsage()
 	cout << "OPTIONS (General)" << endl;
 	cout << "    -baud=" << boldOff << "BAUDRATE  Set serial port baudrate.  Options: " << IS_BAUDRATE_115200 << ", " << IS_BAUDRATE_230400 << ", " << IS_BAUDRATE_460800 << ", " << IS_BAUDRATE_921600 << " (default)" << endlbOn;
 	cout << "    -c " << boldOff << "DEVICE_PORT  Select serial port(s). Options: single port (e.g., COM5 or /dev/ttyUSB0), multiple ports separated by ',' (e.g., COM2,COM4,COM5), \"*\" for all ports, or \"*4\" for first four ports." << endlbOn;
+    cout << "    -c " << boldOff << "//spi/DEVICE_PATH[OPTS]  Select SPI device (e.g., //spi/dev/spi0.0). Optional comma-separated OPTS in brackets:" << endlbOn;
```
kylemallory: `//spi/dev/spi0.0` is not the correct path for a spi port (I'm pretty sure).

cltool/src/cltool.cpp
```diff
@@ -389,8 +390,12 @@ bool cltool_parseCommandLine(int argc, char* argv[])
         else if (matches(a, "-c") && (i + 1) < argc)
         {
             // Supports: single port (e.g., "COM5"), comma-separated ports (e.g., "COM2,COM4,COM5"),
-            // wildcard (e.g., "*" for all ports, "*4" for first 4 ports)
-            g_commandLineOptions.comPort = argv[++i];   // use next argument
+            // wildcard (e.g., "*" for all ports, "*4" for first 4 ports).
```
kylemallory: You still shouldn't need to make checks like this for "useSpi".  It should all be handled natively by the fact that the port is a spi-port, not because there is a "useSpi" flag that is configured.  But I haven't finished reading the PR to see how you are using this flag later.

cltool/src/cltool.h
```diff
@@ -24,6 +24,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 // change these includes to be the correct path for your system
 #include "InertialSense.h" // best to include this file first
 #include "ISDisplay.h"
+#include "core/spiPort.h"
```
kylemallory: Again, cltool should be agnostic of a spiPort.

cltool/src/cltool.h
```diff
@@ -178,6 +179,10 @@ typedef struct cmd_options_s // we need to name this to make MSVC happy, since w
     uint32_t updateFirmwareSlot = 0;
     uint32_t runDurationMs = 0;             // Run for this many millis before exiting (0 = indefinitely)
     bool list_devices = false;              // if true, dumps results of findDevices() including port name.
+    bool useSpi = false;                    // if true, registers SpiPortFactory and targets the spidev path in comPort
```
kylemallory: No need to define these as globals - they should be handled as config parameters to the SpiPortFactory (same as Baud Rate for the SerialPortFactory)

kylemallory (approved):

PR #1186, opened 2026-06-09
SN-8043 - DFU: detect silent flash-write drop at RDP > 0xAA

Hephaestus-Maximus:
## Summary

Fixes SN-8043: `DFUDevice::updateFirmware()` returned `DFU_ERROR_NONE` when the STM32U5 ROM bootloader silently dropped flash writes at **RDP Level 1** (RDP > `0xAA`). Every `DFU_DNLOAD` was ACKed and `waitForState(DNLOAD_IDLE)` succeeded, so the host reported "Complete" while flash stayed empty — a bricked, empty-flash unit reported as success.

Two independent guards now close the silent-success window (suggested fixes **(1)** + **(3)** from the ticket):

### (1) Pre-flight RDP check — `checkReadProtection()`
At the top of both `updateFirmware()` overloads (STM32U5 only; no-op for other processors). Reads `FLASH_OPTR` via the DFU `OPTIONS` segment (readable regardless of RDP level) and gates on the RDP byte:

| RDP | Result |
|---|---|
| `0xAA` (Level 0) | proceed |
| `0xCC` (Level 2, permanent) | `DFU_ERROR_RDP_PERMANENT_LOCKED` |
| anything else (Level 1) | `DFU_ERROR_RDP_LOCKED` |

I chose the **bail-with-clear-error** path rather than auto-regress. The AC explicitly permits either; auto-regress requires a mandatory mass-erase + system reset + USB re-enumeration + re-acquire of the device handle, which can't be verified without hardware and is more invasive. Happy to add auto-regress as a follow-up if preferred — see open question below.

### (3) Post-write readback verification — `verifyFlashWrite()`
After the write loop, uploads the first 64 bytes of the written region and compares against the source image; mismatch → `DFU_ERROR_WRITE_VERIFY_FAILED`. The classic RDP-drop signature is an all-`0xFF` readback. A readback **transport** failure is logged but treated as non-fatal, to avoid false negatives on otherwise-good devices.

### Supporting changes
- New error codes: `DFU_ERROR_RDP_LOCKED` (-9), `DFU_ERROR_RDP_PERMANENT_LOCKED` (-10), `DFU_ERROR_WRITE_VERIFY_FAILED` (-11), with matching `dfuDeviceErrors[]` strings.
- `getErrorName()` is now **bounds-safe** (returns `"UNKNOWN"` for out-of-range indices instead of reading OOB) and **static** (testable; existing `instance->getErrorName()` callers unaffected).
- The pure RDP decision is extracted into `static DFUDevice::rdpVerdict(uint8_t)` for unit testing.

## Acceptance criteria

- [x] **A** (unit, self-verified): `tests/test_ISDFUFirmwareUpdater.cpp` — 5 GTest cases for `rdpVerdict()` (0xAA/0xCC/Level-1 set) and `getErrorName()` (known codes + out-of-range safety). All pass. The full `IS-SDK_unit-tests` target builds and links with these changes.
- [x] `updateFirmware()` no longer returns `DFU_ERROR_NONE` on a silent drop — it returns `DFU_ERROR_RDP_LOCKED` (pre-flight) or `DFU_ERROR_WRITE_VERIFY_FAILED` (post-write).
- [ ] **C** (hardware-in-the-loop, reviewer): put an STM32U5 module at RDP=`0xBB`, run DFU, confirm the failure is surfaced (not "Complete"); and confirm no regression on the normal RDP=`0xAA` provisioning path end-to-end.

## Companion PR

EvalTool change (imx) so the DFU Provisioning dialog displays the meaningful error name (AC #2) — bumps the SDK submodule to this branch. Link added once opened.

## Out of scope

Legacy `ISBootloaderDFU.cpp` (IMX-5 / STM32L4) — different RDP-vs-bootloader semantics; file a sibling ticket if affected (per the ticket).

## Open question for reviewers

Bail vs. auto-regress on RDP Level 1 — this PR bails with a clear error. If the factory flow would rather auto-recover (write OPTR `RDP=0xAA`, mass-erase, re-enumerate, continue), I can implement it as a follow-up; it needs bench validation of the re-enumeration timing.

### DIFF
--- a/tests/test_ISDFUFirmwareUpdater.cpp
+++ b/tests/test_ISDFUFirmwareUpdater.cpp
@@ -1,0 +1,48 @@
+/**
+ * Unit tests for the DFU updater's read-protection (RDP) decision logic and error-name lookup.
+ *
+ * SN-8043: the USB DFU transfer itself requires hardware and is covered by a hardware-in-the-loop
+ * test (Set C). These tests exercise the pure, hardware-independent decision logic that gates the
+ * pre-flight RDP check and the post-write verification error reporting.
+ */
+
+#include <gtest/gtest.h>
+#include "ISDFUFirmwareUpdater.h"
+
+// --- rdpVerdict(): raw FLASH_OPTR RDP byte -> dfu_error verdict ---
+
+TEST(ISDFUFirmwareUpdater_RDP, Level0_Unprotected_Proceeds) {
+    // RDP == 0xAA is the only value that permits flash programming.
+    EXPECT_EQ(DFU_ERROR_NONE, DFUDevice::rdpVerdict(0xAA));
+}
+
+TEST(ISDFUFirmwareUpdater_RDP, Level2_Permanent_Locked) {
+    EXPECT_EQ(DFU_ERROR_RDP_PERMANENT_LOCKED, DFUDevice::rdpVerdict(0xCC));
+}
+
+TEST(ISDFUFirmwareUpdater_RDP, Level1_Locked) {
+    // 0xBB is the production state called out in the bug report; any non-0xAA/0xCC value is Level 1.
+    EXPECT_EQ(DFU_ERROR_RDP_LOCKED, DFUDevice::rdpVerdict(0xBB));
+    EXPECT_EQ(DFU_ERROR_RDP_LOCKED, DFUDevice::rdpVerdict(0x00));
+    EXPECT_EQ(DFU_ERROR_RDP_LOCKED, DFUDevice::rdpVerdict(0xFF));
+    EXPECT_EQ(DFU_ERROR_RDP_LOCKED, DFUDevice::rdpVerdict(0x55));
+    EXPECT_EQ(DFU_ERROR_RDP_LOCKED, DFUDevice::rdpVerdict(0xAB)); // one off from unprotected
+}
+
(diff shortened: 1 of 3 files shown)


kylemallory (approved):

aabogdanov (approved):

kylemallory (approved):

tests/test_ISDFUFirmwareUpdater.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "ISDFUFirmwareUpdater.h"
+
+// --- rdpVerdict(): raw FLASH_OPTR RDP byte -> dfu_error verdict ---
+
+TEST(ISDFUFirmwareUpdater_RDP, Level0_Unprotected_Proceeds) {
+    // RDP == 0xAA is the only value that permits flash programming.
+    EXPECT_EQ(DFU_ERROR_NONE, DFUDevice::rdpVerdict(0xAA));
+}
+
+TEST(ISDFUFirmwareUpdater_RDP, Level2_Permanent_Locked) {
```
shefft, line 19: Minor visual commentary here on the test order, might be a bit cleaner to put in increasing order unless there is a good reason to do it this way:
Level0
Level2
Level1

shefft (commented):
Made one comment on one file

shefft (approved):

PR #1212, opened 2026-07-10
SN-6695 custom port example, now with SDK virtual test port replaced with custom port

shefft:
The first version of this new Example Project was built upon the existing SDK virtual serial test port code.  Now that port has been replaced with a refactored project-local virtual test port, with accompanying significant modifications to the comments and README to guide the user through creating both a port (or underlying channel implementation) AND the custom Port Factory.

21/Jul/2026 edit: Significant changes have been made  to include the PortManager demonstration with resulting code modifications, rename files, and reword and reorganize large portions of the README and diagrams.

2/Aug/2026 edit:  Changes to be rid of the multi-port init helper and go to single-port only, called from bind, and dynamically allocate ports one at a time instead of a group of static ports.  Name, type, etc now set on that init from bind.  Application messaging is now all to standard out, with logging restricted to SDK extensions like the custom port factory.  Simplified Port Manager calls to do less filtering and put port type specification inside locatePorts instead.  Re-structured STEPS comments and README to match updates.

Upon approval, must be SQUASH-MERGED.

---
**22/Jul/2026 review addendum (Hephaestus):** Reviewed against a checkout of the head commit, cross-referenced with `ring_buffer.c`/`base_port.h`/`PortManager.h` in the SDK. Requested changes — see the PR review for details:
- `customPortWrite()`'s `throw new std::out_of_range(...)` throws a pointer instead of a value (won't be caught by a normal `catch (const std::out_of_range&)`, and leaks); should just return `PORT_ERROR__WRITE_FAILURE` like its sibling functions do.
- `CustomVirtualExample.cpp` is still present but no longer built (superseded by `main.cpp`) — looks like a loose end from the file-rename commits; the README's `main.cpp` link also still points at the old file.

---
**22/Jul/2026 review addendum (Tyler):**  addressed both issues.  With the exception, opted for following Kyle's previous suggestion of pushing C++ code out of that particular file in favor of C which also aligns with Heph's recommendation to be rid of it and just return and handle an error code.


ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.cpp
```diff
@@ -0,0 +1,109 @@
...
+        #endif
+        #if (NUM_COM_PORTS > 3)
+            TEST3_PORT, TEST2_PORT, // PORT2 <-> PORT3
+        #endif
+        #if (NUM_COM_PORTS > 5)
+            TEST5_PORT, TEST4_PORT, // PORT4 <-> PORT5
+        #endif
+};
+
+/** STEP 3: Implementations of our core functions for this custom virtual port
```
kylemallory, line 20: This jumps to "Step 3" but there are no Step 1 and Step 2 in this file.

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.h
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+#ifndef CUSTOM_VIRTUAL_PORT_H
+#define CUSTOM_VIRTUAL_PORT_H
+
+/** STEP 1: Include IS core, other needed SDK header files here, and your own channel implementation files
+ * as required
+ */
+#include "core/base_port.h"
+#include "ring_buffer.h"  //optional, depends upon your implementation
```
kylemallory, line 18: does this exist in the current SDK stack?  Genuinely asking, I haven't checked.

kylemallory (changes requested):
I think this is getting really close.  The Step numbering seems a little disconnected, though maybe its because I am only seeing the changes.

I think overall, the README.md is missing a "theory of operation" the explains a summary of the nature of the "virtual port" as well as a overview of the implementation, etc.

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.cpp
shefft: I can understand it might be disconcerting to look at a file and not see steps laid out starting from 1 on up.  The reason for that is not every Step would apply to actions in every file, as the Steps are laid out at a project level.  From within the file, you can see the Step comment and always reference back to the README, and the README references each file which is described within the Step section so there is a tight and logical linkage.  I'm not beholden to this approach, it was an attempt to follow on to the way things had been done in previous examples and keep it familiar, but we can change it.

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.h
shefft: Yes, both of those files, the path is relative to `src/`

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.cpp
shefft: If we were to keep the Step system but re-orient the Steps to be processed at file level instead of project level, my proposal would probably be to divide the README project construction sections to be headed by file name instead of Steps (and if numbered, using numbering independent of the new Steps), and then put a series of Steps inside each of those sections so that each file has  Steps starting at 1 and going up.

ExampleProjects/CustomPort/CustomVirtual/README.md
```diff
@@ -1,39 +1,80 @@
...
+
+One is the `base_port` which is a C struct object that provides a data-in and data-out genericized interface to layer on top of the data transportation channel underneath that the user creates, custom to their application.  It “knows the essential things” about a port.  It provides a framework for maintaining port status and performing port operations.  Many channel implementations can be used and the header file provides port definitions used across the entire product line & SDK.  
+
+The other is the `PortFactory` which is an abstract C++ class that is responsible for discovery of available ports of a particular type.  There should be one implementation for each type of discoverable port.  This does NOT return a port, only a name or some other identifier that can be used by the port implementation to create the port.  Each factory is responsible for identifying & validating possible ports of a particular type.  Also responsible for allocating, configuring, and deallocating a port of that type, by its name.  As an abstract class, this allows for third-party locators to be implemented for custom port types.
```
kylemallory: I'm not sure that this statement about "only a name or some other identifier" is correct. The {Custom}PortFactory is also responsible for allocating and deallocating the port.

The PortFactory informs the PortManager of which ports are available.
The PortManager requests that the PortFactory in turn bind/allocate the ports, if needed.
The PortManager requests that the PortFactory release/de-allocate the ports when no longer needed.

tmcd249 (approved):

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+static int customPortWrite(port_handle_t port, const unsigned char* buf, unsigned int len)
+{
+    custom_port_t* destPort = boundPorts[portId(port)];
+
+    if (ringBufWrite(&destPort->portRingBuf, (unsigned char*)buf, len))
+    {   
+        // Buffer overflow
+        throw new std::out_of_range(utils::string_format("customPortWrite ring buffer overflow on %s: %d !!!\n",
```
Hephaestus-Maximus: **Blocking:** `throw new std::out_of_range(...)` throws a pointer (`std::out_of_range*`), not a `std::out_of_range` value. A typical `catch (const std::out_of_range&)` will not catch this — it'll propagate to `std::terminate()` — and even if it were caught by pointer, the `new`'d object would leak. Since this is customer-facing example code meant to teach the recommended pattern, it's teaching a C++ anti-pattern.

Also, every other `base_port_t` hook in this SDK (including this one, given `PORT_ERROR__WRITE_FAILURE` is defined right below) communicates failure via return codes, not exceptions — the `return PORT_ERROR__WRITE_FAILURE;` right after the `throw` is dead code, which is a tell that the throw wasn't the intended error path. Recommend dropping the `throw` entirely and just returning `PORT_ERROR__WRITE_FAILURE`, consistent with `customPortRead`/`customPortFree`/etc. and the rest of the base_port_t API contract.

ExampleProjects/CustomPort/CustomVirtual/README.md
```diff
@@ -1,39 +1,106 @@
...
+```
+
+The following implementation instructions identify some examples of similar code to that found under corresponding "STEP X" markings in the source files.  Please refer to the source file code directly and **treat the incomplete snippets of code in this README as orientation only**.
 
 ## Files
 
 #### Project Files
 
-* [CustomVirtualExample.cpp](https://github.com/inertialsense/inertial-sense-sdk/tree/release/ExampleProjects/CustomPort/CustomVirtual/CustomVirtualExample.cpp)
+* [main.cpp](https://github.com/inertialsense/inertial-sense-sdk/tree/release/ExampleProjects/CustomPort/CustomVirtual/CustomVirtualExample.cpp)
```
Hephaestus-Maximus: **Nitpick:** the link text was updated from `CustomVirtualExample.cpp` to `main.cpp`, but the URL still points at `.../CustomVirtualExample.cpp`. Leftover from the file rename — should point at `main.cpp`. (Related to the broader point in the review body about `CustomVirtualExample.cpp` still being present but no longer built.)

Hephaestus-Maximus (changes requested):
In-depth review of the custom-port example rewrite (virtual test port replacement + `PortManager` demonstration). Verified against a checkout of the branch (not just the diff) — cross-referenced `ring_buffer.c`, `base_port.h`, and `PortManager.h`/`.cpp` in the SDK to confirm behavior, and confirmed CI is green on this exact head commit (`f563102`).

**Overall:** the rewrite is a real improvement — dropping the dependency on `${IS_SDK_DIR}/tests/test_serial_utils.*` (code customers shouldn't need to reach into) in favor of a fully project-local `custom_virtual_port.*`, and adding the `PortManager` walkthrough, both make this a better teaching example than before. Kyle's earlier "needs a theory of operation" ask also looks addressed — the new Purpose and Design section covers `base_port_t` / `PortFactory` / `PortManager` roles well.

Two things I'd like fixed before merge, left as inline comments:

1. **`custom_virtual_port.cpp` `customPortWrite()`** — `throw new std::out_of_range(...)` throws a pointer, not a value; a normal `catch (const std::out_of_range&)` won't catch it and it'll terminate the program, plus it leaks the allocation either way. It's also inconsistent with the return-code convention the rest of `base_port_t` uses (the `return PORT_ERROR__WRITE_FAILURE;` right after the throw is unreachable, which is the tell). Since this is the pattern customers will copy, I'd rather see it just return the error code like its siblings do.

2. **Orphaned `CustomVirtualExample.cpp`** — it's no longer referenced by `CMakeLists.txt` (superseded by `main.cpp`), but it's still in the tree and was still edited in this PR (STEP renumbering). Anyone browsing the folder will find two near-duplicate entry points, one of which is dead. The rename commit's own message ("...need to do the same wit[h]...", truncated) suggests this was already flagged as unfinished — just calling it out so it doesn't slip through before squash-merge. Also flagged inline: the README's Files list still links `main.cpp`'s bullet to the old `CustomVirtualExample.cpp` URL.

Smaller, non-blocking:

- `custom_virtual_port.h` comments `#include "com_manager.h"` as "optional, depends upon your implementation," but `NUM_COM_PORTS` (used for `g_customPorts[NUM_COM_PORTS]`) only resolves because `com_manager.h` transitively pulls in `data_sets.h`. A customer who takes the "optional" note at face value and drops that include for a simpler port type will hit a compile error. Worth a one-line callout, or include `data_sets.h` directly.
- The `static` function prototypes in `custom_virtual_port.h` (`customPortRead`, `customPortWrite`, etc.) are only ever used inside `custom_virtual_port.cpp`. Declaring them in the shared header exposes internal-only symbols to every including TU for no benefit — could just live at the top of the `.cpp`.

Everything else — `CMakeLists.txt`, the `PortFactory` implementation (`bindPort`/`validatePort`/`locatePorts`/`releasePort`), and the `main.cpp` `PortManager` usage — checked out against the actual `PortManager.h` API (`addPortFactory`, `discoverPorts`, `getPort`) and matches; the described destructor-driven `releasePort` behavior in the README log sample is accurate too.

Per the PR description, this is intended to be squash-merged on approval, so I've focused on the cumulative diff rather than the individual WIP commits.

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.cpp
shefft: Will address this.

shefft: Done

ExampleProjects/CustomPort/CustomVirtual/README.md
```diff
@@ -49,177 +116,208 @@ typedef struct base_port_s {
...
+    {   
+   //...
 }
+//etc
 ```
 
-test_serial_utils.cpp also provides an initialization function that links the handles of the test port channel implementation to the base_port, allowing us to use the generic base_port API regardless of the details of the underlying channel implementation:
+These will provide the underlying functionality for the `base_port_t` extension once properly assigned to the base port's API.  As you review this example project or complete your own port implementation, it may be useful to compare [custom_virtual_port.h](https://github.com/inertialsense/inertial-sense-sdk/tree/release/ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.h) and [base_port.h](https://github.com/inertialsense/inertial-sense-sdk/tree/release/ExampleProjects/CustomPort/CustomVirtual/base_port.h) to see how the base port is defined and extended, and understand the API.  In this example, we have demonstrated only a minimal subset of the available port functions in `base_port_t`, but there are many options beyond simply read and write to enable the usage of a wide variety of underlying transport mechanisms.
```
Hephaestus-Maximus: Nitpick, newly introduced with this round's edits: this `base_port.h` link points at `.../ExampleProjects/CustomPort/CustomVirtual/base_port.h`, which doesn't exist — `base_port.h` lives at `src/core/base_port.h` (as correctly linked a few lines up in the Files list and in Step 1). Should point at `https://github.com/inertialsense/inertial-sense-sdk/tree/main/src/core/base_port.h`.

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.cpp
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @author TylerS
+ * @remark Originated as Walt Johnson's tests/test_serial_utils.cpp
+ * @copyright Copyright (c) 2026 Inertial Sense, Inc. All rights reserved.
+ */
+
+/** Include C++ library and other needed SDK header files here
+ */
+#include "custom_virtual_port.h"
+#include "util/util.h"
```
Hephaestus-Maximus: Minor cleanup: now that the `throw new std::out_of_range(...)` (which used `utils::string_format`) is gone, this include is unused — nothing else in the file references `util/util.h`. Non-blocking, just flagging so it doesn't linger.

Hephaestus-Maximus (approved):
Re-reviewed following the fix-up commits (`bcd98bd`, `d2f742e`, `5bc9c33`, `75a0622`, `50164b4`) against my prior request-changes review at `f563102`. Checked out the branch and diffed the two commit points directly rather than re-reading the whole PR.

Both blocking items from the previous review are resolved correctly:

1. `customPortWrite()` no longer throws — it now just `return`s `PORT_ERROR__WRITE_FAILURE` on overflow, matching the rest of `base_port_t`'s return-code convention, and `main.cpp` now checks `wbytes < 0` and handles it. Good fix.
2. `CustomVirtualExample.cpp` has been deleted (no longer orphaned/dead), and the README's `main.cpp` link now correctly points at `main.cpp` instead of the old file.

CI is green on the current head (`50164b4`).

Two very minor, non-blocking items found while re-reviewing, left inline — neither blocks approval:
- A `base_port.h` link added in this round's README wording points into the example folder instead of `src/core/base_port.h` (the file doesn't exist at the linked path).
- `custom_virtual_port.cpp`'s `#include "util/util.h"` is now unused now that the `utils::string_format(...)` call it supported has been removed.

Approving — the substantive issues are addressed and CI passes.

ExampleProjects/CustomPort/CustomVirtual/README.md
shefft: Resolved in the next commit

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.cpp
shefft: Resolved with latest commit

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.cpp
```diff
@@ -116,9 +109,20 @@ void CustomVirtualPortFactory::locatePorts(std::function<void(PortFactory*, uint
...
     log_msg(IS_LOG_PORT_FACTORY, IS_LOG_LEVEL_INFO, "Locating ports with regex pattern '%s'", pattern.c_str());
 
-    /** Call user implemented function to populate portNames as present on system; can use count returned if desired
+    /** User implementation to populate names of ports as present on system
+     */
+    std::vector<std::string> portNames = {};
+    portNames.clear();
+    portNames.resize(NUM_COM_PORTS);
+    
+    /** Grab each string with the unique identifying names the underlying test port implementation dictates
```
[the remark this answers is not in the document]
shefft: Agreed, this was hard coded and the next commit will change it to use pType.

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.h
```diff
@@ -0,0 +1,82 @@
...
+#include "com_manager.h" //optional, depends upon your implementation
+
+/**
+ * PORT IMPLEMENTATION used for unit and functional tests
+ * This is a generic port implementation that provides both bridging and loopback capability
+ * There are 6 ports defined:
+ *      LOOPBACK (TEST0_PORT, TEST1_PORT):
+ *         All data that is written to the port is placed into a ringbuffer which feeds subsequent reads
+ *         If no data has been written, there is not data to be read.  If you want to test synchronous
```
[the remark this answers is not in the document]
shefft: Original comment from old test port code, I'm leaving as is, no wrong information

ExampleProjects/CustomPort/CustomVirtual/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * For this simple example, we pass an arbitrary string into a loopback virtual test port
+     * underlying implementation for base_port from Inertial Sense, and compare the read back. 
+     */
+    while ( portIsOpened(port) && (run_cnt > 0) ) {
+        run_cnt--;
+        rbytes = 0;
+        wbytes = 0;
+
+        log_msg(IS_LOG_PORT, IS_LOG_LEVEL_INFO, "Attempting to send msg '%s'", wbuf);
```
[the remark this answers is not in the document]
shefft: Made suggested change into next commit

ExampleProjects/CustomPort/CustomVirtual/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+        if (portFree(port) >= wlen) {          // write fixed message if space available
+            wbytes = portWrite(port, wbuf, wlen);
+            
+            // write error check
+            if ( wbytes < 0 ) {
+                printf("Exiting with write error\r\n");
+                if ( wbytes == PORT_ERROR__WRITE_FAILURE )
+                    printf("PORT_ERROR__WRITE_FAILURE\r\n");
+                exit(0);
+            }
```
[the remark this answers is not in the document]
shefft: Made suggested change into next commit

ExampleProjects/CustomPort/CustomVirtual/README.md
```diff
@@ -49,177 +116,208 @@ typedef struct base_port_s {
...
+   int portNum = 0;
+   for (custom_port_t& port : g_customPorts) {
+      port.base.pnum = portNum;
+      port.base.ptype = PORT_TYPE__COMM;
+      if (portNum <= 1)
+         port.base.ptype |= PORT_TYPE__LOOPBACK;  // only PORT0 and PORT1 are Loopbacks
+
+         port.base.portRead = customPortRead;
+         port.base.portWrite = customPortWrite;
+         //etc
```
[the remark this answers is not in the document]
shefft: Made suggested change into next commit

ExampleProjects/CustomPort/CustomVirtual/custom_virtual_port.h
```diff
@@ -0,0 +1,82 @@
...
+ *         functionality, use a loopback and perform all your writes, then read from the same port to
+ *         ensure the data was written (or is read/parsed) correctly.
+ *
+ *      BRIDGE (TEST2_PORT <-> TEST3_PORT, TEST4_PORT <-> TEST5_PORT):
+ *         All data that is written to the port of written into the ringbuffer of the paired port.
```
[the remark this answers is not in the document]
shefft: Made suggested change into next commit

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.cpp
```diff
@@ -98,11 +85,17 @@ bool CustomVirtualPortFactory::releasePort(port_handle_t port) {
...
-    if (pType != (PORT_TYPE__LOOPBACK | PORT_TYPE__COMM) )           
-            return false;   // we can only validate this port type - all others fail
-        
-    /** Add here any other custom validation that could/should be done on this port type
-     */
+    if (pType != (PORT_TYPE__LOOPBACK | PORT_TYPE__COMM) ) {
+        log_msg(IS_LOG_PORT_FACTORY, IS_LOG_LEVEL_INFO, "Port type validation failed: %d", pType );
+        return false;   // we can only validate this port type - all others fail
+    }
+
```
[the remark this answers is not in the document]
shefft: Addressed for port count and made more robust, will leave the type alone for now as it filters as desired for demo.

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.cpp
```diff
@@ -85,8 +73,7 @@ bool CustomVirtualPortFactory::releasePort(port_handle_t port) {
     /** If you allocated your port object on the heap, free the memory (delete) here;
      * In this example we use a virtual port with the static object, so there is no memory to free, only clear
      */
-    //for example, delete (serial_port_t*)port;
-    memset(port, 0, sizeof(test_port_t));
+    memset(port, 0, sizeof(custom_port_t));      //or for example, delete (serial_port_t*)port;
```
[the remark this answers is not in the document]
shefft: We'll leave it this way for now for purposes of the demo, and assume that release of the port which only happens when PortManager is destroyed would necessitate re-instancing of PortManager and let's assume that will be accompanied by a new PortFactory anyway

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.cpp
```diff
@@ -98,11 +85,25 @@ bool CustomVirtualPortFactory::releasePort(port_handle_t port) {
...
+   
+    /** Check port name to make sure it would be found, matching the naming described in the custom port definition */    
+    const std::regex pattern("^TEST([0-9]+)$");
+    std::smatch m;
+    if (! std::regex_match(pName, m, pattern) ) {
+        log_msg(IS_LOG_PORT_FACTORY, IS_LOG_LEVEL_INFO, "Port name validation failed: '%s'", pName.c_str() );
+        return false;
+    }
+
+    int index = std::stoi(m[1].str());
```
Hephaestus-Maximus, line 102: Minor, non-blocking: `std::stoi` throws `std::out_of_range` for a numeric string too large to fit in an `int` (e.g. a malformed/adversarial CLI port arg like `TEST99999999999999999999`), which would be an uncaught exception here. Vanishingly unlikely to matter for this demo's actual inputs, but since `pName` does ultimately come from `argv[1]` via `bindPort()`, worth a passing mention in case a customer copies this validation pattern verbatim for a case where the input is less trusted.

Hephaestus-Maximus (approved):
Re-reviewed after another round of fixes since my last approval. Diffed directly against the commit I approved (`50164b4`) to isolate what's actually new, rather than re-reading the whole PR.

Both items from my previous follow-up review are resolved:
- The `base_port.h` link now correctly points at `src/core/base_port.h` instead of a nonexistent path in the example folder.
- The now-dead `#include "util/util.h"` is removed.

Additional independent improvements in this round, all verified correct:
- `locatePorts()` was hardcoding `PORT_TYPE__LOOPBACK | PORT_TYPE__COMM` in both the `validatePort()` call and the callback, ignoring its own `pType` parameter — now correctly uses `pType` throughout. Real fix (the function used to silently ignore what callers asked it to filter on).
- `validatePort()`'s port-name check now uses `^TEST([0-9]+)$` plus an explicit `index < NUM_COM_PORTS` bounds check, instead of the old hardcoded `^TEST[0-5]$` — more robust, correctly ties validation to `NUM_COM_PORTS` rather than a magic number.
- `main.cpp`'s write-failure path now `return 1;` instead of `exit(0)` — correctly signals failure via exit code.
- Minor `reinterpret_cast` fix for the `%s` format arg, and a doc-comment typo fix.

One very minor, non-blocking note left inline: `std::stoi` in the new bounds-check can throw on a pathologically long numeric suffix — essentially irrelevant for this demo's actual inputs, just flagging since it's the kind of thing a customer might copy verbatim into a context with less-trusted input.

CI is green on both Linux and Windows. Approving.

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.cpp
shefft: The throw and lack of handling for the exception is probably satisfactory behavior for this simple demo, so let's keep moving.

waltjohnson (approved):

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.h
```diff
@@ -7,58 +7,43 @@
...
         static CustomVirtualPortFactory instance;
         return instance;
     }
 
+    /** Default constructor */
+    CustomVirtualPortFactory() {
+        /** STEP 9: On construction, set the SDK message logger verbosity level, as we use the logger in this class for
+         * some custom status and error messages
+         */
+        IS_SET_LOG_LEVEL(IS_LOG_LEVEL_INFO);
```
kylemallory: You shouldn't do this here... this should be done in the application/main.cpp -- otherwise, putting this here (in any real-world situation) risks overriding another valid attempt to set the level - but it does it in a singleton constructor.

We should probably discuss what your rationale was for this, and specifically for placing it here.

ExampleProjects/CustomPort/CustomVirtual/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+     * we'll make local references, and we register the virtual port factory
+     */
+    PortManager& pm = PortManager::getInstance();
+    CustomVirtualPortFactory& vpf =  CustomVirtualPortFactory::getInstance();
+    pm.addPortFactory(&vpf);
+          
+    /** We are interested in finding all ports matching a certain name pattern, but then we'll reference the one
+     specifically indicated on the command line, which is virtual loopback in this case
+    */
+    pm.discoverPorts(vpf.discoverPattern, PORT_TYPE__COMM | PORT_TYPE__LOOPBACK);
```
kylemallory: This is correct - but for the purposes here, it simplifies a bit to just call discoverPort() without any arguments, which should return all ports that the registered factories know about (rather than applying a specific filter).

ExampleProjects/CustomPort/CustomVirtual/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+    port_handle_t port = pm.getPort(argv[1], PORT_TYPE__COMM | PORT_TYPE__LOOPBACK);
```
kylemallory: At this point, all we are really interested in in the portName, so I think you can just specify getPort(argv[1]), and it should be sufficient, and simplifies the called (though, again, this is technically still correct).

ExampleProjects/CustomPort/CustomVirtual/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+    CustomVirtualPortFactory& vpf =  CustomVirtualPortFactory::getInstance();
+    pm.addPortFactory(&vpf);
+          
+    /** We are interested in finding all ports matching a certain name pattern, but then we'll reference the one
+     specifically indicated on the command line, which is virtual loopback in this case
+    */
+    pm.discoverPorts(vpf.discoverPattern, PORT_TYPE__COMM | PORT_TYPE__LOOPBACK);
+    port_handle_t port = pm.getPort(argv[1], PORT_TYPE__COMM | PORT_TYPE__LOOPBACK);    
+    
+    if (port == nullptr) {
```
kylemallory: You can perform this check to confirm it was allocated, but the more robust method is to call `portIsValid(port)` which should return true if it was initialized properly (even if its was a non-null pointer it could still be invalid).

ExampleProjects/CustomPort/CustomVirtual/main.cpp
```diff
@@ -0,0 +1,134 @@
...
+        printf("Failed to allocate port\r\n");
+        return -2;
+    }
+
+    /** Binding a port does not open a port.. so let's open it using base_port functions */
```
kylemallory: Don't mention "binding" here - since we didn't actually call the bindPort() call.  But its true, that discoverPorts() and getPort() DO NOT open the port.

kylemallory (changes requested):
A few, very minor, changes needed.

ExampleProjects/CustomPort/CustomVirtual/main.cpp
shefft: This was inherited.  Updated in next commit to use portIsValid instead.

ExampleProjects/CustomPort/CustomVirtual/main.cpp
shefft: This was inherited.  Revising in the next commit.

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.h
shefft: Work in mid-stream.  I realized this command had been left out previously and threw it in just before commit as part of our decision on moving logging all into the PF, dropped in this spot momentarily just to test it and remind myself of the move for the next day's wrap up and the rest of the work that needed doing.  The step # is not even correct because the needed re-structuring of the README and code comments was still to come.  IMO also this wouldn't be the best place for such a command ultimately anyway, and the code and README are all addressed in the next commit.

ExampleProjects/CustomPort/CustomVirtual/README.md
shefft: Older comment on old copy, which has been revised.

ExampleProjects/CustomPort/CustomVirtual/main.cpp
shefft: Went for the simplification and moved the filtering into the port factory on locate instead.

ExampleProjects/CustomPort/CustomVirtual/main.cpp
shefft: Went for the simplification of these PM calls.

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.cpp
```diff
@@ -80,14 +68,14 @@ bool CustomVirtualPortFactory::releasePort(port_handle_t port) {
...
+    log_msg(IS_LOG_PORT_FACTORY, IS_LOG_LEVEL_INFO, "Release comm port '%s'", portName(port) );
+        
     /** If you allocated your port object on the heap, free the memory (delete) here;
-     * In this example we use a virtual port with the static object, so there is no memory to free, only clear
+     * or clear only for static allocation
      */
-    //for example, delete (serial_port_t*)port;
-    memset(port, 0, sizeof(test_port_t));
-
+    memset(port, 0, sizeof(custom_port_t));
```
Hephaestus-Maximus, line 76: Minor, non-blocking: `memset(port, 0, sizeof(custom_port_t))` immediately before `delete` is redundant — `custom_port_t` is fully POD here (verified `ring_buf_t` in `ring_buffer.h` is a plain struct of raw pointers/ints, no non-trivial members anywhere in the union), so there's no destructor logic this could interfere with, and `delete` will free the memory regardless of what's in it. Not a bug, just dead work — could drop the `memset` (or add a one-line comment if it's intentionally there as a "wipe before free" habit worth modeling for customers, since that *would* matter for a real port type holding sensitive buffers).

ExampleProjects/CustomPort/CustomVirtual/README.md
```diff
@@ -49,177 +116,219 @@ typedef struct base_port_s {
...
-        port.base.portAvailable = testPortAvailable;
-        port.base.portName = testPortName;
+static int customPortRead(port_handle_t port, unsigned char* buf, unsigned int len)
+{
+    return ringBufRead(&((custom_port_t*)port)->portRingBuf, buf, len);
+}
+
+static int customPortWrite(port_handle_t port, const unsigned char* buf, unsigned int len)
+{
+    custom_port_t* destPort = boundPorts[portId(port)];
```
Hephaestus-Maximus: This illustrative snippet doesn't match the actual `customPortWrite()` in `custom_virtual_port.cpp` — the real implementation is:
```cpp
custom_port_t* destPort = static_cast<custom_port_t*>(port);  //loopback
```
`boundPorts` and `portId(...)` don't exist anywhere in this project or the SDK (grepped both) — a customer trying to cross-reference this snippet against the real file will be confused about where `boundPorts` is supposed to come from. I know the README says to "treat the incomplete snippets ... as orientation only," but this one doesn't just simplify the real code, it shows a different (and non-existent) mechanism entirely. Worth updating to match the actual self-loopback cast, verified via an actual build+run of this project (both `TEST0` and `TEST1` loopback correctly, matching your documented log output exactly).

Hephaestus-Maximus (approved):
## Re-review: Custom Port example, dynamic single-port + PortManager (SN-6695)

Third re-review round from me on this PR. My prior two rounds (approved 2026-07-22 and 2026-07-25) covered the exception-throw fix, the orphaned `CustomVirtualExample.cpp` removal, the `locatePorts()`/`validatePort()` hardening, and CI. Kyle's `CHANGES_REQUESTED` on 2026-08-01 ("A few, very minor, changes needed") has no inline comments or elaborating text attached that I can find via the API — I couldn't determine what specifically he meant, so rather than guess, I re-verified the entire current head from scratch, independent of assuming his concern was addressed.

### What's new since my last approval
Per the description's "2/Aug" edit: replaced the multi-port init helper with single-port, dynamically-allocated binding (one `custom_port_t` per `bindPort()` call instead of a static array), moved naming/type into that init path, switched application messaging to stdout with logging reserved for the SDK-facing factory code, and simplified the `PortManager` integration.

### Verified by actually building and running the project (not just reading the diff)
Hit a pre-existing CMake quirk in this sandbox where the "mirror consumer's build-dir name under IS_SDK_DIR" convention (`include_is_sdk_find_library.cmake`) fails with "no rule to make target .../build/all" — confirmed this is **not specific to this PR** by reproducing the identical failure on the unmodified, sibling `ExampleProjects/ISComm` project. Worked around it by building `InertialSenseSDK` via the working `tests/` CMake path and compiling+linking the three `CustomVirtual` source files directly with `g++ -std=c++20`. Clean compile, zero warnings.

Ran the resulting binary for real:
```
$ ./CustomVirtual TEST0
Attempting to bind and open virtual port TEST0
Attempting to send msg 'IMPORTANT MESSAGE'
Loopback test good on comm port 'TEST0', 17 bytes sent/recvd
Program complete, see inertial_sense.log for more results details
```
Matches the README's documented console output exactly. Also tested `TEST1` (works), an invalid port name (`BOGUS` — cleanly rejected with "Failed to allocate port", exit code -2, no crash), and no-args (correct usage message). The generated `inertial_sense.log` matches the README's documented log sample byte-for-byte in structure — confirms `PortManager::discoverPorts()` really does bind *both* `TEST0` and `TEST1` (not just the requested one) and releases both on exit via the `PortManager` destructor, exactly as the README's Step 8 claims.

### Two new, non-blocking nitpicks (posted inline)
- `CustomVirtualPortFactory::releasePort()` does `memset(port, 0, sizeof(custom_port_t))` immediately before `delete` — redundant since `custom_port_t` is fully POD (verified `ring_buf_t` is a plain-pointer struct), so there's no destructor logic to protect and `delete` frees the memory regardless. Harmless, just dead work.
- The README's `customPortWrite()` example snippet (Step 3) shows `custom_port_t* destPort = boundPorts[portId(port)];` — this doesn't match the actual implementation (`static_cast<custom_port_t*>(port)`, a self-loopback) and `boundPorts`/`portId()` don't exist anywhere in the project or SDK. Given the README explicitly says these snippets are "orientation only," this is more a documentation-accuracy nit than a functional issue, but it could genuinely confuse someone cross-referencing the real file.

### Verdict
Approving. Both blocking items from my first review round are still resolved, the newest architectural change (dynamic single-port binding via `PortManager`) works correctly end-to-end (built, ran, verified output against docs), and CI is green on both platforms. The two nitpicks above are cosmetic and don't need to hold up merge.

ExampleProjects/CustomPort/CustomVirtual/CustomVirtualPortFactory.cpp
shefft: This serves as illustration for users and follows the pattern of other example projects that use both a mem clear and a delete.  However, I don't care if someone wants to get rid of it, but will just leave it for now.

ExampleProjects/CustomPort/CustomVirtual/README.md
shefft: Great catch.  Fixed for the next commit.

Hephaestus-Maximus:
## Failure-path testing: instrumented every branch, captured real console+log output for each

Following up on my prior approval with deeper verification: I mapped every failure/error branch reachable from `main.cpp`, `CustomVirtualPortFactory.cpp`, and `custom_virtual_port.cpp`, and confirmed each one's actual behavior by running the built binary — not by reading the code and assuming. Five of the eight branches below aren't reachable through the CLI at all (the fixed `portNames = {"TEST0","TEST1"}` list means `PortManager` never calls into most of `validatePort()`'s own checks), so I wrote a small scratch driver (`test_failure_paths.cpp`, compiled and run locally, **not committed — deleted after this session, not part of the diff**) that calls `CustomVirtualPortFactory`/`base_port` directly to exercise those. Full output below for all 8.

Built via the same workaround as my last review (`tests/` CMake path for `InertialSenseSDK`, then `g++ -std=c++20` linking the three `CustomVirtual` sources directly against it — the pre-existing, PR-unrelated `ExampleProjects` CMake issue is still there).

### 1. No arguments — `main.cpp:37-42`
```
$ ./CustomVirtual
Please pass the virtual loopback port as the only argument (i.e. TEST0 or TEST1)
$ echo $?
255   # (-1)
```
No log file created (exits before any logging).

### 2/3. Valid port names `TEST0` / `TEST1` — happy path, for contrast
```
$ ./CustomVirtual TEST0
Attempting to bind and open virtual port TEST0
Attempting to send msg 'IMPORTANT MESSAGE'
Loopback test good on comm port 'TEST0', 17 bytes sent/recvd
Program complete, see inertial_sense.log for more results details
$ echo $?
0
```
```
inertial_sense.log:
[03:25:20.206025] INFO   (IS_LOG_PORT_FACTORY) :: Locating ports with regex pattern '(.+)'
[03:25:20.207008] INFO   (IS_LOG_PORT_FACTORY) :: Bind new comm port 'TEST0'
[03:25:20.207853] INFO   (IS_LOG_PORT_FACTORY) :: Bind new comm port 'TEST1'
[03:25:21.208105] INFO   (IS_LOG_PORT_FACTORY) :: Release comm port 'TEST0'
[03:25:21.208182] INFO   (IS_LOG_PORT_FACTORY) :: Release comm port 'TEST1'
```
(`TEST1` produces the identical pattern.) Confirms `discoverPorts()` binds *both* ports regardless of which was requested, and both get released on exit.

### 4. Invalid port name via the CLI (`BOGUS`) — `main.cpp:67-70`
```
$ ./CustomVirtual BOGUS
Attempting to bind and open virtual port BOGUS
Failed to allocate port
$ echo $?
254   # (-2)
```
```
inertial_sense.log:
[03:25:33.486998] INFO   (IS_LOG_PORT_FACTORY) :: Locating ports with regex pattern '(.+)'
[03:25:33.488144] INFO   (IS_LOG_PORT_FACTORY) :: Bind new comm port 'TEST0'
[03:25:33.489138] INFO   (IS_LOG_PORT_FACTORY) :: Bind new comm port 'TEST1'
[03:25:33.489225] INFO   (IS_LOG_PORT_FACTORY) :: Release comm port 'TEST0'
[03:25:33.489249] INFO   (IS_LOG_PORT_FACTORY) :: Release comm port 'TEST1'
```
**Worth knowing:** there's no "validation failed" log line here. Traced why: `PortManager::getPort("BOGUS")` (`PortManager.cpp:147-154`) is a plain lookup against `knownPorts`, populated only by whatever `locatePorts()` enumerated (`TEST0`, `TEST1`). `"BOGUS"` was never in that set, so this returns `nullptr` immediately — `CustomVirtualPortFactory::validatePort()` is **never called** for this case. The exit code and message are correct, but this test exercises `PortManager`'s lookup-miss, not the factory's own validation logic. Confirmed directly (`portmanager_getport_miss` below reproduces the identical log, proving no factory call happens).

### 5-7. `validatePort()`'s three internal checks — not reachable via the CLI, called directly
```
$ ./test_failure_paths validate_wrong_type
validatePort("TEST0", 0x0001) returned: false
inertial_sense.log: [03:25:43.478929] ERROR  (IS_LOG_PORT_FACTORY) :: Port type validation failed: 1

$ ./test_failure_paths validate_bad_regex
validatePort("COM3", LOOPBACK|COMM) returned: false
inertial_sense.log: [03:25:43.486813] ERROR  (IS_LOG_PORT_FACTORY) :: Port name validation failed: 'COM3'

$ ./test_failure_paths validate_index_oor
validatePort("TEST99", LOOPBACK|COMM) returned: false
inertial_sense.log: [03:25:43.494233] ERROR  (IS_LOG_PORT_FACTORY) :: Port index out of range: 'TEST99'
```
All three branches (`CustomVirtualPortFactory.cpp:89-106`) behave exactly as coded, each producing its own distinct `ERROR`-level log line. `bindPort("BOGUS", ...)` called directly reproduces the regex-mismatch line too (`Port name validation failed: 'BOGUS'`) — confirming scenario 4's silence is a `PortManager` characteristic, not a factory bug.

### 8. `portOpen()` — dead code in the current implementation
```
$ ./test_failure_paths portopen_deadcode
portIsValid(port)  = 1
portIsOpened(port) = 1  (already true at bind time)
portOpen(port) called directly anyway -> returned -1 (PORT_ERROR__NOT_SUPPORTED)
```
`initCustomPort()` (`custom_virtual_port.cpp:69-84`) never assigns `.portOpen`, and unconditionally sets `PORT_FLAG__OPENED` at bind time. So in `main.cpp:73`, `!portIsOpened(port)` is always `false`, short-circuiting `&&` before `portOpen()` is ever called — the `"Failed to open port"` / exit `-3` branch is unreachable as the code stands today. Confirmed by calling `portOpen()` directly: it correctly returns `PORT_ERROR__NOT_SUPPORTED` (no function pointer assigned), which is exactly what would flow into that branch if it were ever reached. Non-blocking — flagging for awareness, not requesting a change on an already-approved PR.

### 9. Write-buffer overflow — `custom_virtual_port.cpp:31-41`, main.cpp's write-error branch
```
$ ./test_failure_paths write_overflow
Attempting to bind and open virtual port TEST1
Attempting to send msg of 8193 bytes (buffer capacity is 8192)
Exiting with write error
PORT_ERROR__WRITE_FAILURE
$ echo $?
252   # (-4)
```
One byte over `PORT_BUFFER_SIZE` (8192) reliably trips `ringBufWrite()`'s overflow flag, `customPortWrite()` returns `PORT_ERROR__WRITE_FAILURE` (-5), and `main.cpp`'s existing check (`wbytes < 0` → print both messages → `return -4`) fires exactly as written. No crash, no corruption — the ring buffer wraps correctly even when told to overwrite unread data.

### 10. `releasePort(nullptr)` — null-guard
```
$ ./test_failure_paths release_null
releasePort(nullptr) returned: false (expect false, no crash, no log line)
```
Confirmed: returns `false` immediately, no log line emitted, no crash.

## Summary

Every failure branch in this example behaves correctly and matches its own code exactly — no surprises, no crashes, no corruption anywhere, including in the two paths (`portOpen()`, buffer overflow) that required calling internals directly since the CLI/PortManager combination can't reach them. One non-blocking observation worth a maintainer's awareness: the `"Failed to open port"` branch in `main.cpp` is currently dead code given how `initCustomPort()` sets `PORT_FLAG__OPENED` — not something I'd hold up merge for, just noting it since nothing in this or my prior review exercised it before now.

Hephaestus-Maximus (approved):
## Re-review, round 4: Custom Port example

Diffed directly against my last-approved commit (`1b9696d8`) rather than re-reading the whole PR.

### README nitpick resolved
The `customPortWrite()` example snippet now correctly shows `static_cast<custom_port_t*>(port)` instead of the stale `boundPorts[portId(port)]` I flagged last round. Matches the real implementation.

### New change, verified good — and actually a nice catch
`main.cpp`'s port-open check changed from `if (!portIsOpened(port) && (portOpen(port) != PORT_ERROR__NONE))` to `if (!portIsOpened(port))`. Checked why: `initCustomPort()` never assigns `base.portOpen` — grepped `custom_virtual_port.cpp`/`.h`, no such assignment exists. The *old* code's `portOpen(port)` call would only ever fire through an unassigned function pointer, but it never actually did, because `portIsOpened(port)` was already `true` at that point (`initCustomPort()` sets `PORT_FLAG__OPENED` at bind time) — so the `&&` short-circuit always skipped it. It was latent, unreachable-in-practice code that would have been a real landmine (crash via a null function-pointer call) for anyone who copied this pattern into a port that *isn't* pre-marked open at init. Removing it and checking the flag directly is a genuine improvement, not just a rewording.

Rebuilt and re-ran the current head end-to-end (same method as last round, working around the sandbox's pre-existing CMake quirk) — `TEST0`/`TEST1` still loopback correctly, log output unchanged. No regressions.

The `const uint16_t pType` parameter tweak and the removed dead commented-out `g_customPorts[]` line are cosmetic, no concerns.

Approving again.

Hephaestus-Maximus (approved):
## Re-review, round 5: Custom Port example

Diffed against my last-approved commit (`9a9db419`). This round is documentation-only — README link cleanup and matching wording to the code changes from last round.

Verified:
- The `initCustomPort()` signature snippet and the `portOpen()`-removal snippet in the README now match the actual code exactly (both were updated last round in the code but not yet in the README's illustrative snippets — now consistent).
- The documented console-output sample was updated to match `main.cpp`'s new printf wording ("Attempting to bind and **use**..." vs. "...and open...") — rebuilt and ran the current head, output matches exactly.
- CI green on both platforms.

One minor, non-blocking observation: the "Files" list section was converted to relative links (`./main.cpp`, `../../../src/...`), which is a nice portability improvement, but the inline links throughout the "Implementation" walkthrough sections (e.g. the `custom_virtual_port.cpp` links) are still absolute `github.com/.../tree/release/...` URLs — so the README now has two different link styles for the same files. Not a regression (those were already like that), just an incomplete pass given the commit's stated goal — worth a follow-up if there's a next round, not blocking this one.

Approving.

tmcd249 (approved):

kylemallory (approved):

PR #1213, opened 2026-07-10
SN-8285: Fixes field 15 of powtlv to be 0-360 not -180 to 180

tmcd249:
- (SDK) (SN-8285, PR-1213) Fixed field 15 of $POWTLV to be 0 to 360 not -180 to 180


src/protocol_nmea.cpp
```diff
@@ -1296,7 +1296,12 @@ int nmea_powtlv(char a[], const int aSize, gnss_pos_t &pos, gnss_vel_t &vel)
 
     nmea_sprint_f(a, aSize, n, ",%.3f", vel.vel[2]);            // 14
 
+    extern debug_array_t          g_debug;
```
kylemallory: Why was g_debug array added here?

kylemallory (approved):

shefft (approved):

PR #1222, opened 2026-07-15
SN-8268 fix Simple_Discovery example project build, and add to ExProj meta-build

shefft:
There were changes made to ISDevice that affected the Example Projects like Simple_Discovery.  The only problem turned out to be the original build issue I ran into on my machine with shared pointer vs classic pointer mismatch, which is fixed here, other runtime issues were found to be local environmental/build related and were resolved.

Added the Simple_Discovery project to the ExampleProjects cmake commands so that it is included in the meta-build, and verified that executable runs.  This is done because in the ExampleProjects documentation it seems to indicate that all existing examples should be built as part of the meta-build.

After approval, will SQUASH-MERGE into develop.

----
25th Jul update:  Regarding CI Windows build error not from this change, but inherited - used suggested Copilot fix and automated commit and merge from develop.  Windows build now completes successfully.  Just in case, requesting re-review from human approvers as well.


kylemallory (approved):

tmcd249 (approved):

ExampleProjects/Simple_Discovery/Example-2_Device-Discovery.cpp
```diff
@@ -143,4 +145,4 @@ int main(int argc, const char** argv) {
 #else
     return main_discovery(portPattern);
```
Hephaestus-Maximus, line 146: **Blocking — this is a live CI failure, not a hypothetical:** the "General Software Windows" check on this exact commit fails to compile this file:

```
Example-2_Device-Discovery.cpp(141): error C2065: 'portPattern': undeclared identifier
Example-2_Device-Discovery.cpp(146): error C2065: 'portPattern': undeclared identifier
```

Root cause, a few lines up (not touched by this PR's diff, but this PR is what first puts this file into the build on Windows, since `Simple_Discovery` was previously excluded from the meta-build):
```c
#if PLATFORM_IS_LINUX
    const char* portPattern = "(.+)";
#else
    const char* portArg = "COMM1";      // <-- declares portArg, not portPattern
#endif
    if (argc > 1)
        portPattern = argv[1];          // <-- unconditionally references portPattern
```
On the non-Linux (`#else`) branch, only `portArg` is declared, but `portPattern` is referenced unconditionally right after (and again here at `main_discovery(portPattern)`) — so it's undeclared on Windows/Mac. `portArg` itself is only ever used under `#ifdef EXPLAINED`, which isn't defined in the normal build, so it looks like a straightforward one-line typo: the `#else` branch should declare `portPattern` (e.g. `const char* portPattern = "COMM1";`), not `portArg`.

The PR description says "verified that executable runs" and the ticket explicitly asks to confirm it "builds both ways like the other working applications" — this suggests only the Linux/local path was actually exercised before the Windows CI result came back failing. Given the two approvals are timestamped before (Kyle, seconds before this failing run started) and after (Tony, over a day after the failure completed) the CI result, it looks like this may have been missed rather than knowingly accepted — flagging clearly either way since the fix is trivial but the current state would break the Windows build once merged.

Hephaestus-Maximus (changes requested):
Reviewed the `Simple_Discovery` build fix and meta-build integration. Checked CI results directly rather than just reading the diff, since this PR's entire purpose is build correctness.

**The `std::shared_ptr<ISDevice>` fix is correct** — matches the originally-reported GCC error exactly (`dm.getDevices().front()` returns a `shared_ptr<ISDevice>`, and `device` is now declared as one too; the rest of the function's usage — `!device`, `device->connect()`, etc. — is all consistent with shared_ptr semantics).

**Blocking: the "General Software Windows" CI check on this exact commit (`14ebd8a`) fails to compile**, and it's a real, reproducible bug, not flakiness — left the root cause and a one-line fix inline. Short version: `main()`'s `#if PLATFORM_IS_LINUX ... #else ... #endif` block declares `portArg` on the non-Linux branch but the code right after unconditionally references `portPattern`, which is undeclared there. This is pre-existing code this PR's diff doesn't touch, but it's dormant until now because `Simple_Discovery` was never previously part of the build — this PR is what first exercises it on Windows, and it fails immediately (confirmed identical, still-broken code is also on `develop`, so this isn't something that's since been fixed elsewhere).

Given the PR description says "verified that executable runs" and the ticket explicitly asks to confirm the example "builds both ways like the other working applications," this looks like the Windows leg just wasn't checked before merge — worth a look, since the fix is one line but leaving it as-is would break the Windows build the moment this merges.

Requesting changes on that basis; everything else here looks ready.

ExampleProjects/Simple_Discovery/Example-2_Device-Discovery.cpp
shefft: We committed a fix for this with Copilot's help

Hephaestus-Maximus (approved):
Re-reviewed after the fix for the Windows CI failure I flagged. The new commit does exactly what I suggested — renames `portArg` to `portPattern` on the non-Linux branch of `main()`'s platform conditional, so both branches now declare the same variable name that gets used unconditionally afterward.

Waited for CI to actually complete rather than trusting the diff alone (this is precisely the class of bug where "looks right" and "actually builds" can diverge): both **General Software Linux** and **General Software Windows** now pass on the current head commit (`fc2b364`) — confirmed the Windows job got past the `Build SDK Examples` step that failed before.

Approving — the fix is correct and now CI-verified on the platform that was actually broken.

ExampleProjects/Simple_Discovery/Example-2_Device-Discovery.cpp
shefft, line 143: Since there is no main_explained() in this example, it might make sense for another task to clean up this unnecessary code, as it was probably copied from another project.

waltjohnson (approved):

kylemallory (approved):

PR #1223, opened 2026-07-15
SN-8307: Fix DID_CANFD_CONFIG mapping under-covering can_config_t

kylemallory:
## Problem

The EvalTool debug build aborts during static initialization (SIGABRT) before the GUI appears:

```
DataMapper ERROR [12can_config_t]: mapped size 66 != struct size 168 (diff 102 bytes)
Assertion `false && "Size of mapped fields does not match struct size"' failed. (ISDataMappings.h:242)
```

## Root cause

`can_config_t` is **168 bytes** — its two arrays are sized `NUM_CIDS` (27): `can_period_mult[27]` (54B) + `can_transmit_address[27]` (108B) + `can_setting` (2B) + `can_receive_address` (4B).

`PopulateMapCanFdConfig` (DID_CANFD_CONFIG) maps only the CAN-FD subset — the first `NUM_FDCIDS` (10) slots of each shared array — summing to **66 bytes** (22 members). The `DataMapper` destructor requires mapped fields to span the whole struct (`totalSize == sizeof(struct)`), so `66 != 168` aborts startup in debug builds.

SN-8301 fixed the classic `PopulateMapCanConfig` (DID_CAN_CONFIG) mapper to fully cover all 27 slots, but left the FD mapper under-covering the shared struct.

## Fix

Add reserved `DATA_TYPE_BINARY` filler mappings for the FD-unused tail of both arrays (indices `NUM_FDCIDS..NUM_CIDS-1`), keyed off `NUM_FDCIDS`/`NUM_CIDS`. The FD mapping now spans the full 168 bytes (66 + 102 = 168).

## Validation

EvalTool debug build (develop) starts cleanly — no `DataMapper ERROR`, no assertion, no segfault, graceful shutdown.

Jira: SN-8307

### DIFF
--- a/src/ISDataMappings.cpp
+++ b/src/ISDataMappings.cpp
@@ -1998,6 +1998,13 @@ static void PopulateMapCanFdConfig(data_set_t data_set[DID_COUNT], uint32_t did)
     mapper.AddMember2("can_fd_transmit_address[FDCID_GNSS2_POS]",         offsetof(can_config_t, can_transmit_address) + sizeof(uint32_t) * FDCID_GNSS2_POS,         DATA_TYPE_UINT32, "", "FD transmit address for GNSS2 position",            DATA_FLAGS_DISPLAY_HEX);
     mapper.AddMember2("can_fd_transmit_address[FDCID_GNSS1_RTK_POS_REL]", offsetof(can_config_t, can_transmit_address) + sizeof(uint32_t) * FDCID_GNSS1_RTK_POS_REL, DATA_TYPE_UINT32, "", "FD transmit address for GNSS1 RTK relative",       DATA_FLAGS_DISPLAY_HEX);
     mapper.AddMember2("can_fd_transmit_address[FDCID_GNSS2_RTK_CMP_REL]", offsetof(can_config_t, can_transmit_address) + sizeof(uint32_t) * FDCID_GNSS2_RTK_CMP_REL, DATA_TYPE_UINT32, "", "FD transmit address for GNSS2 RTK compassing relative", DATA_FLAGS_DISPLAY_HEX);
+
+    // can_config_t is shared with DID_CAN_CONFIG, whose arrays are sized NUM_CIDS (27). CAN-FD
+    // only uses the first NUM_FDCIDS (10) slots of each array, so the mappings above cover just
+    // that subset. The DataMapper invariant requires the mapped fields to span the whole struct,
+    // so account for the FD-unused tail of both arrays as reserved regions.
+    mapper.AddMember2("can_fd_period_mult_reserved",      offsetof(can_config_t, can_period_mult)      + sizeof(uint16_t) * NUM_FDCIDS, DATA_TYPE_BINARY, "", "Reserved (classic-CAN period-mult slots unused in FD mode)",      DATA_FLAGS_READ_ONLY, 1.0, (uint32_t)(sizeof(uint16_t) * (NUM_CIDS - NUM_FDCIDS)));
+    mapper.AddMember2("can_fd_transmit_address_reserved", offsetof(can_config_t, can_transmit_address) + sizeof(uint32_t) * NUM_FDCIDS, DATA_TYPE_BINARY, "", "Reserved (classic-CAN transmit-address slots unused in FD mode)", DATA_FLAGS_READ_ONLY, 1.0, (uint32_t)(sizeof(uint32_t) * (NUM_CIDS - NUM_FDCIDS)));
 }
 
 static void PopulateMapDiagMsg(data_set_t data_set[DID_COUNT], uint32_t did)


aabogdanov (approved):

src/ISDataMappings.cpp
```diff
@@ -1998,6 +1998,13 @@ static void PopulateMapCanFdConfig(data_set_t data_set[DID_COUNT], uint32_t did)
...
+
+    // can_config_t is shared with DID_CAN_CONFIG, whose arrays are sized NUM_CIDS (27). CAN-FD
+    // only uses the first NUM_FDCIDS (10) slots of each array, so the mappings above cover just
+    // that subset. The DataMapper invariant requires the mapped fields to span the whole struct,
+    // so account for the FD-unused tail of both arrays as reserved regions.
+    mapper.AddMember2("can_fd_period_mult_reserved",      offsetof(can_config_t, can_period_mult)      + sizeof(uint16_t) * NUM_FDCIDS, DATA_TYPE_BINARY, "", "Reserved (classic-CAN period-mult slots unused in FD mode)",      DATA_FLAGS_READ_ONLY, 1.0, (uint32_t)(sizeof(uint16_t) * (NUM_CIDS - NUM_FDCIDS)));
+    mapper.AddMember2("can_fd_transmit_address_reserved", offsetof(can_config_t, can_transmit_address) + sizeof(uint32_t) * NUM_FDCIDS, DATA_TYPE_BINARY, "", "Reserved (classic-CAN transmit-address slots unused in FD mode)", DATA_FLAGS_READ_ONLY, 1.0, (uint32_t)(sizeof(uint32_t) * (NUM_CIDS - NUM_FDCIDS)));
```
Hephaestus-Maximus, line 2007: Nit: this codebase already has a convention for exactly this situation. `PopulateMapRtosInfo`'s `"T{i}.reserved"` filler field (ISDataMappings.cpp:1885) uses `DATA_FLAGS_HIDDEN` ("Do not print to screen") for a size-padding entry that shouldn't be surfaced to consumers. These two new `*_reserved` entries only carry `DATA_FLAGS_READ_ONLY`, which doesn't suppress display. Any code that dynamically enumerates all mapped fields for a DID via `ElementToInfo`/`NameToInfoMap` (e.g. EvalTool's generic field-driven config grids, or a CSV/JSON dump) will now show two new rows named `can_fd_period_mult_reserved` / `can_fd_transmit_address_reserved` for DID_CANFD_CONFIG — they'll render as empty strings (DATA_TYPE_BINARY has no string renderer), but they're a confusing/meaningless addition to any UI or export that lists all fields. Suggest adding `DATA_FLAGS_HIDDEN` alongside `DATA_FLAGS_READ_ONLY` on both new entries, matching the rtos_info_t precedent.

Hephaestus-Maximus (commented):
## Review summary

**Verified the fix works.** Built the SDK unit tests against this branch (`cmake .. -DCMAKE_BUILD_TYPE=Debug`, then `IS-SDK_unit-tests`) and confirmed:
- With this commit: `ISDataMappings.*` (5 tests) pass cleanly, no `DataMapper ERROR`, no assertion.
- With this commit reverted (parent `0db50c8b`, SN-8301's fix still in place): `ISDataMappings.StringToDataToString` — the *first* test that runs, unrelated to CAN config — immediately aborts with the exact reported error: `DataMapper ERROR [12can_config_t]: mapped size 66 != struct size 168 (diff 102 bytes)`. So this is a real, reproducible bug, and the existing test suite already exercises the `cISDataMappings` singleton init path that triggers it (worth noting: this means CI *should* already catch this class of bug, as long as it builds/runs tests without `NDEBUG`).

**Math checks out.** `NUM_CIDS=27`, `NUM_FDCIDS=10`. Reserved tail = `(27-10)*sizeof(uint16_t) + (27-10)*sizeof(uint32_t)` = 34 + 68 = 102 bytes. Existing FD mapping (10 period-mult + 10 transmit-address + can_setting + can_receive_address) = 66 bytes. 66 + 102 = 168 = `sizeof(can_config_t)`. `offsetof(...) + sizeof(elem)*NUM_FDCIDS` correctly starts each reserved region right after the last FD-used slot.

**Approach is idiomatic for this codebase.** `AddMember2` + `DATA_TYPE_BINARY` + explicit `typeSize` for a filler/reserved span is an established pattern already used for `gnss_raw_t::dataBuf` (ISDataMappings.cpp:1856).

**One suggestion (left inline):** the two new reserved entries use `DATA_FLAGS_READ_ONLY` only. `PopulateMapRtosInfo`'s `"T{i}.reserved"` filler (ISDataMappings.cpp:1885) — solving the identical padding problem — uses `DATA_FLAGS_HIDDEN` instead, which actually suppresses the field from display. Without it, anything that dynamically enumerates all fields for a DID (EvalTool's generic config grids, CSV/JSON dumps via `ElementToInfo`) will now show two new blank "reserved" rows for DID_CANFD_CONFIG.

**Soft suggestion (non-blocking):** this exact bug class — `can_config_t`'s DataMapper coverage silently falling short of `sizeof(struct)` — has now shipped twice for two different DIDs sharing the struct (SN-8301 for DID_CAN_CONFIG, this ticket for DID_CANFD_CONFIG). Coverage today is *incidental* (any test that touches `cISDataMappings` happens to trigger it). Given the repeat, a small explicit test naming `can_config_t`/CAN-FD coverage as its intent would fail with a much clearer signal than an assertion inside an unrelated string round-trip test.

No functional issues found — fix is correct and minimal.

Hephaestus-Maximus (approved):
## Final assessment

Weighing the `DATA_FLAGS_HIDDEN` nit against the fix: it's cosmetic-only. Worst case, EvalTool's generic field-driven grids or a CSV/JSON dump surface two extra blank "reserved" rows for DID_CANFD_CONFIG — no crash, no data corruption, no incorrect protocol behavior, and it doesn't touch the wire format or any value a firmware/host actually reads or writes. That's a nitpick, not a defect blocking merge.

The core change — fixing the DataMapper size-invariant abort that crashes the EvalTool debug build on startup — is correct, minimal, and independently verified (see prior comment: red/green against the existing test suite, arithmetic checked by hand).

**Approving.** The `DATA_FLAGS_HIDDEN` nit is worth picking up, either as a quick amendment here or a fast follow-up — leaving that to the author's judgment on timing.

PR #1224, opened 2026-07-16
SN-8307: Accept legacy GNSS1/GNSS2 firmware target names as CXD5610 aliases

kylemallory:
## What

The 3.1.0 firmware manifest renames the `GNSS1`/`GNSS2` update+validate steps and targets to `CXD1`/`CXD2`. This adds backward-compatible aliases in `ISFirmwareUpdater::cmd_SetTarget` so the legacy `GNSS1`/`GNSS2` target names still resolve to `TARGET_SONY_CXD5610__1`/`__2`.

This keeps older manifests and scripts working during the transition. The aliases are marked for removal 06/30/2027.

## Why

SDK-side counterpart to the manifest changes in the tied imx PR — without it, older manifests referencing `GNSS1`/`GNSS2` targets would fail to resolve after the rename.

## Related

- Tied imx PR (firmware manifest + semver 3.1.0): inertialsense/imx#1610
- Jira: SN-8307

### DIFF
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -950,6 +950,8 @@ void ISFirmwareUpdater::cmd_SetTarget(ISFwUpdaterCmd& cmd) {
         else if (targetName == "GPX1") setTarget(fwUpdate::TARGET_GPX1);
         else if (targetName == "CXD1") setTarget(fwUpdate::TARGET_SONY_CXD5610__1);
         else if (targetName == "CXD2") setTarget(fwUpdate::TARGET_SONY_CXD5610__2);
+        else if (targetName == "GNSS1") setTarget(fwUpdate::TARGET_SONY_CXD5610__1);        // TODO: Remove this 06/30/2027 - obsoleted by CXD1
+        else if (targetName == "GNSS2") setTarget(fwUpdate::TARGET_SONY_CXD5610__2);        // TODO: Remove this 06/30/2027 - obsoleted by CXD2
         else if (targetName == "SEPT") setTarget(fwUpdate::TARGET_SEPTENTRIO);
         else {
             handleCommandError(cmd, -1, "Invalid Target specified: %s  (Valid targets are: IMX5, IMX6, GPX1, CXD1, CXD2, SEPT)", targetName.c_str());


aabogdanov (approved):

src/ISFirmwareUpdater.cpp
```diff
@@ -950,6 +950,8 @@ void ISFirmwareUpdater::cmd_SetTarget(ISFwUpdaterCmd& cmd) {
         else if (targetName == "GPX1") setTarget(fwUpdate::TARGET_GPX1);
         else if (targetName == "CXD1") setTarget(fwUpdate::TARGET_SONY_CXD5610__1);
         else if (targetName == "CXD2") setTarget(fwUpdate::TARGET_SONY_CXD5610__2);
+        else if (targetName == "GNSS1") setTarget(fwUpdate::TARGET_SONY_CXD5610__1);        // TODO: Remove this 06/30/2027 - obsoleted by CXD1
+        else if (targetName == "GNSS2") setTarget(fwUpdate::TARGET_SONY_CXD5610__2);        // TODO: Remove this 06/30/2027 - obsoleted by CXD2
```
Hephaestus-Maximus, line 954: Suggestion, non-blocking: `cmd_SetTarget`'s string->target mapping has no unit test coverage at all (checked `tests/test_ISFirmwareUpdate.cpp` and `test_ISDFUFirmwareUpdater.cpp` — neither exercises this function). This exact mapping table has now had two related changes in two days (SN-8169 dropped `GNSS1`/`GNSS2`, this PR restores them as aliases) — a small parametrized test asserting each string resolves to its expected `target_t` (including these two aliases) would catch a future accidental drop before it reaches a manifest-compatibility incident again.

Hephaestus-Maximus (approved):
## Review summary — inertial-sense-sdk#1224

**Tier:** Trivial-mechanical diff (2 lines, additive, follows an existing pattern exactly) with real hardware-update blast radius, so I reasoned through the semantics and history rather than rubber-stamping the shape of the change.

### What I verified independently

- **This is the correct fix for the exact gap I flagged in imx#1610's review**: SDK commit `41985e7d` (SN-8169, merged 2026-07-13) dropped `GNSS1`/`GNSS2` as valid target strings from `cmd_SetTarget`, leaving only `IMX5, IMX6, GPX1, CXD1, CXD2, SEPT`. Any older manifest or script still using `GNSS1`/`GNSS2` (the pre-rename convention) would now hit `Invalid Target specified`. This PR restores those two names as aliases, unblocking backward compatibility during the transition.
- **Historical accuracy of the mapping**: `git log -S` on this file shows that before `41985e7d`, `"GNSS1"`/`"GNSS2"` mapped to exactly `TARGET_SONY_CXD5610__1`/`__2` — the same enum values this PR restores. This is a faithful restoration, not a new/guessed mapping.
- **No ambiguity with other GNSS hardware**: `FirmwareUpdate.h` also defines `TARGET_UBLOX_F9P` (a distinct u-blox GNSS target family), so "GNSS" isn't a name uniquely owned by the Sony CXD5610 — but grepped `cmd_SetTarget` and confirmed `TARGET_UBLOX_F9P` is never reachable via a string match in this function at all, so there's no collision risk between the legacy `GNSS1`/`GNSS2` alias and u-blox targets.
- **Single source of truth**: grepped the whole repo for `"CXD1"`/`"CXD2"` string literals — `ISFirmwareUpdater.cpp::cmd_SetTarget` is the only place that parses these target strings, so no other file needs the same alias added for consistency.
- **Downstream transparency**: `setTarget()` only receives the resolved `target_t` enum, never the original string — so nothing downstream can behave differently based on whether the input was the legacy or current name. The alias is fully transparent.
- **Convention match**: the `// TODO: Remove this 06/30/2027 - obsoleted by CXD1` comment style matches an existing precedent in the same file (`// TODO: This can be removed after 2.1.0 release`, line 604) — consistent with how this codebase already flags scheduled removals.
- **CI**: Linux + Windows build jobs pass (confirms it compiles cleanly). `ROS1`/`ROS2 Bridge Tests` show `skipping` — unrelated, repo-wide-disabled jobs (unrelated to this change, same as noted on the prior PR in this chain).
- Already has one human approval (@aabogdanov).

### Findings
- **Blocking:** none.
- **Suggestion (inline, non-blocking):** `cmd_SetTarget`'s entire string→target mapping has zero unit test coverage (checked `test_ISFirmwareUpdate.cpp`, `test_ISDFUFirmwareUpdater.cpp` — neither touches it). Given this exact table has now changed twice in two days (dropped, then aliased back), a small test asserting each accepted string resolves to its expected `target_t` would catch the next accidental drop before it becomes a fleet-compatibility incident again.

### Description grade: 5/5

Concise, accurate "what/why," correctly links the tied `imx#1610` PR and Jira ticket, and states the removal date for the deprecation. No edit needed.

### Verdict: **APPROVE**

Verified via: diff-read + git-history cross-check (confirming the mapping is a faithful restoration, not a guess) + whole-repo grep confirming this is the sole parse site and there's no collision with other GNSS hardware targets + CI build pass. No build/red-green harness needed — the change is a single, additive `else if` branch mirroring three sibling lines already proven correct by the passing CI build; the marginal value of a custom repro harness here was low relative to cost.

shefft (approved):

PR #1225, opened 2026-07-16
SN-8310 Improved accuracy of IMU integrals

aabogdanov:
Added Runge-Kutta-4 to Bortz to improve coning accuracy, sculling is the same but its accuracy is improved due to better coning.
Firmware size reduced by about 230 bytes due to changes in multi-to-single IMU functions.

---
### Review addendum (Hephaestus) — description completeness: 3/5

A few details a reader would need that aren't in the description above (see inline comments for full detail):

- The RK4 rewrite drops the `imuLast` parameter from `integrateDeltaThetaVelBortz` entirely — the coning integral's angular rate (`imu->pqr`) and the sculling term's acceleration (`imu->acc`) are now sourced *only* from the current raw sample, held constant across the RK4 sub-stages, rather than interpolating between the previous and current raw samples the way the old `Nsteps`-substep method did. Worth confirming this is deliberate (it lines up with "we will not be using oversampling like we used to do" from the Jira description) rather than an incidental side effect of the rewrite.
- A synthetic numerical comparison (old vs. new vs. a fine-grained ground-truth integration of the same Bortz ODE) under single-tone coning vibration showed the "better accuracy" claim isn't uniform across all sample-rate/vibration-frequency ratios — new was consistently better at 1 kHz sampling in my test, but old was consistently better at 4 kHz sampling. It'd help future readers to know what regime the Matlab simulation covered.
- The `multiToSingleImu`/`multiToSingleImuAxis` changes are a behavior-preserving refactor (consolidating duplicated gyro/accel loops into one parameterized pass) — confirmed by hand-tracing the mask arithmetic, not just a"changes" note.


src/filters.cpp
```diff
@@ -446,64 +439,67 @@ float deltaThetaDeltaVelTrapezoidal(pimu_t *output, imu_t *imu, imu_t *imuLast)
...
+    mag_theta2 = DOT_VEC3(theta);
+    mag_theta4 = mag_theta2 * mag_theta2;
+    Kw = Kw0 + mag_theta2 * Kw1 + mag_theta4 * Kw2; // + mag_theta4 * mag_theta2 * Kw3; <--- the last term is negligibly small
+    for (int i = 0; i < 3; i++) {
+        theta_dot[i] = omega[i] + 0.5f * thxwb[i] + Kw * thxthxwb[i];
+    }
+}
 
-void integrateDeltaThetaVelBortz(ixVector3 theta, ixVector3 vel, imui_t *imu, imui_t *imuLast, int Nsteps, float dt)
+static void integrateDeltaThetaVelBortz(ixVector3 theta, ixVector3 dvel, imui_t *imu, float dt)
```
Hephaestus-Maximus, line 377: Worth confirming this is deliberate: this function dropped the `imuLast` parameter entirely, so `imu->pqr` (the current raw sample, held constant across all 4 RK4 stages) is now the *only* angular-rate information used for the coning integral — the previous sample's rate is no longer read anywhere in this function. Similarly the sculling term below uses only `imu->acc` (current sample), not any blend with `imuLast->acc`.

I built a standalone harness replicating both the old (Nsteps=2, linearly-interpolated-between-samples) and new (RK4, endpoint-only) integrators against a fine-grained RK4 "ground truth" of the same Bortz ODE, driven by synthetic single-tone coning vibration at various frequency/sample-rate ratios and amplitudes. Results were genuinely mixed rather than a clean win for the new method: at 1 kHz sampling the new method was consistently more accurate than the old one across the frequencies I tried, but at 4 kHz sampling the *old* method came out ahead in every case I tried, and at large amplitude + aggressive frequency/sample-rate ratios (5 samples/cycle) the old method's result diverged outright (float overflow) while the new method stayed bounded (arguably itself a point in the new method's favor — the old Euler-style substepping is less numerically stable at that extreme, though that's a fairly aggressive vibration scenario). I can share the harness if useful, but the summary is: the "better accuracy" claim looks real for at least some regimes (matching your Matlab result), but I don't have enough confidence to say it holds universally across sample-rate/vibration-frequency ratios — it'd be good to know what regime your Matlab simulation covered, and whether it's representative of the sample rates this actually runs at in the field.

src/filters.cpp
```diff
@@ -101,17 +101,14 @@ void iir_filter_s16(iif_filter_t *f, short input[], float output[])
...
 {
-    int i;
-    float alpha;
+    float alpha = 1.0f;
     
-    if (sampleCount == 0)
-        alpha = 1.0f;
-    else
+    if (sampleCount > 0)
         alpha = 1.0f / (float)sampleCount;
```
Hephaestus-Maximus, line 107: Nitpick, very low priority: this reorders the logic from "alpha=1 if sampleCount==0 else 1/sampleCount" to "alpha=1; override if sampleCount>0" — behaviorally identical for sampleCount>=0, but differs for negative `sampleCount` (old would compute a negative alpha; new leaves it at 1.0). No callers of this function exist in this repo (grepped), and `sampleCount` is semantically a count so negative shouldn't occur in practice — flagging only because it's a silent behavior change for an out-of-domain input, not because I think it matters here.

src/filters.cpp
```diff
@@ -173,53 +169,52 @@ void recursive_moving_mean_var_filter(float *mean, float *var, float input, int
 
 void multiToSingleImu(imu_t *result, const imus_t *imus, const int numDevices)
```
Hephaestus-Maximus, line 169: Verified this refactor (and the equivalent one in `multiToSingleImuAxis` below) is behavior-preserving. Traced the mask arithmetic by hand: old computed `(GYR_X_OK << (d*BITSIZE)) << a`, new computes `(GYR_X_OK << iaxis) << (idev*BITSIZE)` — same total shift amount since shifts add commutatively and `d`/`a` map 1:1 to `idev`/`iaxis`. Status bit set/clear logic and the zero-fallback for axes with no valid device data are both preserved. Matches the described ~230-byte firmware size reduction from consolidating the duplicated gyro/accel loops into one parameterized pass — good simplification.

Hephaestus-Maximus (commented):
## Review summary — inertial-sense-sdk#1225

**Tier: Substantial/numerically-sensitive.** This changes the core coning/sculling integration algorithm (`filters.cpp`) that feeds attitude/velocity preintegration for the whole nav stack, plus a mechanical refactor of the multi-IMU averaging functions. Full diff read in context, an isolated compile of the changed file, a hand-traced equivalence check of the averaging refactor, and a standalone numerical comparison of the old vs. new coning integrator against a fine-grained ground truth.

### What I verified

- **`multiToSingleImu`/`multiToSingleImuAxis` refactor — verified correct, no regression.** Hand-traced the mask/shift arithmetic: old computed `(GYR_X_OK << (d*BITSIZE)) << a`, new computes `(GYR_X_OK << iaxis) << (idev*BITSIZE)` — identical total shift since the two loop variables map 1:1 and shifts add commutatively. Status-bit set/clear and the zero-fallback for axes with no valid device are both preserved. This matches the claimed ~230-byte firmware size reduction from consolidating the duplicated gyro/accel code into one parameterized loop.
- **Compiles cleanly.** Couldn't build the full SDK target (unrelated missing system dependency, `libusb.h`, in files this PR doesn't touch), but compiled `filters.cpp` in isolation with `-Wall -Wextra`: zero errors, and the one warning present is pre-existing and in an unrelated function this diff doesn't touch.
- **No downstream breakage from the API signature change.** `deltaThetaDeltaVelBortz` dropped its `Nsteps` parameter (a public declaration in `filters.h`). Searched GitHub code search across the `inertialsense` org for both `deltaThetaDeltaVelBortz` and `CON_SCUL_INT_STEPS` — no other repo calls this function or references the removed constant.
- **The coning/sculling numerics — mixed results, not a clean confirmation either way.** Extracted `bortz()`/`integrateDeltaThetaVelBortz` (new) and the old `Nsteps`-substep equivalent into a standalone harness, and compared both against a fine-grained (50 sub-step) RK4 integration of the *same* Bortz ODE as ground truth, driven by synthetic single-tone coning vibration (varying frequency, amplitude, and sample rate). Findings (see inline comment for detail):
  - At 1 kHz sampling, the new (RK4, endpoint-only) method was consistently more accurate than the old (2-substep, interpolated) method across the frequencies I tried — this supports the PR's claim.
  - At 4 kHz sampling, the *old* method was consistently more accurate in every case I tried — this doesn't support the claim.
  - At an aggressive amplitude/frequency combination (5 samples/cycle, 0.15 rad amplitude), the old method's result diverged (float overflow) while the new method stayed bounded — arguably a stability point in the new method's favor, though it's a fairly extreme scenario.
  - I don't know what sample-rate/vibration-frequency regime the author's Matlab simulation covered, so I can't tell whether my test's "OLD wins at 4kHz" result contradicts their validated scenario or is simply outside what they tested (this device likely runs at a specific, known sample rate — if that's within the regime where NEW wins, the mixed result elsewhere may not matter in practice).
- **Confirmed a concrete, checkable fact**, not a numerics judgment call: `integrateDeltaThetaVelBortz`'s new signature drops `imuLast` entirely — both the coning (`imu->pqr`) and sculling (`imu->acc`) computations now use only the current raw sample, held constant across the RK4 stages, rather than interpolating with the previous sample the way the old method did. This is a real, describable change in what information is used, not explicitly called out in the description (which frames the change as being about the integration method, not the sampling).
- No existing or added test coverage for `filters.cpp`'s integration functions (pre-existing gap, not introduced by this PR) — given the numerical subtlety here, a unit test asserting integration accuracy against a known analytic scenario (e.g. constant angular rate, or a specified coning-vibration profile) would make the "improved accuracy" claim independently verifiable in CI going forward, not just in a one-off Matlab simulation.
- Two low-priority nitpicks (inline): a now-fully-commented-out dead if-block, and a subtle behavior change for negative `sampleCount` input to `running_mean_filter` (no callers found in this repo, so unlikely to matter).

### Description grade: 3/5 → appended addendum

The description correctly states what changed and why, but omits the `imuLast`-dropped detail and doesn't give a sense of what regime the Matlab validation covered — appended a short addendum with those specifics rather than rewriting the author's text.

### Verdict: **COMMENT (no approve/reject)**

I can't reach confident enough ground to approve or block on the core numerical claim — my synthetic test shows real, sample-rate-dependent variance in which method is actually more accurate, so "improved coning accuracy" is true in some regimes and not others by my testing, and I don't have the author's Matlab setup or real hardware log data to know which regime matters for this device's actual operating sample rate. The mechanical refactor (multi-IMU averaging) is verified correct and I'd approve that part on its own. Given the safety-relevant nature of attitude/velocity integration, I'd want the author (or Kyle) to either confirm the sample-rate regime this runs at is one where the new method wins, or share the Matlab validation scope, before merging — this is a domain-numerics call that needs the person who ran the simulation, not something I should guess a verdict on.

src/filters.cpp
aabogdanov: It is deliberate. The 1 Khz samples are already _average_  IMU observations over 1 ms intervals (the IMU internally integrates the measurements and outputs their average).

aabogdanov:
@Hephaestus-Maximus how can I share my matlab script with you?

src/filters.cpp
aabogdanov: How can I share my Matlab code with you?

shefft (approved):

Hephaestus-Maximus (approved):
## Follow-up: correcting my earlier COMMENT verdict — upgrading to APPROVE

Alex shared his `coning_demo.m` Matlab validation script directly. Rebuilt my comparison in C++ against his exact methodology, and it revealed a real flaw in my earlier synthetic test — one that fully explains the "mixed results" I reported, and reverses my conclusion.

### What I got wrong the first time

My original harness had two methodological gaps relative to Alex's validated approach:

1. **No epoch reset.** I let `theta` accumulate continuously for a full 0.5s (up to ~500 raw samples) without ever converting it to a quaternion increment and resetting to zero. Alex's script resets the rotation vector to zero every `Nepoch=14` samples (matching how `pimu_t.theta` is actually consumed in the real pipeline). The Bortz small-angle series (`Kw0 + theta^2*Kw1 + theta^4*Kw2`) is only valid for small `theta` — letting it grow to multiple radians (which it did in several of my original test cases) pushes both methods far outside their intended small-angle operating envelope, and whatever differences show up there aren't representative of real usage.
2. **Wrong ground truth.** I compared against a fine-grained integration of the *same Bortz ODE* as "truth," which only measures discretization error relative to Bortz's own approximation — not against true rotation kinematics. Alex's script integrates the actual quaternion kinematics (`qdot = 0.5*Xi(q)*w`) via RK4 at 10x the sample rate using the exact analytical `w(t)`, which is the right way to isolate real error.

I also confirmed his `W` samples are the **time-averaged** true rate over each interval (`wx_avg(k) = (1/dt)*integral(A*sin(Omega*t), t_k, t_k+dt)`), not a raw instantaneous point sample — this answers the "is dropping `imuLast` intentional" question from my first review comment on `integrateDeltaThetaVelBortz`: if `imu->pqr` in the firmware represents a per-interval average (as it would from a delta-angle-based gyro output), then using only the current interval's own average, held constant across the RK4 sub-stages, is the *correct* model — there's no missing information from not blending with the previous interval's average.

### Corrected result

Reimplementing his exact methodology in C++ (epoch-reset rotation vector, quaternion ground truth, time-averaged `W`), with the OLD (pre-#1225, `Nsteps=2` interpolated) integrator substituted in alongside his validated NEW (RK4) one:

At his exact validated point (`fs=1kHz, Omega=20 rad/s, eta=15deg, 60s run, Nepoch=14`):

```
                                     Roll        Pitch       Yaw
OLD Bortz (Nsteps=2, interpolated)  0.093 deg   0.091 deg   0.128 deg
NEW Bortz (RK4, PR #1225)           0.017 deg   0.017 deg   0.092 deg
```

NEW wins on all three axes (~5x better roll/pitch, ~1.4x better yaw). I then swept sample rate (500 Hz – 4 kHz), cone frequency (10–80 rad/s), and cone half-angle (5–30 deg) around that point — **NEW was more accurate than OLD in every single scenario I tried**, not just the validated one. This is a much more decisive and confidence-inspiring result than my first (flawed) test suggested, and it directly confirms the PR's "improved coning accuracy" claim under realistic operating conditions.

### Updated verdict: **APPROVE**

Withdrawing my earlier "mixed results, can't confirm" position — that was an artifact of my test's methodology, not a real property of the algorithm. The core numerical claim now checks out under a methodologically sound, epoch-aware, quaternion-ground-truth comparison, robust across a reasonably wide parameter sweep. The `multiToSingleImu` refactor was already independently verified correct in my first pass. Thanks to Alex for sharing the validation script — happy to share the corrected C++ port if useful for future regression checks.

src/filters.cpp
```diff
@@ -35,11 +35,11 @@ void init_iir_filter(iif_filter_t *f)
     int alpha;
     float TsFc;
 
-    if (f->opt.n_channels > MAX_NUMBER_IIR_CHANNELS)
-    {
+//    if (f->opt.n_channels > MAX_NUMBER_IIR_CHANNELS)
```
kylemallory: Do we want to keep this code?  If its dead, let's remove it.  If its not, comment why we want to keep it (and for how long).

kylemallory (dismissed):

src/filters.cpp
aabogdanov: The code is a safety check. Not sure why it was disabled. I re-enabled it.

Hephaestus-Maximus (approved):
## Re-review after cleanup commits — still APPROVE

Re-requested review, so I re-checked what changed since my last APPROVE (several "code simplified/reduced" / "cleanup" / "bug fix" commits, tip now at `e2ad8ae`).

### Open items from before — now resolved

- **`imuLast` dropped from the coning integral** (my original question): confirmed by Alex directly — "The 1 kHz samples are already *average* IMU observations over 1 ms intervals (the IMU internally integrates the measurements and outputs their average)." This matches what I verified independently against his Matlab script and my corrected numerical harness.
- **Dead/disabled code question** (Kyle's comment): Alex identified the disabled `init_iir_filter` safety check as unintentionally disabled and re-enabled it (`exit(1)` guard against `n_channels > MAX_NUMBER_IIR_CHANNELS`). Unrelated to the coning/sculling work itself, but a reasonable, small defensive fix — confirmed it compiles cleanly in the current tip.

### What changed in this round

- `integratePimu` and the `deltaThetaDeltaVelBortz` wrapper were merged/inlined and the dead code around them reordered a few times (hence the repeated "simplified/reduced" / "bug fix" commits) — I diffed each intermediate step and the net effect is a pure refactor: `deltaThetaDeltaVelBortz()` (which just computed `dt` and called the static `integrateDeltaThetaVelBortz` helper) was removed, and `integratePimu` now computes `dt` and calls that helper directly. No behavior change — confirmed by tracing the arithmetic through both versions.
- The core numerics (`bortz()` and `integrateDeltaThetaVelBortz()` — the RK4 coning/sculling functions I validated extensively against Alex's `coning_demo.m` methodology) are **byte-for-byte unchanged** at the current tip. Re-diffed them directly to confirm.
- `multiToSingleImuExc` now uses `inv_count_upto10()` instead of a raw division, matching the pattern already used elsewhere in this file — harmless, consistent cleanup.
- Compiled `filters.cpp` in isolation at the current tip (`e2ad8ae`): clean, zero errors, same one pre-existing unrelated warning as before.

### One very minor nitpick (not blocking)

The disabled `integrateDeltaThetaVelRoscoe()` (inside `#if 0`/`#endif`, dead legacy code) now has a missing comma and a stray semicolon in its parameter list (`ixVector3 delta_veloc_last` / `float dt;` / `)`). Completely inert since it's preprocessed out, but would fail to compile if anyone ever flips `#if 0` to `#if 1` without noticing. Not worth blocking on.

### Verdict: **APPROVE** (re-confirmed)

No functional or numerical regressions in this round of cleanup — the substantive work (RK4 Bortz coning/sculling, validated against your Matlab script) is untouched, and the refactor around it is behavior-preserving. Good to go from my side.

PR #1227, opened 2026-07-19
ROS2: fix compile/link errors and port to current ISDevice-based SDK API

jbergstrom88:
This gets ROS/ros2 actually compiling, linking, and streaming data correctly
against the SDK's current API (as of develop, ~commit ccb95f66). It was
previously broken by several categories of issues -- see commit-by-commit
breakdown below. Tested end-to-end against a RUG-3 (IMX-5) on ROS 2 Jazzy /
Ubuntu 24.04, streaming INS2/INS4/INL2 states, dual-GNSS pos/vel/sat,
magnetometer, barometer, PIMU, raw IMU, and diagnostics successfully.

## Commits

**Relative path fixes** -- CMakeLists.txt assumed ROS/ros2 sits one level
deeper than it does relative to the sibling inertial-sense-sdk checkout
(comment in the file even calls this out: "stays in .external file to
reference submodule").

**SET_CALLBACK implementation** -- this macro was entirely commented out,
with a TODO citing ISDevice::registerDataHandler(), which doesn't exist
in the SDK. Implemented via ISDevice::registerIsbDataHandler() chaining:
installs our own handler on top of ISDevice's existing processIsbMsgs,
calling it first (preserving internal sysParams/flash-config sync) before
dispatching to a per-DID callback table. Also fixes several related API
drift issues found in the process:
- DeviceInfo()/BroadcastBinaryData() moved from InertialSense to ISDevice
- DeviceByPortName()/DeviceByPort() are declared but have no implementation
  anywhere (linker error) -- switched to getDevices(), which is inline
- getPorts() return type changed from vector to set
- RTKRel msg constants drifted (GNSS_STATUS_* in .cpp vs GPS_STATUS_* in
  the generated .msg)
- gps2 NavSatFix had its own publisher created but was never actually
  published to anywhere (hardcoded to gps1's publisher) -- now publishes
  independently using GNSS2's own data

**RTK NTRIP client/server stubs** -- OpenConnectionToServer/
ClientServerByteCount/CreateHost/CloseServerConnection are called but have
no implementation anywhere in the current SDK. Stubbed with clear
TODO(SDK) markers and safe failure behavior (logs an error, returns
false/0) rather than silently pretending to succeed. RTK NTRIP client/
server functionality is disabled until a real implementation exists.

**argv parsing fix** -- main() used ROS1-style argc>1/argv[1] parsing,
which picks up the literal string "--ros-args" as a YAML path whenever
launched with parameter overrides (e.g. `--ros-args -p port:=...`),
silently discarding them. Fixed with rclcpp::remove_ros_arguments().

## Known gaps (not fixed here, flagged with TODO(SDK) comments)

- RTK NTRIP client/server has no underlying SDK implementation
  (OpenConnectionToServer/CreateHost/ClientServerByteCount/
  CloseServerConnection)
- ISDevice::registerDataHandler() referenced in the original TODO comment
  doesn't exist -- SET_CALLBACK's replacement here is a workaround
  (handler chaining), not that intended API

Runs successfully w/ 
$$ ros2 run inertial_sense_ros2 inertial_sense_ros2_node ~/my_rug3_params.yaml --ros-args -p port:=/dev/ttyACM0 -p baudrate:=921600

### DIFF
--- a/ROS/ros2/src/inertial_sense_ros2.cpp
+++ b/ROS/ros2/src/inertial_sense_ros2.cpp
@@ -21,6 +21,7 @@
 #include <chrono>
 #include <stddef.h>
 #include <unistd.h>
+#include <set>
 //#include <ISPose.h>
 //#include <duration.hpp>
 #include "ISEarth.h"
@@ -30,6 +31,9 @@
 //#include "ISMatrix.h"
 
 #include "ParamHelper.h"
+
+// Single active instance -- see comment above s_instance_'s declaration in inertial_sense_ros.h
+InertialSenseROS* InertialSenseROS::s_instance_ = nullptr;
 
 #define STREAMING_CHECK(streaming, DID)      if (!streaming){ streaming = true; rclcpp::Logger logger_IS_resp_rec = rclcpp::get_logger("IS_response_received"); logger_IS_resp_rec.set_level(rclcpp::Logger::Level::Debug); RCLCPP_DEBUG(logger_IS_resp_rec,"InertialSenseROS: %s response received", cISDataMappings::DataName(DID)); }
 //#define STREAMING_CHECK(streaming, DID)      if (!streaming){ streaming = true; RCLCPP_DEBUG("InertialSenseROS: %s response received", cISDataMappings::DataName(DID)); }
@@ -49,6 +53,8 @@ void odometryIdentity(nav_msgs::msg::Odometry& msg_odom) {
 
 InertialSenseROS::InertialSenseROS(YAML::Node paramNode, bool configFlashParameters): nh_(rclcpp::Node::make_shared("nh_"))
 {
+    s_instance_ = this;
+
     // Should always be enabled by default
     rs_.did_ins1.enabled = true;
     rs_.did_ins1.topic = "did_ins1";
@@ -83,7 +89,11 @@ void InertialSenseROS::initialize(bool configFlashParameters)
 void InertialSenseROS::terminate()
 {
     IS_.Close();
-    IS_.CloseServerConnection();
+    // TODO(SDK): InertialSense::CloseServerConnection() does not exist in the current SDK (see
+    // note in connect_rtk_client() above -- the whole RTK NTRIP client/server connection
+    // subsystem is unimplemented right now). Since OpenConnectionToServer()/CreateHost() are
+    // themselves stubbed to no-ops, no server connection is ever actually opened, so there is
+    // nothing here that needs closing.
     sdk_connected_ = false;
 
     // ROS equivalent to shutdown advertisers, etc.
@@ -605,16 +615,16 @@ void InertialSenseROS::configure_data_streams(bool firstrun) // if firstrun is t
             uint16_t gnssSatSigConst = flashCfg.gnssSatSigConst;
 
             if (gnssSatSigConst & GNSS_SAT_SIG_CONST_GPS) {
-                msg_NavSatFix.status.service |= NavSatFixService::SERVICE_GPS;
+                msg_NavSatFix2.status.service |= NavSatFixService::SERVICE_GPS;
             }
             if (gnssSatSigConst & GNSS_SAT_SIG_CONST_GLO) {
-                msg_NavSatFix.status.service |= NavSatFixService::SERVICE_GLONASS;
+                msg_NavSatFix2.status.service |= NavSatFixService::SERVICE_GLONASS;
             }
             if (gnssSatSigConst & GNSS_SAT_SIG_CONST_BDS) {
-                msg_NavSatFix.status.service |= NavSatFixService::SERVICE_COMPASS; // includes BeiDou.
+                msg_NavSatFix2.status.service |= NavSatFixService::SERVICE_COMPASS; // includes BeiDou.
             }
             if (gnssSatSigConst & GNSS_SAT_SIG_CONST_GAL) {
-                msg_NavSatFix.status.service |= NavSatFixService::SERVICE_GALILEO;
+                msg_NavSatFix2.status.service |= NavSatFixService::SERVICE_GALILEO;
             }
         }
         NavSatFixConfigured = true;
@@ -693,7 +703,28 @@ bool InertialSenseROS::connect(float timeout)
             RCLCPP_ERROR(rclcpp::get_logger("open_port_error"),"InertialSenseROS: Unable to open serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
             sleep(1); // is this a good idea?
         } else {
-            RCLCPP_INFO(rclcpp::get_logger("serial_port_connected_info"),"InertialSenseROS: Connected to IMX SN%d on \"%s\", at %d baud", IS_.DeviceInfo().serialNumber, cur_port.c_str(), baudrate_);
+            // Device discovery/validation happens asynchronously, driven by IS_.Update(); give it
+            // a brief bounded window to find and validate the ISDevice on the port we just opened.
+            //
+            // NOTE: InertialSense::DeviceByPortName() and DeviceByPort() are declared in
+            // InertialSense.h but have no implementation anywhere in the SDK (link error:
+            // undefined reference). Using IS_.getDevices() instead, which is defined inline in
+            // the header (returns deviceManager directly) and is guaranteed to link. Since
+            // IS_.Open() only ever opens a single port/device in this node's design, taking the
+            // first (only) entry once the list is non-empty is safe and correct here.
+            for (int attempt = 0; !device_ && attempt < 100; attempt++) {
+                IS_.Update();
+                auto& devices = IS_.getDevices();
+                if (!devices.empty())
+                    device_ = devices.front();
+                if (!device_) usleep(20000); // 20ms
+            }
+            if (device_) {
+                registerIsbDispatch();
+                RCLCPP_INFO(rclcpp::get_logger("serial_port_connected_info"),"InertialSenseROS: Connected to IMX SN%d on \"%s\", at %d baud", device_->DeviceInfo().serialNumber, cur_port.c_str(), baudrate_);
+            } else {
+                RCLCPP_WARN(rclcpp::get_logger("serial_port_connected_info"),"InertialSenseROS: Connected on \"%s\", at %d baud, but no ISDevice was found/validated on that port within the timeout.", cur_port.c_str(), baudrate_);
+            }
             port_ = cur_port;
             break;
         }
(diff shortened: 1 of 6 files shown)


ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -1604,6 +1690,45 @@ void InertialSenseROS::GPS_pos_callback(eDataIDs DID, const gnss_pos_t *const ms
...
+        // the original) but, in the upstream code, was never actually published to -- only
+        // rs_.gps1_navsatfix.pub_nsf was ever used (see the primaryGpsDid block below), so
+        // gps2's NavSatFix would never publish regardless of its own "enable" setting. This
+        // mirrors the same logic gps1 uses, independently, using GNSS2's own data and publisher.
+        if (rs_.gps2_navsatfix.enabled && msg->status & GNSS_STATUS_FIX_MASK)
+        {
+            msg_NavSatFix2.header.stamp = ros_time_from_week_and_tow(msg->week, msg->timeOfWeekMs / 1.0e3);
+            msg_NavSatFix2.header.frame_id = frame_id_;
+            msg_NavSatFix2.status.status = -1;                            // Assume no Fix
+            if (msg->status & GNSS_STATUS_FIX_MASK >= GNSS_STATUS_FIX_2D) // Check for fix and set
```
Hephaestus-Maximus, line 1704: **Blocking — confirmed logic bug, not just style.** `&` has lower precedence than `>=` in C++, so this parses as `msg->status & (GNSS_STATUS_FIX_MASK >= GNSS_STATUS_FIX_2D)`, not `(msg->status & GNSS_STATUS_FIX_MASK) >= GNSS_STATUS_FIX_2D`.

Checked the actual constants in `src/data_sets.h`: `GNSS_STATUS_FIX_MASK = 0x00001F00`, `GNSS_STATUS_FIX_2D = 0x00000200`. So `GNSS_STATUS_FIX_MASK >= GNSS_STATUS_FIX_2D` is always true at compile time (7936 >= 512), collapsing the whole condition to `msg->status & 1` — testing bit 0 of the raw status word, which has nothing to do with fix quality. The same pattern two lines below (`>= GNSS_STATUS_FIX_RTK_SINGLE`, 0x00000A00) collapses the same way.

This exact pattern already exists in the pre-existing gps1 NavSatFix code a bit further down in this file (not part of this diff, so not blocking on its own, but since you're already touching this function it'd be worth fixing both while you're in here) — worth confirming this wasn't copied from there on purpose. Needs parens: `(msg->status & GNSS_STATUS_FIX_MASK) >= GNSS_STATUS_FIX_2D`.

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -1604,6 +1690,45 @@ void InertialSenseROS::GPS_pos_callback(eDataIDs DID, const gnss_pos_t *const ms
...
+            {
+                msg_NavSatFix2.status.status = NavSatFixStatusFixType::STATUS_FIX;
+            }
+
+            if (msg->status & GNSS_STATUS_FIX_SBAS) // Check for SBAS only fix
+            {
+                msg_NavSatFix2.status.status = NavSatFixStatusFixType::STATUS_SBAS_FIX;
+            }
+
+            if (msg->status & GNSS_STATUS_FIX_MASK >= GNSS_STATUS_FIX_RTK_SINGLE) // Check for any RTK fix
```
Hephaestus-Maximus, line 1714: Same operator-precedence bug as the fix-check above (`&` binds looser than `>=`): this collapses to `msg->status & 1` for the same reason (`GNSS_STATUS_FIX_MASK=0x1F00 >= GNSS_STATUS_FIX_RTK_SINGLE=0xA00` is always true). Needs `(msg->status & GNSS_STATUS_FIX_MASK) >= GNSS_STATUS_FIX_RTK_SINGLE`.

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -693,7 +703,28 @@ bool InertialSenseROS::connect(float timeout)
...
+            // Device discovery/validation happens asynchronously, driven by IS_.Update(); give it
+            // a brief bounded window to find and validate the ISDevice on the port we just opened.
+            //
+            // NOTE: InertialSense::DeviceByPortName() and DeviceByPort() are declared in
+            // InertialSense.h but have no implementation anywhere in the SDK (link error:
+            // undefined reference). Using IS_.getDevices() instead, which is defined inline in
+            // the header (returns deviceManager directly) and is guaranteed to link. Since
+            // IS_.Open() only ever opens a single port/device in this node's design, taking the
+            // first (only) entry once the list is non-empty is safe and correct here.
+            for (int attempt = 0; !device_ && attempt < 100; attempt++) {
```
Hephaestus-Maximus, line 715: Worth double-checking: `device_` is never reset to null anywhere in this diff, and this discovery loop is guarded by `!device_` — so if `connect()` is called again while `device_` is already set from a prior successful connection, this loop is skipped entirely (falls straight through to reusing the old `device_`).

That path is reachable today: `initializeIS()` calls `connect()` twice when `factory_reset_` is enabled -- once before issuing `SYS_CMD_MANF_FACTORY_RESET` (which reboots the device), and once again right after. The second call would skip re-discovery and reuse the pre-reset `device_`/port registration rather than validating a fresh one, even though `IS_.Open()` reopens the port and the hardware just rebooted.

I can't verify from here whether the underlying `ISDevice` object/shared_ptr the manager hands back after a reboot is the same instance (in which case this is fine) or a new one (in which case `device_` would be stale) -- that needs someone who knows the `DeviceManager`/reconnect internals, or a factory-reset test on real hardware. If it's actually fine, a short comment here explaining why would help the next reader; if not, resetting `device_ = nullptr` before this loop (or specifically around the factory-reset path) would close the gap.

Hephaestus-Maximus (changes requested):
## Review summary — inertial-sense-sdk#1227

**Tier: Substantial.** This is a real, well-scoped fix for the ROS2 bridge's compile/link errors after an SDK API refactor (InertialSense → ISDevice), plus a genuine hardcoded-gps2-publisher bug fix and an argv-parsing fix. Read the full diff in context, cross-referenced every API-drift claim in the description directly against the current SDK source, and traced the new `device_`/callback-dispatch machinery through its actual call sites.

### What I verified independently

I don't have a ROS2 toolchain in this environment (and CI's `ROS2 Bridge Tests` job is unconditionally disabled repo-wide, unrelated to this PR), so I couldn't build/run the ROS2 package myself. Instead I checked every specific API-drift claim in the PR description against the actual SDK source in this branch, and all of them check out:

- **`DeviceByPort`/`DeviceByPortName` declared but unimplemented** — confirmed. Neither has a definition anywhere in the repo, yet `InertialSense.cpp` itself calls `DeviceByPort()` internally (in `SetSysCmd`/`BroadcastBinaryData`). The reason the main SDK build doesn't already fail on this: `libInertialSenseSDK.a` is a static library, and the linker only needs to resolve symbols for functions actually pulled in by something the final binary calls — since nothing in cltool/tests currently calls those specific overloads, the dangling reference stays latent until something (like the ROS2 bridge, in its original form) actually calls `DeviceByPort`/`DeviceByPortName` directly. `getDevices()` (used in the fix) is defined inline in the header, so it's always safe to link against.
- **3-arg `BroadcastBinaryData(id, period, callback)` compiled out via `#if 0`** — confirmed, at `InertialSense.cpp:751`.
- **`getPorts()` returns `std::set` not `std::vector`** — confirmed in `InertialSense.h`.
- **`registerIsbDataHandler`'s "returns the previous handler, caller must chain it" contract** — confirmed via the doc comment on `ISDevice::registerIsbDataHandler`, and the PR's `isbDispatchThunk`/`prev_isb_handler_` chaining correctly implements that contract.
- The gps2-NavSatFix-never-actually-published claim is plausible from reading the surrounding code (the publisher existed but nothing in the pre-existing code path wrote to it independently of gps1) — didn't chase the git history to find exactly when this drifted, but the fix itself (mirroring gps1's logic against gps2's own message/publisher) is structurally sound.

### Findings

**Blocking (2, inline):**
1. **Confirmed logic bug, not style**: `msg->status & GNSS_STATUS_FIX_MASK >= GNSS_STATUS_FIX_2D` (and the RTK-single equivalent two lines below) — `&` has lower precedence than `>=` in C++, so this parses as `msg->status & (GNSS_STATUS_FIX_MASK >= GNSS_STATUS_FIX_2D)`. I checked the actual constants: `GNSS_STATUS_FIX_MASK = 0x1F00`, `GNSS_STATUS_FIX_2D = 0x200`, `GNSS_STATUS_FIX_RTK_SINGLE = 0xA00` — the `>=` comparison is always true at compile time for both, so both conditions collapse to `msg->status & 1`, testing an unrelated bit rather than actual fix quality. This exact pattern already exists in the pre-existing (untouched-by-this-diff) gps1 code a bit further down — not blocking on its own since it's out of scope, but worth fixing both while this function is already open.
2. **`device_` reconnect gap**: never reset to null anywhere, and the device-discovery loop is skipped if it's already set. `initializeIS()`'s factory-reset path calls `connect()` twice in a row (once pre-reset, once post-reboot) — the second call would silently reuse the pre-reset `device_` rather than re-validating. I can't verify from source alone whether the `DeviceManager` hands back the same object across a reboot (in which case this is fine) or a new one (in which case it's stale) — flagged as a question for someone who knows those internals, or worth a factory-reset hardware test.

**Suggestion:**
- No test coverage for the new gps2 NavSatFix logic or the SET_CALLBACK dispatch chain — consistent with the broader (pre-existing, repo-wide) gap that `ROS2 Bridge Tests` is disabled in CI, not something to hold this PR to alone.

### Description grade: 5/5

Excellent — accurate, specific, commit-by-commit breakdown, an explicit "Known gaps" section with `TODO(SDK)` markers for what's genuinely unimplemented upstream, and a concrete hardware validation statement. No edit needed.

### Verdict: **REQUEST_CHANGES**

The operator-precedence bug in finding #1 is a confirmed, provable logic error in code this PR adds (not a style nit) that would make gps2's NavSatFix fix-quality reporting effectively meaningless in production — worth fixing before merge given how easy the fix is. Everything else in this PR (the API-port work, the CMake path fix, the argv fix, the RTK stubs) is solid, well-documented, and independently verified against the SDK source. Once #1 is fixed (and #2 is answered or addressed), this looks very mergeable.

tmcd249 (approved):

ROS/ros2/src/inertial_sense_ros2.cpp
shefft: Maybe use parentheses to make intent explicit, even if order of ops let's this fall out as intended

ROS/ros2/src/inertial_sense_ros2.cpp
shefft: same, use parentheses

shefft (changes requested):
I like Heph's suggestions for those conditional statements

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -83,7 +89,11 @@ void InertialSenseROS::initialize(bool configFlashParameters)
 void InertialSenseROS::terminate()
 {
     IS_.Close();
-    IS_.CloseServerConnection();
+    // TODO(SDK): InertialSense::CloseServerConnection() does not exist in the current SDK (see
```
kylemallory, line 92: We should discuss this.  RTK corrections has changed, and is its own class/framework outside of the InertialSense class.  We should probably look to remove all "does not exist..." comments, since its not terribly meaningful to note what doesn't exist.  If anything it should note "TODO: Implement RTK corrections using the new corrections frameworks."

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -693,7 +703,28 @@ bool InertialSenseROS::connect(float timeout)
             RCLCPP_ERROR(rclcpp::get_logger("open_port_error"),"InertialSenseROS: Unable to open serial port \"%s\", at %d baud", cur_port.c_str(), baudrate_);
             sleep(1); // is this a good idea?
         } else {
-            RCLCPP_INFO(rclcpp::get_logger("serial_port_connected_info"),"InertialSenseROS: Connected to IMX SN%d on \"%s\", at %d baud", IS_.DeviceInfo().serialNumber, cur_port.c_str(), baudrate_);
+            // Device discovery/validation happens asynchronously, driven by IS_.Update(); give it
+            // a brief bounded window to find and validate the ISDevice on the port we just opened.
+            //
+            // NOTE: InertialSense::DeviceByPortName() and DeviceByPort() are declared in
```
kylemallory, line 709: There is probably room for a discussion regarding how the SDK has changed (significantly) since the ROS2 node was implemented.  Understanding that this is working - this points to the above comment about referencing things which don't exist anymore.  Rather than calling out the missing components, this should be rephrased to address the intended future behavior instead.

Ultimately, the majority of the ROS nodes are built on the InertialSense class, which is effectively being deprecated in favor of the PortManager/DeviceManager and ISDevice.  We should be outline this shift as the forward-looking objective.

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -739,13 +813,13 @@ bool InertialSenseROS::firmware_compatiblity_check()
...
   //         IS_SDK_VERSION_MINOR,
   //         IS_SDK_VERSION_REVIS,
-  //         IS_.DeviceInfo().protocolVer[0],
-  //         IS_.DeviceInfo().protocolVer[1],
-  //         IS_.DeviceInfo().protocolVer[2],
-  //         IS_.DeviceInfo().protocolVer[3],
-  //         IS_.DeviceInfo().firmwareVer[0],
-  //         IS_.DeviceInfo().firmwareVer[1],
-  //         IS_.DeviceInfo().firmwareVer[2]
+  //         device_->DeviceInfo().protocolVer[0],
```
kylemallory, line 816: All of this appears to be dead/deprecated code?? If so, just remove it.  It does not one any good by keeping around code that references an "older way of doing things".

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -876,7 +950,14 @@ void InertialSenseROS::connect_rtk_client(RtkRoverCorrectionProvider_Ntrip& conf
     int RTK_connection_attempt_count = 0;
     while (++RTK_connection_attempt_count < config.connection_attempt_limit_)
     {
-        config.connected_ = IS_.OpenConnectionToServer(RTK_connection);
+        // TODO(SDK): InertialSense::OpenConnectionToServer() does not exist in the current SDK
```
kylemallory, line 953: See my other comment at the start regarding the CorrectionService. And actually, this HAS been implemented in 3.0.0 (RTCM3) and 3.1.0 (NTRIP)

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -909,7 +990,11 @@ void InertialSenseROS::rtk_connectivity_watchdog_timer_callback()
         return;
     }
 
-    int latest_byte_count = IS_.ClientServerByteCount();
+    // TODO(SDK): InertialSense::ClientServerByteCount() does not exist in the current SDK (see
```
kylemallory, line 993: Replace with forward-looking todo.

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -971,10 +1056,11 @@ void InertialSenseROS::start_rtk_server(RtkBaseCorrectionProvider_Ntrip& config)
 {
     // [type]:[ip/url]:[port]
     std::string RTK_connection = config.get_connection_string();
-    if (IS_.CreateHost(RTK_connection))
-        RCLCPP_INFO_STREAM(rclcpp::get_logger("started_rtk_ntrip_server"),"InertialSenseROS: Successfully started RTK Base NTRIP correction server at" << RTK_connection);
-    else
-        RCLCPP_ERROR_STREAM(rclcpp::get_logger("failed_to_start_ntrip_server"),"InertialSenseROS: Failed to start RTK Base NTRIP correction server at " << RTK_connection);
+    // TODO(SDK): InertialSense::CreateHost() does not exist in the current SDK (see note in
```
kylemallory, line 1059: Replace with forward-looking todo.

ROS/ros2/src/inertial_sense_ros2.cpp
```diff
@@ -1915,27 +2040,27 @@ void InertialSenseROS::RTK_Rel_callback(eDataIDs DID, const gnss_rtk_rel_t *cons
         uint32_t fixStatus = msg->status & GNSS_STATUS_FIX_MASK;
         if (fixStatus == GNSS_STATUS_FIX_3D)
         {
-            rtk_rel.e_gps_status_fix = inertial_sense_ros2::msg::RTKRel::GNSS_STATUS_FIX_3D;
+            rtk_rel.e_gps_status_fix = inertial_sense_ros2::msg::RTKRel::GPS_STATUS_FIX_3D;
```
kylemallory, line 2043: Are these OUR status flags?  Or ROS's?  We transitioned away from GPS_STATUS_* to GNSS_STATUS_* because it is more contextually correct.  I would hope that ROS didn't do the opposite...

ROS/ros2/test/test_client_reconnect.cpp
```diff
@@ -26,7 +24,13 @@ void connect_rtk_client(const std::string& rtk_correction_protocol, const std::s
   {
     ++RTK_connection_attempt_count;
 
-    bool connected = IS_.OpenConnectionToServer(RTK_connection);
+    // TODO(SDK): InertialSense::OpenConnectionToServer() does not exist in the current SDK
```
kylemallory, line 27: Replace with forward-looking todo.

kylemallory (changes requested):

PR #1228, opened 2026-07-20
SN-8314 IMX-5 fwUpdate: reject oversized image before erase (ISbl page-capacity guard)

kylemallory:
### Problem
The r3.1.0 IMX-5 image's top data address is `0x0807049C` — 1180 bytes past `0x08070000`, into logical page 7. A **v6g** bootloader exposes only 7 logical pages (0..6), so it rejects `select_page(7)`; the SDK surfaces this as an opaque `ERR_FLASH_WRITE_FAILURE` at **99.8%** — after it has already erased and written pages 0..6. Neither update path (ISv1 `cISBootloaderISB` nor ISv2 `ISBFirmwareUpdater`) had a preflight size check.

### Fix (this PR — `ISBFirmwareUpdater` only)
A pre-erase size guard at the `UPLOADING → ERASING` transition — the earliest point at which both the ISbl version and the fully-buffered image are known:

- **`calcFinalImageTopAddress()`** — parses the buffered Intel-HEX for the highest absolute flash address written (the `.hex` text size is unrelated to the flashed extent; records are coalesced/gap-filled, so they must be parsed).
- **`getWritableFlashTopAddress()`** — maps the ISbl version to the writable top: `<= v6g -> 0x08070000` (7 pages); `v6j+ -> 0x08078000` (8-page window minus the WRP calibration region, see SN-8313). v6h/v6i were never released.
- If image-top exceeds the writable top, abort **before erase** with a message naming the version, addresses, and overflow.

### Validated on hardware (SN520487 / ISbl v6g)
- **r3.1.0 (oversized):** fails at ~10% with **0 erase operations** and a non-zero exit: `image writes up to 0x0807049C, but writable flash ends at 0x08070000 (1180 bytes over). Aborting before erase; flash is left unmodified.`
- **r3.0.0 (fits, top `0x0806F394`):** IMX-5 + GPX-1 + both CXD5610 update and validate normally.

### Known limitation (documented on SN-8314)
On v6g an aborted (no-write) update still leaves the device in ISbl: entering ISbl invalidates the app-validity hash, re-validated only by an actual write — there is no protocol command to re-hash existing flash, so there is no SDK-only recovery and no bootloader change is deployable (v6g shipping, IMX-5 EOL). Recover by flashing a compatible (smaller) image; the durable fix is compressing the IMX-5 firmware to fit v6g.

### DIFF
--- a/src/ISBFirmwareUpdater.cpp
+++ b/src/ISBFirmwareUpdater.cpp
@@ -9,6 +9,15 @@
 #include "ISBFirmwareUpdater.h"
 #include "InertialSense.h"
 
+// IMX-5 (STM32L4) flash geometry — mirrors the bootloader constants in
+// hw-libs/mcu/STM32L4/drivers/d_flash.h. Used to bound-check an image against the
+// target ISbl's writable flash window before erasing (SN-8314).
+#define IMX5_FLASH_START_ADDR       0x08000000u
+#define IMX5_LOGICAL_PAGE_SIZE      0x10000u    //!< 64K ISbl "logical page" (select_page granularity)
+#define IMX5_ISBL_PAGES_LEGACY      7           //!< v6g and earlier expose logical pages 0..6
+#define IMX5_ISBL_PAGES_EXPANDED    8           //!< v6j and later expose logical pages 0..7
+#define IMX5_WRP_RESERVED           0x8000u     //!< top 32K of the 8-page window is WRP calibration/config (not app-writable; see SN-8313)
+
 
 /**
  * This is an internal method used to send an update message to the host system regarding the status of the update process
@@ -186,8 +195,30 @@ bool ISBFirmwareUpdater::fwUpdate_step(fwUpdate::msg_types_e msg_type, bool proc
     if ((session_status >= fwUpdate::READY) && (session_status <= fwUpdate::FINALIZING)) {
         switch (updateState) {
             case UPLOADING: // transfer
-                if (transferProgress >= 1.0f)
+                if (transferProgress >= 1.0f) {
+                    // SN-8314: Preflight the fully-buffered image against the target ISbl's writable
+                    // flash window BEFORE erasing. Erasing invalidates the existing APP; if the new
+                    // image then can't be written (e.g. it needs a logical page this bootloader
+                    // rejects), the device would be stranded in ISbl with no valid firmware. This is
+                    // the earliest point at which BOTH the ISbl version and the complete image are known.
+                    uint32_t imageTop = calcFinalImageTopAddress();
+                    uint32_t flashTop = getWritableFlashTopAddress();
+                    if (imageTop > flashTop) {
+                        fwUpdate_sendProgressFormatted(IS_LOG_LEVEL_ERROR,
+                            "(ISB) Firmware image is too large for %s bootloader (ISbl v%1d%c): image writes up to 0x%08X, but writable flash ends at 0x%08X (%u bytes over). Aborting before erase; flash is left unmodified.",
+                            device->getIdAsString().c_str(),
+                            device->devInfo.firmwareVer[0], device->devInfo.firmwareVer[1],
+                            imageTop, flashTop, (unsigned)(imageTop - flashTop));
+                        // NOTE: The device is left in ISbl and cannot be returned to APP from here. On
+                        // entry to ISbl the bootloader invalidates the app-validity header (hash), and
+                        // it is only re-validated by an actual write -- there is no protocol command to
+                        // re-hash the existing (unmodified) flash. Recover by flashing a compatible
+                        // (smaller) image. See SN-8314.
+                        session_status = fwUpdate::ERR_NOT_ENOUGH_MEMORY;
+                        break;
+                    }
                     updateState = ERASING;
+                }
                 session_status = fwUpdate::IN_PROGRESS;
                 break;
             case ERASING: // prepare for erase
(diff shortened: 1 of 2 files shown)


aabogdanov (approved):

tmcd249 (approved):

tmcd249 (approved):

src/ISBFirmwareUpdater.cpp
```diff
@@ -1213,6 +1244,85 @@ is_operation_result ISBFirmwareUpdater::process_hex_stream(ByteBufferStream& byt
...
+    while ((lineLength = is_isb_read_line(*imgStream, line)) != 0) {
+        if ((lineLength < 11) || (line[0] != ':'))
+            continue;
+
+        uint32_t byteCount = parseHex(line + 1, 2);
+        uint32_t offset16  = parseHex(line + 3, 4);
+        uint32_t recType   = parseHex(line + 7, 2);
+
+        if (recType == 0x04) {          // Extended Linear Address: sets the page base for following data
+            ela = parseHex(line + 9, 4);
```
Hephaestus-Maximus, line 1286: Nitpick, non-blocking: only the generic `lineLength < 11` check guards this before parsing the ELA record's 4 data hex digits at offset 9-12. A well-formed `:02000004XXXXcc` record is always >= 15 chars, so a malformed/truncated type-04 record with `lineLength` between 11 and 14 would read past the actual record content into whatever's left over in `line[]` from a previous read (still memory-safe, since `line` is a fixed 1024-byte buffer, just a wrong `ela` value from stale bytes). Not a realistic concern given the input is a trusted build-toolchain-generated .hex, not adversarial input -- just flagging for completeness since it's new code.

src/ISBFirmwareUpdater.cpp
```diff
@@ -1213,6 +1244,85 @@ is_operation_result ISBFirmwareUpdater::process_hex_stream(ByteBufferStream& byt
...
+ * image, reads it record-by-record, and restores the original stream position on exit so it does
+ * not disturb the subsequent write pass.
+ *
+ * @return the exclusive top absolute flash address written by the image, or 0 if no image is buffered.
+ */
+uint32_t ISBFirmwareUpdater::calcFinalImageTopAddress() {
+    if (!imgStream)
+        return 0;
+
+    const std::size_t savedPos = imgStream->tellg();    // preserve the write pass's read position
```
Hephaestus-Maximus, line 1259: Nitpick, non-blocking: `imgStream->tellg()` returns a `std::streampos`/`std::streamoff`-compatible type, not `std::size_t` -- assigning to `std::size_t` works via implicit conversion for any realistic (small, non-error) position here, but `std::streamoff` (or just `auto`) would be the more idiomatically correct type to guard against a hypothetical `tellg()` failure (-1) being silently reinterpreted as a huge unsigned value. Low priority given `imgStream` is already null-checked and this only ever runs mid-upload on a stream that's actively being read from.

Hephaestus-Maximus (approved):
## Review summary — inertial-sense-sdk#1228

**Tier: Moderate/safety-relevant** — a self-contained pre-erase size guard for IMX-5 firmware updates. Read the diff in context, cross-referenced the version-encoding and error-code choices against the rest of the codebase, and traced the state-machine integration to confirm the new error path actually halts the update rather than silently continuing.

### What I verified

- **`firmwareVer[1]` as an ASCII letter (`'g'`, `'j'`) is internally consistent.** The same field is printed via `%c` in the new error message and compared via `minor <= 'g'` in `getWritableFlashTopAddress()` — both treat it as a raw ASCII byte, not a numeric index, and they agree with each other. Can't independently verify against real device firmware without hardware, but the PR's own hardware validation (SN520487 / ISbl v6g) exercises exactly this comparison path and the described behavior matches.
- **`m_isb_props.processor` is populated before this code can run.** Traced back to `fetch_device_info_and_signature()`, which sets it from the device's hardware ID during the initial bootloader handshake — necessarily before any UPLOADING state is reached.
- **`ERR_NOT_ENOUGH_MEMORY` is the semantically correct existing error code**, not a repurposed unrelated one: `protocol/FirmwareUpdate.h` documents it as "indicates that the requested firmware file size would exceed the available slot size" — exactly this scenario.
- **The new error path actually halts the update.** `session_status = ERR_NOT_ENOUGH_MEMORY` (-4) falls outside the `[READY(2)..FINALIZING(4)]` range gating the whole switch statement, so the very next `fwUpdate_step()` call no-ops instead of proceeding to `ERASING` — confirmed by reading the enum values directly, not just assuming.
- **`calcFinalImageTopAddress()`'s record parsing is consistent with the real write path's data domain.** The existing `process_hex_stream()` only supports a single hex digit of Extended-Linear-Address (comment: "Only support up to 10 pages currently") vs. this new function's full 4-digit parse — different generality, but since IMX-5 images only ever use ELA 0x0000-0x0007 (7-8 logical pages), both parse to the same value for any real image. Reuses the existing `is_isb_read_line()` utility rather than duplicating stream-reading logic.
- Description is excellent: precise problem statement with exact addresses, a clear before/after, an explicit hardware validation section naming the actual test device and both the failing (oversized) and passing (fits) cases, and an honest "known limitation" section about what this PR does *not* fix (v6g can't self-recover once stranded in ISbl — that's an inherent protocol limitation, not something fixable in this diff).

### Findings

- **Blocking:** none.
- **Nitpicks (2, inline, non-blocking):** the ELA-record length check could be marginally stricter for defense-in-depth (not a realistic risk given the input is a trusted build artifact, not adversarial); `std::size_t` for a stream position is slightly non-idiomatic vs. `std::streamoff`/`auto`.
- **Suggestion:** no unit tests exist for `calcFinalImageTopAddress()`/`getWritableFlashTopAddress()`, both of which are cleanly testable in isolation (feed a small hardcoded hex string; check the v6g/v6h boundary and the never-released v6h/v6i handling). Given the hardware validation already covers the critical real-world case, this is a nice-to-have for pinning the boundary logic long-term, not a blocker.

### Description grade: 5/5

No edit needed — this is close to a model bug-fix PR description: precise problem, precise fix, precise validation, honest about what's still broken and why.

### Verdict: **APPROVE**

Already has two human approvals (Alex, Tony) and passing CI (Linux/Windows; ROS1/2 bridge jobs are disabled repo-wide, unrelated to this change). My own read confirms the core logic (address parsing, version-boundary comparison, error-code choice, state-machine integration) is correct and the hardware validation targets exactly the failure mode this PR fixes.

PR #1229, opened 2026-07-20
SN-8294: Doxygen documentation for supplemental data structures and math

Hephaestus-Maximus:
## Summary

Part of [SN-8279](https://inertialsense.atlassian.net/browse/SN-8279) ([SDK] Doxygen documentation epic), story [SN-8294](https://inertialsense.atlassian.net/browse/SN-8294). Follows the exact style precedent established in SN-8293 (`data_sets.h`).

Documents all 15 files in the story's scope, in three logical groups:

1. **Data structure headers** — `data_sets_canbus.h`, `luna_data_sets.h`, `imx_defaults.h`, `rtk_defines.h`.
2. **Algorithmic APIs** — `convert_ins.h`, `time_conversion.h`, `filters.h`, `statistics.h`.
3. **GNSS and math** — `ISGnss.h`, `ISEarth.h`, `ISMatrix.h` (923 lines), `ISPose.h`, `ISPolynomial.h`, `IS_calibration.h`, `IS_calibration_convert.h`.

Every file gets an `@file` block with `@brief` + prose; struct/enum members get trailing `//!<` comments; functions get `@param`/`@return` with units and coordinate-frame conventions stated (radians, NED/ECEF/body frame, etc.), verified against each `.cpp`/`.c` implementation rather than guessed. **Exception**: per the epic's own reduced bar for `ISMatrix.h` ("at least a brief description" for its many small vector/matrix operations), that file gets concise one-line `@brief` blocks rather than exhaustive per-argument docs.

## Notable findings (flagging for reviewers)

While verifying docs against implementations, found and fixed two genuine pre-existing header/definition drifts — both zero-risk (confirmed no callers anywhere in the `imx` tree rely on the broken names):

- **`filters.h`**: `running_mean_filter`'s declared parameter order (`mean`, `input`) was backwards relative to its actual implementation (`input`, `mean`) in `filters.cpp` — a landmine for any future caller trusting the header. Reordered to match.
- **`statistics.h`**: the header declared `double mean(double*, ...)` and `double variance(double*, ...)` overloads that don't exist anywhere in the compiled library — only `mean_d`/`variance_d` are actually defined. Renamed the declarations to match the real, linkable symbols.

Also fixed several stale `@param` names that didn't match their signatures (`ISgpst2time`'s `weel` → `week`, `ISutc2gpst`'s `args` → `t`, a combined `t1, t2` `@param` that doesn't parse correctly, `UtcDateTimeToGpsTime`'s `datetime[7]`/`gpsLeapSeconds` mismatches).

Left the disabled `#if 0` blocks in `time_conversion.h` and `filters.h` untouched (dead code, not part of the compiled API).

## Test plan

- [x] Ran a baseline `doxygen` pass (installed locally) against all 15 files before touching anything, then after each of the three phases, diffing warnings to catch regressions immediately (caught and fixed a few — an unsupported `<Fc>`-looking tag, `@ref`s to `static` functions that Doxygen doesn't extract under this `Doxyfile`'s `EXTRACT_STATIC=NO`).
- [x] SDK-wide doxygen warning count: 717 → 703 (net improvement). Remaining 20 warnings in `filters.h`/`ISEarth.h` are pre-existing "multiple @param documentation sections" issues caused by duplicate doc comments in `filters.cpp`/`ISEarth.c`/`ISPose.c` — those files aren't in this story's scope, left untouched.
- [x] Every one of the 15 files confirmed to have an `@file` block (no file left with a plain MIT-only header).

[SN-8279]: https://inertialsense.atlassian.net/browse/SN-8279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SN-8294]: https://inertialsense.atlassian.net/browse/SN-8294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ISEarth.h
+++ b/src/ISEarth.h
@@ -91,24 +91,24 @@ void lla2ecef(const double *LLA, double *Pe);
 
 /**
  * @brief Find NED (north, east, down) offset from llaRef to lla, single precision.
- * @param llaRef Reference latitude (rad), longitude (rad), MSL altitude (m).
- * @param lla    Target latitude (rad), longitude (rad), MSL altitude (m).
+ * @param llaRef Reference latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
+ * @param lla    Target latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  * @param result Output: NED offset (meters) from llaRef to lla.
  */
 void lla2ned(ixVector3 llaRef, ixVector3 lla, ixVector3 result);
 
 /**
  * @brief Find NED (north, east, down) offset from llaRef to lla, double precision.
- * @param llaRef Reference latitude (rad), longitude (rad), MSL altitude (m).
- * @param lla    Target latitude (rad), longitude (rad), MSL altitude (m).
+ * @param llaRef Reference latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
+ * @param lla    Target latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  * @param result Output: NED offset (meters) from llaRef to lla.
  */
 void lla2ned_d(double llaRef[3], double lla[3], ixVector3 result);
 
 /**
  * @brief Find NED (north, east, down) offset from llaRef to lla, double precision, degrees input.
- * @param llaRef Reference latitude (degrees), longitude (degrees), MSL altitude (m).
- * @param lla    Target latitude (degrees), longitude (degrees), MSL altitude (m).
+ * @param llaRef Reference latitude (degrees), longitude (degrees), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
+ * @param lla    Target latitude (degrees), longitude (degrees), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  * @param result Output: NED offset (meters) from llaRef to lla.
  */
 void llaDeg2ned_d(double llaRef[3], double lla[3], ixVector3 result);
@@ -116,31 +116,31 @@ void llaDeg2ned_d(double llaRef[3], double lla[3], ixVector3 result);
 /**
  * @brief Find LLA from an NED (north, east, down) offset relative to llaRef, single precision.
  * @param ned    NED offset (meters) from llaRef.
- * @param llaRef Reference latitude (rad), longitude (rad), MSL altitude (m).
- * @param result Output: latitude (rad), longitude (rad), MSL altitude (m).
+ * @param llaRef Reference latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
+ * @param result Output: latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  */
 void ned2lla(ixVector3 ned, ixVector3 llaRef, ixVector3 result);
 
 /**
  * @brief Find LLA from an NED (north, east, down) offset relative to llaRef, double precision.
  * @param ned    NED offset (meters) from llaRef.
- * @param llaRef Reference latitude (rad), longitude (rad), MSL altitude (m).
- * @param result Output: latitude (rad), longitude (rad), MSL altitude (m).
+ * @param llaRef Reference latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
+ * @param result Output: latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  */
 void ned2lla_d(ixVector3 ned, double llaRef[3], double result[3]);
 
 /**
  * @brief Find LLA from an NED (north, east, down) offset relative to llaRef, double precision, degrees output (WGS-84 standard).
  * @param ned    NED offset (meters) from llaRef.
- * @param llaRef Reference latitude (degrees), longitude (degrees), MSL altitude (m).
- * @param result Output: latitude (degrees), longitude (degrees), MSL altitude (m).
+ * @param llaRef Reference latitude (degrees), longitude (degrees), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
+ * @param result Output: latitude (degrees), longitude (degrees), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  */
 void ned2llaDeg_d(ixVector3 ned, double llaRef[3], double result[3]);
 
 /**
  * @brief Find the delta-LLA equivalent of an NED (north, east, down) offset relative to llaRef, single precision.
  * @param ned     NED offset (meters) from llaRef.
- * @param llaRef  Reference latitude (rad), longitude (rad), MSL altitude (m).
+ * @param llaRef  Reference latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  * @param deltaLLA Output: delta latitude (rad), delta longitude (rad), delta altitude (m).
  */
 void ned2DeltaLla(ixVector3 ned, ixVector3 llaRef, ixVector3 deltaLLA);
@@ -148,7 +148,7 @@ void ned2DeltaLla(ixVector3 ned, ixVector3 llaRef, ixVector3 deltaLLA);
 /**
  * @brief Find the delta-LLA equivalent of an NED (north, east, down) offset relative to llaRef, double precision.
  * @param ned     NED offset (meters) from llaRef.
- * @param llaRef  Reference latitude (rad), longitude (rad), MSL altitude (m).
+ * @param llaRef  Reference latitude (rad), longitude (rad), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  * @param deltaLLA Output: delta latitude (rad), delta longitude (rad), delta altitude (m).
  */
 void ned2DeltaLla_d(ixVector3 ned, double llaRef[3], double deltaLLA[3]);
@@ -156,7 +156,7 @@ void ned2DeltaLla_d(ixVector3 ned, double llaRef[3], double deltaLLA[3]);
 /**
  * @brief Find the delta-LLA equivalent of an NED (north, east, down) offset relative to llaRef, double precision, degrees output.
  * @param ned     NED offset (meters) from llaRef.
- * @param llaRef  Reference latitude (degrees), longitude (degrees), MSL altitude (m).
+ * @param llaRef  Reference latitude (degrees), longitude (degrees), altitude (m; MSL or WGS-84, any datum consistent between llaRef and lla/result).
  * @param deltaLLA Output: delta latitude (degrees), delta longitude (degrees), delta altitude (m).
  */
 void ned2DeltaLlaDeg_d(ixVector3 ned, double llaRef[3], double deltaLLA[3]);
--- a/src/ISEarth.h
+++ b/src/ISEarth.h
@@ -10,6 +10,19 @@ The above copyright notice and this permission notice shall be included in all c
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT, IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 */
 
+/**
+ * @file ISEarth.h
+ * @brief Earth-relative coordinate transformations: ECEF/LLA/NED conversions, range/bearing,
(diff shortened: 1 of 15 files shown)


kylemallory (changes requested):
There are a number of comments which previously included units for represenational values, which have since had the units removed. And others which previously didn't have units, but now do.  The PR should be consistent.  If a unit of measure is listed, let's keep it, unless its clear from the description - ie latitude, or longitude (which are traditionally in degrees). If there are not units for values which clearly do have a unit, but may not be apparently (is an azim/elev in degrees or radian?) then we should always show the unit.  Most importantly, let's be consistent.  Review the proposed changes against the original, re-include the units if they were removed (unless obvious), and keep the areas where units were added.

Hephaestus-Maximus:
Addressed in 485551c2.

Audited the full diff against \`develop\` for every place a unit annotation was present in the original comment and absent in the rewrite, or vice versa. Almost everywhere the unit had simply moved (from a leading \`/** */\` block to a trailing \`//!<\`) with the text intact — one real inconsistency turned up:

The classic-CAN and CAN FD payload union summaries (\`is_can_payload\` / \`is_canfd_payload\`) described several fields without their units even though the underlying struct's own field docs stated them right above — e.g. the union said "WGS84 latitude" while \`is_can_ins_lat\`'s own doc said "WGS84 latitude (degrees)". Fixed by stating the unit in the summary line everywhere the referenced struct has one: degrees for lat/lon, meters for altitude/ECEF/NED offsets, m/s for velocities, radians for angles/rates — matching each struct's own docs, per your point about azimuth/elevation-style values where the unit isn't obvious without stating it.

Lat/lon in the individual struct docs (`is_can_ins_lat`, etc.) already stated "(degrees)" and were left as-is, consistent with your note that traditionally-obvious units are fine to state explicitly when already present, or omit only when truly unambiguous.

Note: there are no inline review comment threads on this PR to mark resolved via GitHub's thread mechanism — your review was a single top-level "Changes requested" with no per-line comments — so this reply is the resolution; happy to address anything else you spot on re-review.

tmcd249 (approved):

shefft (approved):

Hephaestus-Maximus:
CI is green now (commit 24363380): both **General Software Linux** and **General Software Windows** pass.

## Root cause

\`rtk_defines.h\`'s new \`@file\` block and the constellation-enable comment both wrote macro-name lists like \`MINPRN*/MAXPRN*/NSAT*\` as prose. The literal \`*/\` substring is a comment terminator regardless of surrounding text, so it closed the enclosing \`/** */\` block immediately — everything after (including \`@author\`, \`@copyright\`, and the \`ENAGAL\`/\`ENACMP\` \`#define\`s' own doc comment) got parsed as real code. That's what produced "stray '@' in program" and "unknown type name 'MAXPRN'" on Linux, and the equivalent MSVC errors on Windows.

Reworded both spots to name the macros without adjacent `*` + `/` sequences, then:
- Grepped every other file touched in this PR for the same pattern (`[A-Za-z0-9_]*/[A-Za-z0-9_]`) — `rtk_defines.h` was the only offender.
- Did a full local rebuild of the SDK (matching CI's toolchain) — clean.
- Ran the full unit test suite — all 414 tests pass.
- Re-ran doxygen — warning count unaffected.

kylemallory (approved):

src/ISEarth.h
```diff
@@ -47,130 +60,193 @@ typedef ixMatrix4     ixMatrix4;
...
-/*
- *  Find NED (north, east, down) from LLAref to LLA
- *
- *  lla[0] = latitude (rad)
- *  lla[1] = longitude (rad)
- *  lla[2] = msl altitude (m)
+/**
+ * @brief Find NED (north, east, down) offset from llaRef to lla, single precision.
+ * @param llaRef Reference latitude (rad), longitude (rad), MSL altitude (m).
+ * @param lla    Target latitude (rad), longitude (rad), MSL altitude (m).
```
aabogdanov: @Hephaestus-Maximus here and elsewhere: there should be no specific references to MSL. This function works with both MSL (if the input is MSL) and WGS-84 altitudes, so we should not constrain its description to MSL. I suggest removing MSL from function descriptions where altitude can be either MSL or WGS-84.

waltjohnson (approved):

src/ISMatrix.h
```diff
@@ -893,7 +976,15 @@ static __inline void O0_LPF_Vec3(ixVector3 result, const ixVector3 input, f_t al
...
-// First order Low-Pass Filter
+/**
+ * @brief First-order low-pass filter update with an explicit model-coefficient state (for larger dt).
+ * @param result Filter output and running state; updated in place.
+ * @param input  New input sample.
+ * @param c1     Model-coefficient state; updated in place.
+ * @param alph   Filter alpha parameter (input gain).
+ * @param beta   Filter beta parameter (memory gain).
+ * @param dt     Time since the last update (seconds).
+ */
```
aabogdanov, line 995: This function is not a traditional 1st order LPF. Essentially, it is a Fused Kinematic Tracker with LPF. It estimates derivative, then applies LPF to the estimated derivative, then integrates the state using the filtered estimated velocity.

src/ISMatrix.h
```diff
@@ -872,16 +915,56 @@ static __inline int isFinite_array_d(double *a, int size)
...
+ * @param initVal      Initial value for the filter's running output.
+ */
 void LPFO0_init_Vec3(sLpfO0 *lpf, f_t dt, f_t cornerFreqHz, const ixVector3 initVal);
-// output[n+1] = beta*output[n] + alpha*input
+
+/**
+ * @brief Update a zero-order low-pass filter with a new sample: v[n+1] = beta*v[n] + alpha*input.
+ * @param lpf   Filter state; updated in place.
+ * @param input New input sample.
+ */
```
aabogdanov, line 960: Usually this is called 1st order LPF

src/ISMatrix.h
```diff
@@ -872,16 +915,56 @@ static __inline int isFinite_array_d(double *a, int size)
...
 void LPFO0_Vec3(sLpfO0 *lpf, const ixVector3 input);
 
-// Zero order Low-Pass Filter
+/**
+ * @brief Zero-order (single-pole) low-pass filter update, standalone (no sLpfO0 state struct needed): result = beta*result + alpha*input.
+ * @param result Filter output and running state; updated in place.
+ * @param input  New input sample.
+ * @param alph   Filter alpha parameter (input gain).
+ * @param beta   Filter beta parameter (memory gain).
```
aabogdanov, line 969: Usually this is called 1st order LPF

aabogdanov (changes requested):

src/ISMatrix.h
```diff
@@ -24,36 +41,39 @@ extern "C" {
...
 #define RECIPNORM_VEC4(v)   (1.0f/_MAX(MAG_VEC4(v),  EPSF32))
 #define RECIPNORM_VEC3D(v)  (1.0l/_MAX(MAG_VEC3D(v), EPSF64))
 #define RECIPNORM_VEC4D(v)  (1.0l/_MAX(MAG_VEC4D(v), EPSF64))
 
+/** Unwrap each element of a 3-vector of radian angles into the canonical +-pi range (see UNWRAP_RAD_F32). */
 #define UNWRAP_VEC3(v)          {UNWRAP_RAD_F32(v[0]); UNWRAP_RAD_F32(v[1]); UNWRAP_RAD_F32(v[2]) }
 
+/** Type-correct zero literal for x's type (float/double/long double/int), used by the VEC3_*_ZERO macros below. */
 #ifdef __cplusplus
 #define ZERO_OF(x)                          static_cast<std::remove_reference_t<decltype(x)>>(0)
```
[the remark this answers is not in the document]
Hephaestus-Maximus: Assessed as invalid — checked every build path in the dependency chain (this SDK, imx firmware Makefiles for both IMX-5 and GPX-1, is-imx, is-gpx) and found no C++11 or C++14 target anywhere:

- SDK root `CMakeLists.txt`: C++17; `tests/CMakeLists.txt`: C++20
- `cltool`: C++17; ROS1/ROS2 wrappers and all `ExampleProjects/*`: C++20
- imx firmware `IS_IMX` Makefile: `-std=gnu++17`; `IS_GPX-1` Makefile: `-std=gnu++20`
- is-gpx test suites: C++20 (no override found in is-imx, which inherits Zephyr's own C++17+ minimum)

`std::remove_reference_t` is a C++14 feature, so every one of these is well above the requirement. Since I can't find an actual C++11/14-only build path anywhere this header is consumed, leaving `ZERO_OF(x)` as-is rather than adding unnecessary compatibility code for a build configuration that doesn't exist in this codebase. Happy to revisit if there's a specific toolchain/target I missed.

src/time_conversion.h
```diff
@@ -63,34 +108,56 @@ int gpsTowMsToUtcWeekday(int gpsTowMs, int leapS);
 std::tm stdGpsTimeToUtcDateTime(uint32_t gpsSecondsOfWeek, uint32_t gpsWeek, int leapSeconds);
```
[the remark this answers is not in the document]
Hephaestus-Maximus: Valid — confirmed against the signature. `stdGpsTimeToUtcDateTime(uint32_t gpsSecondsOfWeek, uint32_t gpsWeek, int leapSeconds)` takes both by value and returns a `std::tm`, so they're inputs, not outputs. The doc was copy-pasted from the neighboring `stdUtcDateTimeToGpsTime()` (which genuinely outputs those same-named params by reference) without updating the direction. Fixed both `@param` lines to say Input instead of Output.

src/convert_ins.h
```diff
@@ -10,14 +10,62 @@ The above copyright notice and this permission notice shall be included in all c
...
+ * @param result Output INS-1 solution: Euler angles w.r.t. NED, same LLA position, and NED offset
+ *               from refLla (if provided).
+ * @param refLla Optional reference WGS84 latitude/longitude/altitude (degrees, degrees, meters)
+ *               used to compute result->ned. If NULL, result->ned is zeroed.
+ */
 void convertIns2ToIns1(ins_2_t *ins2, ins_1_t *result, double *refLla=NULL);
+
+/**
+ * @brief Convert ins_3_t (quaternion/NED, ECEF position) to ins_1_t (Euler/NED, LLA position).
+ * @param ins3   Source INS-3 solution (quaternion body rotation w.r.t. NED, LLA position).
```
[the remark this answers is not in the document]
Hephaestus-Maximus: Valid — verified against the actual struct definition in data_sets.h: `ins_3_t` has `double lla[3]` (WGS84 lat/lon + ellipsoid height) and `float msl`, no ECEF field at all; `ins_4_t` is the one with `double ecef[3]`. The `@brief` line was also self-inconsistent with the `@param` description directly below it in the same doc block, which already correctly said "LLA position" for ins3. Fixed the `@brief` line to match.

src/ISEarth.h
Hephaestus-Maximus: Agreed and fixed. The lla2ned/ned2lla/ned2DeltaLla family (and their _d/Deg variants) works on whatever altitude datum llaRef and lla/result share -- the NED math doesn't care whether it's MSL or WGS-84 ellipsoid height. Removed "MSL" from all 15 occurrences across this function family; left the one genuinely MSL-specific function (barometric-pressure-to-MSL-altitude conversion) untouched since that one really is about MSL specifically.

src/ISMatrix.h
Hephaestus-Maximus: Agreed and fixed — reworded to note this is usually called a 1st-order LPF, without renaming the O0/O1 symbols themselves (the O0/O1 naming refers to how many extra derivative states the filter tracks, not classic DSP filter order, per the difference between this and O1_LPF_Vec3 below).

src/ISMatrix.h
Hephaestus-Maximus: Agreed and fixed — same as the LPFO0_init_Vec3/sLpfO0 thread above, reworded to note this is usually called a 1st-order LPF.

src/ISMatrix.h
Hephaestus-Maximus: Valid — traced the actual implementation and it matches your description exactly: (1) estimates the derivative as (input-result)/dt, (2) low-pass filters that derivative estimate into c1 via O0_LPF_Vec3(), (3) integrates c1*dt onto the previous result to predict the current state, (4) blends that prediction with the new raw input via another O0_LPF_Vec3() stage. Rewrote the doc to describe it as a fused kinematic tracker with LPF smoothing rather than a plain first-order low-pass filter.

Hephaestus-Maximus:
Addressed all outstanding review feedback in \`769adf5e\` and \`a583849d\`:

**From Copilot (all valid, fixed):**
- `time_conversion.h`: `stdGpsTimeToUtcDateTime()`'s `@param` direction was mislabeled (inputs called outputs).
- `convert_ins.h`: `ins_3_t` docstring wrongly said "ECEF position" — it's WGS84 LLA (ECEF is `ins_4_t`).
- `ISMatrix.h`: `ZERO_OF(x)`'s `std::remove_reference_t` C++14 concern — assessed as not applicable to this codebase; posted the specific evidence in-thread (every build path found, from this SDK through imx firmware to is-imx/is-gpx, targets C++17 or newer).

**From @aabogdanov (all valid, fixed):**
- `ISEarth.h`: removed MSL-specific wording from the `lla2ned`/`ned2lla`/`ned2DeltaLla` function family (15 occurrences) — they work with any altitude datum, not just MSL.
- `ISMatrix.h`: reworded the O0 filter family's docs to note they're usually called a 1st-order LPF (without renaming the O0/O1 symbols, which refer to tracked derivative-state count, not DSP filter order).
- `ISMatrix.h`: `O1_LPF_Vec3` isn't a traditional LPF — rewrote its doc to accurately describe it as a fused kinematic tracker with LPF smoothing, per your description of the actual algorithm.

Re-verified after each fix: doxygen zero warnings on the touched files, full local rebuild clean, full unit test suite green (414 passed, 3 skipped, 0 failed). All review threads resolved. Ready for re-review whenever convenient.

aabogdanov (approved):

PR #1233, opened 2026-07-22
SN-8320 evaltool silently skips update

waltjohnson:
This pull request improves how firmware update notifications are handled and displayed, especially when an update is skipped due to version checks. The changes ensure that user-facing messages accurately reflect the outcome and reason for skipped updates, and that notifications are tracked and surfaced distinctly from errors or generic success states.

**Notification handling and user messaging:**

* When an update is skipped because the firmware is not newer, the user now sees a clear message: "Update Skipped — Firmware is not newer (Use 'Force Update' to override)", and this is recorded as a notification rather than an error. The notification is preserved and surfaced in the UI, even if no upload session was started.
* The updater now distinguishes between a plain successful update and a success with notifications (such as a version skip), ensuring that the most relevant message (like a notification) is shown to the user instead of a generic success message.

**State management improvements:**

* The `ISFwUpdateState` class now tracks a `hasNotifications` flag, which is properly copied, assigned, and reset with state changes, ensuring notification state is accurately maintained across operations. [[1]](diffhunk://#diff-f5472e2adc7abb2d101aa2b677b1f356e4d0aef626ff2b1e5f6df5c142eaff0eL188-R188) [[2]](diffhunk://#diff-f5472e2adc7abb2d101aa2b677b1f356e4d0aef626ff2b1e5f6df5c142eaff0eR200) [[3]](diffhunk://#diff-f5472e2adc7abb2d101aa2b677b1f356e4d0aef626ff2b1e5f6df5c142eaff0eR230) [[4]](diffhunk://#diff-f5472e2adc7abb2d101aa2b677b1f356e4d0aef626ff2b1e5f6df5c142eaff0eR241)
* The update step logic in `ISFirmwareUpdater::step()` sets the update state to `SUCCESS_WITH_NOTIFICATIONS` if any notifications were recorded, rather than treating all non-error completions as plain successes.

### DIFF
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -1209,7 +1209,7 @@ void ISFirmwareUpdater::cmd_UploadImage(ISFwUpdaterCmd& cmd) {
                         "Ignoring Update. New firmware %s is older than device %s.",
                         imageVerStr.c_str(), targetVerStr.c_str());
                     cmd.status = ISFwUpdaterCmd::CMD_SUCCESS;
-                    cmd.resultMsg = "Update Skipped — Firmware Not Newer (Use \"Force Update\" to Override)";
+                    cmd.resultMsg = "Update Skipped — Firmware is not newer (Use \"Force Update\" to override)";
                     // Record this as a notification (not an error) so the caller can surface it
                     // distinctly from a plain "Completed Successfully" — otherwise a version-skip
                     // looks identical to a real upload in the UI.
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -1215,6 +1215,13 @@ void ISFirmwareUpdater::cmd_UploadImage(ISFwUpdaterCmd& cmd) {
                     // looks identical to a real upload in the UI.
                     {
                         auto lk = updateState.lock();
+                        // Because we never called initializeUpload() (no session was actually
+                        // started), updateState.target is still stale for this target. Sync it
+                        // here — otherwise refreshUpdateState()'s "session target changed" check
+                        // (which runs right after this, in the same step()) sees a mismatch and
+                        // clears updateState.messages, wiping the notification we just added
+                        // before it's ever displayed.
+                        updateState.target = target;
                         updateState.messages.emplace_back(activeStep, cmd, IS_LOG_LEVEL_WARN, cmd.resultMsg);
                         updateState.hasNotifications = true;
                     }
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -531,12 +531,19 @@ bool ISFirmwareUpdater::step() {
     fnStep.mark("Finished fwUpdate_step().");
 
     if (fwUpdate_isDone()) {
-        if (hasErrors())
-            updateState.state = ISFwUpdateState::UPDATER_DONE_WITH_ERRORS;
-        else if (updateState.hasNotifications)
-            updateState.state = ISFwUpdateState::SUCCESS_WITH_NOTIFICATIONS;
-        else
-            updateState.state = ISFwUpdateState::UPDATER_SUCCESSFUL;
+        bool errored = hasErrors();
+        {
+            // hasNotifications is written elsewhere (e.g. cmd_UploadImage) under updateState.lock();
+            // take the same lock here so this read-and-write of updateState fields is synchronized
+            // against concurrent readers/writers (e.g. the UI thread via getSnapshot()).
+            auto lk = updateState.lock();
+            if (errored)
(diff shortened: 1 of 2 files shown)


tmcd249 (approved):

src/ISFirmwareUpdater.cpp
```diff
@@ -1200,7 +1209,22 @@ void ISFirmwareUpdater::cmd_UploadImage(ISFwUpdaterCmd& cmd) {
                         "Ignoring Update. New firmware %s is older than device %s.",
                         imageVerStr.c_str(), targetVerStr.c_str());
                     cmd.status = ISFwUpdaterCmd::CMD_SUCCESS;
-                    cmd.resultMsg = "Target is already up to date.";
+                    cmd.resultMsg = "Update Skipped — Firmware is not newer (Use \"Force Update\" to override)";
```
kylemallory, line 1219: You should use the affirmative:: "Update Skipped - Target firmware is newer (use "Force Update" to override)".

kylemallory (approved):

src/ISFirmwareUpdater.cpp
```diff
@@ -527,7 +531,12 @@ bool ISFirmwareUpdater::step() {
     fnStep.mark("Finished fwUpdate_step().");
 
     if (fwUpdate_isDone()) {
-        updateState.state = hasErrors() ? ISFwUpdateState::UPDATER_DONE_WITH_ERRORS : ISFwUpdateState::UPDATER_SUCCESSFUL;
+        if (hasErrors())
+            updateState.state = ISFwUpdateState::UPDATER_DONE_WITH_ERRORS;
+        else if (updateState.hasNotifications)
+            updateState.state = ISFwUpdateState::SUCCESS_WITH_NOTIFICATIONS;
+        else
+            updateState.state = ISFwUpdateState::UPDATER_SUCCESSFUL;
```
Hephaestus-Maximus: Re: Copilot's lock-discipline comment on this line — verified it's a real concern, but it's a pre-existing, file-wide pattern rather than something this PR introduces: the class docs an "always lock" contract, but in practice only 4 of the ~20 call sites that mutate `updateState` fields in this file take `updateState.lock()` (this `step()` completion block has written `updateState.state` unlocked since before this PR). The new `updateState.hasNotifications` read here just follows that same, already-established pattern at this exact call site.

Practically, the specific write (in `cmd_UploadImage`, under lock) and this read happen on the same thread within one `step()` call, so there's no actual concurrent-modification race for *this* interaction — the theoretical risk is against a separate UI thread's `getSnapshot()`, which predates this change and affects nearly every field here, not just the new one.

Not blocking this fix on it, but worth a follow-up ticket to either lock consistently throughout `step()`/`refreshUpdateState()`, or drop the "always lock" claim from the class doc comment if partial locking is intentional (e.g. only for the fields the UI thread actually reads via snapshot).

Hephaestus-Maximus (approved):
Small, well-targeted fix — read the diff and cross-checked it against a checkout of the branch (not just the diff) to verify the `updateState.target` sync claim and the `hasNotifications` propagation through `step()`/`refreshUpdateState()`.

The core logic checks out:
- `ISFwUpdateState::SUCCESS_WITH_NOTIFICATIONS` is now correctly distinguished from a plain success in `step()`, and `refreshUpdateState()` correctly prefers the last recorded notification message over the generic "Completed successfully."
- The `updateState.target = target;` sync added inside the version-skip branch of `cmd_UploadImage` is a real, necessary fix, not just defensive padding — without it, `refreshUpdateState()`'s "session target changed" check (a few lines later, same `step()` call) would see a stale `updateState.target`, clear `updateState.messages`, and wipe the notification before it's ever shown. Confirmed this by reading `refreshUpdateState()` directly.
- `messages.emplace_back(activeStep, cmd, IS_LOG_LEVEL_WARN, cmd.resultMsg)` matches the existing call signature/pattern used elsewhere in the file (line 653), so no type mismatch.

I independently checked the one Copilot comment on this PR (unlocked read of `hasNotifications`) — it's a real observation, but it's consistent with a pre-existing, file-wide gap (only 4 of ~20 `updateState` mutation sites in this file actually take `updateState.lock()`, including the line above it that's written the exact same unlocked way pre-PR), not a new regression from this change, and the specific read/write pair here happens on the same thread within one `step()` call. Left a comment with that context rather than treating it as blocking — full lock-discipline cleanup for this class looks like a separate follow-up.

One minor, non-blocking gap: no test coverage exists for this notification path in `tests/test_ISFirmwareUpdate.cpp` (nothing references `hasNotifications`/`SUCCESS_WITH_NOTIFICATIONS` there today), so this is a manually-verified UX fix rather than a regression-tested one. Not asking to hold up a small text/state fix for that, just flagging the gap.

Approving — logic is sound, CI is green, and two maintainers have already signed off.

PR #1234, opened 2026-07-22
SN-8163 IMX-5 mag gauss uT fix

waltjohnson:
This pull request makes minor improvements to the constants defined for unit conversions in `src/ISConstants.h`. The changes ensure floating-point suffix consistency and introduce a new conversion constant.

* Constants for Tesla-to-Gauss and microTesla-to-Gauss conversions now explicitly use the `f` suffix for floating-point precision.
* Added a new constant `C_GAUSS2UTESLA_F` for Gauss-to-microTesla conversion.

### DIFF
--- a/src/ISConstants.h
+++ b/src/ISConstants.h
@@ -793,8 +793,9 @@ static inline int is_finite(double v)
 #define C_WIE               7.2321151467e-05   // WGS-84 earth rotation rate (rad/s)
 #define C_WIE_F             7.2321151467e-05f
 
-#define C_TESLA2GAUSS_F     10000.0
-#define C_UTESLA2GAUSS_F    0.01
+#define C_TESLA2GAUSS_F     10000.0f
+#define C_UTESLA2GAUSS_F    0.01f
+#define C_GAUSS2UTESLA_F    100.0f
 
 #define C_0p1_DEG2RAD_F     0.00174532925199433f        // = 0.1f * C_0pDEG2RAD_F
 #define C_0p11_DEG2RAD_F    0.00191986217719376f


kylemallory (approved):

Hephaestus-Maximus (approved):
Small SDK-only change, but I traced it through to its actual consumer rather than reviewing `ISConstants.h` in isolation, since this ticket (SN-8163, Highest priority, blocking manufacturing) is fixed across three repos.

**Math check:**
- `C_TESLA2GAUSS_F = 10000.0f` — correct (1 T = 10⁴ G).
- `C_UTESLA2GAUSS_F = 0.01f` — correct (1 µT = 10⁻⁶ T = 10⁻² G).
- `C_GAUSS2UTESLA_F = 100.0f` — correct, and is the exact reciprocal of `C_UTESLA2GAUSS_F` (1/0.01 = 100), so the two stay consistent with each other.
- The `f` suffix fix on the two pre-existing constants doesn't change behavior at any existing call site (10000 and 0.01 round identically in float vs. double for this purpose) — it's a real embedded-systems win, though: several Cortex-M cores common in this product line lack hardware double support, so avoiding an implicit double promotion in a tight sensor loop is worth doing regardless of the mag fix.

**Cross-repo consumer check:** found and read the actual usage in the companion PR (`imx#1618`, not currently requesting my review) — `sensors.cpp`'s `proc_sensors()`:
```c
float mag_tcal = g_sensor.ucal.mag[d].xyz[a] - g_sensor.tcBias.mpu[d].mag[a];   // Remove Temp Comp Biases
#if defined(IMX_5)
mag_tcal *= C_GAUSS2UTESLA_F;       // Convert to uT because IMX-5 temp comp is stored in gauss.
#endif
g_sensor.tcal.mag[d].xyz[a] = mag_tcal;
```
The conversion is applied to the *difference* (bias-subtracted result), which is dimensionally correct only if both `ucal.mag` and `tcBias.mpu.mag` are already in the same unit (Gauss) at that point — consistent with the comment and with the Jira thread's mention that the mag temp-comp table was reintroduced in Gauss. It's also correctly scoped to `IMX_5` only, matching the ticket title. So this SDK constant is a real, necessary, correctly-consumed dependency of the fix — not dead/speculative infrastructure.

Approving. Also worth noting for anyone picking this up: the PR description doesn't mention the bug or the companion `imx#1618`/`is-imx#38` PRs, so a reader looking at this SDK PR in isolation wouldn't know why `C_GAUSS2UTESLA_F` exists — left as an FYI rather than editing the description, since the mechanical description is accurate as-is.

PR #1241, opened 2026-07-25
SN-8347: fail loudly on malformed manifest; untruncated fwUpdate errors

kylemallory:
## Problem

Loading a release `.fpkg` fails with **error -8** (`PKG_ERR_IMAGE_UNKNOWN_PATH`, *"Manifest image path"*) when the manifest references a target image that has no `filename`/backing file — e.g. an IMX-6 stanza left behind after the IMX-6 firmware wasn't bundled in a GPX/IMX-5 release. `processPackageManifest()` validates every referenced image up front, before runtime device targeting, so the dangling reference aborts the whole load.

## Design principle

A manifest that **references** a target but carries no valid image for it is a **malformed manifest** — a genuine critical error, not an "optional, skip it" case. The dangerous failure mode: a customer who *has* an IMX-6 receives a package that looks like it upgrades IMX-6 but silently can't, because we shipped an incomplete manifest.

- Image fully bundled, unit lacks that device → handled at **runtime** by the step's `target …,:JUMP` skip.
- Target genuinely not shipped → its stanzas must be removed from the manifest entirely (generator side — see the paired `is-common` PR).
- Target referenced but image malformed → **hard `-8` error.**

## Changes (`src/ISFirmwareUpdater.cpp`, `.h`)

- **Parser:** a referenced image with no scalar `filename` now returns a hard `-8` (reverted an earlier tolerant-skip approach). Fixed the latent `&&`→`||` so a defined-but-non-scalar `filename` is caught too. `-8` message reworded to state the manifest is malformed.
- **Logging:** `handleCommandError()` formatted into a fixed `char[256]`, truncating long messages mid-word (a long package path made the message 276 bytes, cut at "Manifest image path"). Now formats into a dynamically-sized `std::string`, logged via `"%s"`. Package/manifest parse errors now append ` [step: …] [image: …]` context via new members `pkgErrorStep`/`pkgErrorImage`.

## Testing

- SDK library compiles clean.
- Not runtime-tested in isolation (no device-free harness — `step()` dereferences `device->port`); the generator side (paired PR) is exercised end-to-end and across omit-target combinations.

Jira: SN-8347

### DIFF
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -650,18 +650,24 @@ void ISFirmwareUpdater::handleCommandError(ISFwUpdaterCmd& cmd, int errCode, con
 
     // Format into a dynamically-sized buffer. A fixed buffer here previously truncated long
     // messages mid-word (e.g. a firmware package error whose path pushed the message past 256
-    // bytes). Size the string to the exact formatted length so nothing is lost.
+    // bytes). Grow-and-retry rather than probe with a null destination: some vsnprintf
+    // implementations (older MSVC CRTs) return -1 on truncation instead of the needed length,
+    // so a single nullptr/0 probe can't be trusted to size the buffer correctly everywhere.
     std::string buffer;
     va_list args;
     va_start(args, errMsg);
-    va_list args_copy;
-    va_copy(args_copy, args);
-    int needed = VSNPRINTF(nullptr, 0, errMsg, args_copy);
-    va_end(args_copy);
-    if (needed > 0) {
-        buffer.resize((size_t)needed + 1);
-        VSNPRINTF(&buffer[0], buffer.size(), errMsg, args);
-        buffer.resize((size_t)needed);      // trim the trailing NUL from the string's view
+    size_t bufSize = 256;
+    for (int attempt = 0; attempt < 8; ++attempt) {
+        buffer.resize(bufSize);
+        va_list args_copy;
+        va_copy(args_copy, args);
+        int written = VSNPRINTF(&buffer[0], buffer.size(), errMsg, args_copy);
+        va_end(args_copy);
+        if (written >= 0 && (size_t)written < bufSize) {
+            buffer.resize((size_t)written);    // trim the trailing NUL from the string's view
+            break;
+        }
+        bufSize = (written > 0) ? (size_t)written + 1 : bufSize * 2;
     }
     va_end(args);
 
@@ -1575,6 +1581,7 @@ ISFirmwareUpdater::pkg_error_e ISFirmwareUpdater::processPackageManifest(YAML::N
 
             auto labelName = label.as<std::string>();
             pkgErrorStep = labelName;   // track context so any subsequent error can name this step
+            pkgErrorImage.clear();     // a step-level error shouldn't carry over the previous step's image
 
             if (!cmds.IsSequence())
                 return PKG_ERR_NO_ACTIONS; // actions must be a sequence (of maps)
--- a/src/ISFirmwareUpdater.cpp
+++ b/src/ISFirmwareUpdater.cpp
@@ -648,10 +648,21 @@ void ISFirmwareUpdater::handleCommandError(ISFwUpdaterCmd& cmd, int errCode, con
         fwUpdate_resetEngine();
     }
 
-    char buffer[256];
+    // Format into a dynamically-sized buffer. A fixed buffer here previously truncated long
+    // messages mid-word (e.g. a firmware package error whose path pushed the message past 256
+    // bytes). Size the string to the exact formatted length so nothing is lost.
+    std::string buffer;
     va_list args;
     va_start(args, errMsg);
-    VSNPRINTF(buffer, sizeof(buffer), errMsg, args);
+    va_list args_copy;
+    va_copy(args_copy, args);
+    int needed = VSNPRINTF(nullptr, 0, errMsg, args_copy);
+    va_end(args_copy);
+    if (needed > 0) {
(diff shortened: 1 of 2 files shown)


src/ISFirmwareUpdater.cpp
```diff
@@ -1552,18 +1573,22 @@ ISFirmwareUpdater::pkg_error_e ISFirmwareUpdater::processPackageManifest(YAML::N
             if (!label.IsScalar())
                 return PKG_ERR_INVALID_TARGET; // key must identify a target name
 
+            auto labelName = label.as<std::string>();
+            pkgErrorStep = labelName;   // track context so any subsequent error can name this step
```
Hephaestus-Maximus, line 1583: Minor diagnostic-accuracy gap, non-blocking: `pkgErrorImage` is only cleared inside the inner per-command loop (line 1586), not here alongside `pkgErrorStep`. If a step errors out *before* entering that loop — e.g. `PKG_ERR_NO_ACTIONS` a few lines down when `!cmds.IsSequence()` — `pkgErrorImage` still holds whatever value was left over from the *previous* step's last-processed image. The resulting error message would then show a `[image: ...]` context that belongs to a different step than the one actually named in `[step: ...]`.

Suggest adding `pkgErrorImage.clear();` right after this line, so a step-level error (not an image-level one) never carries over a stale image name. Doesn't affect the actual `PKG_ERR_IMAGE_UNKNOWN_PATH` fail-loudly behavior this PR is about — just the precision of the diagnostic text in an edge case.

Hephaestus-Maximus (approved):
In-depth review of the manifest-validation and error-formatting fix. This touches firmware-package parsing correctness, so I read it carefully and verified the specific claims rather than taking them at face value — cloned the branch to check surrounding code and control flow.

**The core `&&` → `||` fix is correct**, verified by truth table: the intent is "fail unless the filename is both defined and scalar" (`!(defined && scalar)`), which by De Morgan's law is `!defined || !scalar` — the `||` this PR lands on. The old `&&` only tripped when the filename was *both* undefined and non-scalar simultaneously, which is never true for an undefined YAML node (an undefined node isn't scalar either, but that makes the old condition `true && true` in that case, not the gap) — the actual gap was a *defined-but-non-scalar* filename (a list/map), which would fall through to `.as<std::string>()` uncaught. Confirmed this scenario is now correctly caught.

**The `handleCommandError` dynamic-buffer rewrite is correct.** Standard two-pass `vsnprintf` idiom — `va_copy` before the length-probing call so the original `args` is still valid for the real formatting call, `resize(needed+1)` before writing (safe post-C++11 to take `&buffer[0]` since size ≥ 1), then trim back to `needed` afterward. I checked the `VSNPRINTF` macro: it resolves to a custom embedded-safe `vsnprintf_` on `PLATFORM_IS_EMBEDDED` builds (not just desktop libc) — I can't build/run the embedded target myself in this environment, but that library is specifically designed to support the null-buffer/zero-size length-probe call this code relies on, so I'm reasonably confident but flagging that I couldn't verify it on-target.

**The `LOG_FWUPDATE_STATUS(IS_LOG_LEVEL_ERROR, "%s", buffer.c_str())` change is a genuine, if minor, security-relevant fix**, not just style — previously `buffer` (which can contain a firmware package's file path, i.e. externally-influenced content) was passed directly as the *format string* to the status callback, so a `%` in a path would have been reinterpreted as a format specifier. Passing it through a literal `"%s"` closes that.

**One minor, non-blocking diagnostic-accuracy gap, left inline:** `pkgErrorImage` is only cleared inside the per-command loop, not alongside `pkgErrorStep` at the top of each step. A step-level error that returns before entering that loop (e.g. `PKG_ERR_NO_ACTIONS`) can carry over a stale image name from the previous step's last-processed image into the `[image: ...]` diagnostic context. Doesn't affect the actual fail-loudly behavior, just the precision of the log message in that edge case.

CI is green. Approving — the fix logic is sound and well-reasoned, and the PR description is honest about the one thing it couldn't verify (no device-free harness for runtime testing).

waltjohnson (approved):

aabogdanov (approved):

shefft (approved):

src/ISFirmwareUpdater.cpp
Hephaestus-Maximus: Fixed — added `pkgErrorImage.clear();` alongside `pkgErrorStep = labelName;` so a step-level error can't carry a stale image name from a previous step. Pushed in 1a0eda77.

src/ISFirmwareUpdater.cpp
```diff
@@ -648,10 +648,21 @@ void ISFirmwareUpdater::handleCommandError(ISFwUpdaterCmd& cmd, int errCode, con
...
-    VSNPRINTF(buffer, sizeof(buffer), errMsg, args);
+    va_list args_copy;
+    va_copy(args_copy, args);
+    int needed = VSNPRINTF(nullptr, 0, errMsg, args_copy);
+    va_end(args_copy);
+    if (needed > 0) {
+        buffer.resize((size_t)needed + 1);
+        VSNPRINTF(&buffer[0], buffer.size(), errMsg, args);
+        buffer.resize((size_t)needed);      // trim the trailing NUL from the string's view
+    }
```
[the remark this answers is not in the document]
Hephaestus-Maximus: Fixed — replaced the nullptr/0 length-probe with a grow-and-retry loop (starts at 256 bytes, doubles when the return value doesn't confirm a fit, resizes exactly when it does), so a non-conformant vsnprintf returning -1 on truncation is handled instead of relied-upon. Pushed in 1a0eda77.

PR #1243, opened 2026-07-27
SN-8335 SDK log reader example project exception if missing CL arg

shefft:
[SDK] Example Project InertialSense_log_reader_RegCmp does not gracefully handle lack of CL arg

If the file path argument is bogus, the program exits gracefully and prints an error message.

```
./ISLogReaderExample filename
Failed to load log files: ./ISLogReaderExample
```
There is an app usage line in the code, but if the argument is empty it does not reach that and appears to throw an uncaught exception instead.  This case should also be handled gracefully and provide instruction to the user.

```
./ISLogReaderExample
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string: construction from null is not valid
Aborted (core dumped)
```

Found and fixed argc condition check problem, AI reports the fix should work across all platforms consistently.

### DIFF
--- a/ExampleProjects/InertialSense_log_reader_RegCmp/LogReader_RegCmp.cpp
+++ b/ExampleProjects/InertialSense_log_reader_RegCmp/LogReader_RegCmp.cpp
@@ -666,7 +666,7 @@ int main(int argc, char* argv[])
     std::time_t logTime = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
     std::string outDir = "out/";// + logTime;
 
-    if (argc < 1)
+    if (argc <= 1)
     {
         printf("Please pass the data log directory path (i.e. \"C:\\Users\\[username]\\Documents\\Inertial Sense\\Logs\\20180716_172323)\"\r\n");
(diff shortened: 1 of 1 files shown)


Hephaestus-Maximus (approved):
Small, correct fix — verified rather than just read.

The bug is real and the root cause is exactly what the fix implies: `argc` is guaranteed by the C++ standard to be ≥ 1 (`argv[0]` is always the program name), so the old `if (argc < 1)` check could never be true — it was dead code, meaning "no path argument supplied" (`argc == 1`) fell straight through to `logger.LoadFromDirectory(argv[1], ...)` with `argv[1]` null, matching the reported `std::logic_error: basic_string: construction from null is not valid` crash exactly.

Confirmed `argv[1]` is only used once in this function, immediately after the check, so `argc <= 1` (equivalent to `argc < 2`) is sufficient to guard it — no other unguarded access path exists that would need the same fix.

Waited for CI to actually complete rather than trusting the diff alone: both Linux and Windows builds pass on the head commit.

Approving.

kylemallory (approved):

tmcd249 (approved):

PR #1247, opened 2026-07-28
SN-8386 Fix ImxDeviceFactory rejecting IMX-6 devices

jbergstrom88:
## Summary
`ImxDeviceFactory::allocateDevice()` in `src/DeviceFactory.h` only accepted an
exact match against `IS_HARDWARE_IMX_5_0`, silently rejecting IMX-6 (and any
future IMX variant) even after the device fully passes serial handshake,
`stepValidation`, and `completeValidation`. No error/warning is logged at the
rejection point, so from the caller's side (ROS2 node, cltool, etc.) this
looks exactly like a generic connection timeout — very difficult to diagnose.

## Fix
Also accept `IS_HARDWARE_IMX_6_0` in the hardware ID comparison. Updated the
surrounding doc comments to match.

## Testing
Confirmed against real IMX-6 hardware (SN 942756674, fw 3.1.0.71) — connected
successfully and streamed live sensor data (IMU, GPS, mag, baro, INS outputs)
via `inertial_sense_ros2_node` after this fix, where it previously failed
100% of the time with "Unable to open serial port".

## Follow-up consideration (not in this PR)
Longer-term, matching on hardware *type* (`IS_HARDWARE_TYPE_IMX`) rather than
an exact type+version encoding would prevent this same silent-rejection
pattern for future IMX revisions. Flagging as a possible follow-up rather
than bundling into this minimal fix.
---
### Review addendum (Hephaestus) — description completeness: 3/5

This description (and the class/method Doxygen comments in the diff) describe the *originally-planned* fix — an exact match against `IS_HARDWARE_IMX_5_0`/`IS_HARDWARE_IMX_6_0`, with type-based matching explicitly deferred as a "follow-up consideration, not in this PR." The code that actually shipped in the second commit does the opposite: it uses `DEV_INFO_MATCHES_HDW_ID(devInfo, IS_HARDWARE_IMX)`, a type-only wildcard match, immediately — the "follow-up" is already implemented here, just not described as such. See the review for a real regression this wildcard approach introduces (uINS/EVB devices also match), independent of this description gap.


src/DeviceFactory.h
```diff
@@ -191,13 +191,13 @@ class ImxDeviceFactory : public DeviceFactory {
...
-     * @brief Implements DeviceFactory::allocateDevice(); allocates an ISDevice only if devInfo resolves to an IMX-5 hardware Id.
+     * @brief Implements DeviceFactory::allocateDevice(); allocates an ISDevice only if devInfo resolves to an IMX-5 or IMX-6 hardware Id.
      * @param devInfo the device information uniquely identifying the specific device.
      * @param port an associated port (optional) that this device should be bound to.
-     * @return a new ISDevice if devInfo is an IMX-5, otherwise nullptr.
+     * @return a new ISDevice if devInfo is an IMX-5 or IMX-6, otherwise nullptr.
      */
     device_handle_t allocateDevice(const dev_info_t &devInfo, port_handle_t port) override {
-        if (ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_5_0)
+        if (ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_5_0 || ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_6_0)
```
kylemallory: Good catch - but actually, there is a more correct way to do this. the "encoded HDW_ID" actually is a bit mask of type and version.  So:

`((ENCODE_DEV_INFO_HDW_ID(devInfo) & IS_HARDWARE_IMX) == IS_HARDWARE_IMX)`

should cover both bases (and future IMXs) with a single test

kylemallory (changes requested):
Thanks for catching this and making the fix, @jbergstrom88.  However, if you're going to continue to submit PRs (which is encouraged), I'll need you to follow our reporting/git best-practices, which are as follows:
- No engineering work without a SN-* Jira ticket which describes the defect of feature.  If this is tied to an ISCS- card, then create the SN-* card and cross-link back to the support issue (we can keep a full chain of the history of the issue).
- Branch names should contain `SN-*_...` as the start of the branch name ("fix/SN-*..." or "feature/SN-*_..." is also fine (and encouraged).  This allows Jira to track the status of the branch/PR by associated the branch/commits with the Jira ticket.
- The PR title should also template "SN-* - <description>" for the same reasons as the branch.

src/DeviceFactory.h
jbergstrom88: Thank you. The code has been changed.

kylemallory (approved):

src/DeviceFactory.h
```diff
@@ -191,13 +191,13 @@ class ImxDeviceFactory : public DeviceFactory {
...
-     * @brief Implements DeviceFactory::allocateDevice(); allocates an ISDevice only if devInfo resolves to an IMX-5 hardware Id.
+     * @brief Implements DeviceFactory::allocateDevice(); allocates an ISDevice only if devInfo resolves to an IMX-5 or IMX-6 hardware Id.
      * @param devInfo the device information uniquely identifying the specific device.
      * @param port an associated port (optional) that this device should be bound to.
-     * @return a new ISDevice if devInfo is an IMX-5, otherwise nullptr.
+     * @return a new ISDevice if devInfo is an IMX-5 or IMX-6, otherwise nullptr.
      */
     device_handle_t allocateDevice(const dev_info_t &devInfo, port_handle_t port) override {
-        if (ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_5_0)
+        if (DEV_INFO_MATCHES_HDW_ID(devInfo, IS_HARDWARE_IMX))
```
Hephaestus-Maximus: **This introduces a real regression: `DEV_INFO_MATCHES_HDW_ID(devInfo, IS_HARDWARE_IMX)` also matches uINS and EVB devices, not just IMX.**

I wrote a small standalone program against `data_sets.h` on this branch to check every hardware type against this exact call, and confirmed it empirically:
```
IMX-5.0              matches IS_HARDWARE_IMX=1  matches IMX_5_0=1  matches IMX_6_0=0
IMX-6.0              matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=1
IMX-7.0 (future)      matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=0
GPX-1.0               matches IS_HARDWARE_IMX=0  matches IMX_5_0=0  matches IMX_6_0=0
uINS-3.2              matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=0
EVB-2.0               matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=1
```
`uINS-3.2` and `EVB-2.0` both match `IS_HARDWARE_IMX` — meaning `ImxDeviceFactory::allocateDevice()` will now `make_shared<ISDevice>(...)` for a uINS or EVB device, treating it as an IMX. `EVB-2.0` even matches `IS_HARDWARE_IMX_6_0` specifically.

**Root cause**: `DEV_INFO_MATCHES_HDW_ID(di, hdwId)` = `(ENCODE_DEV_INFO_TO_HDW_ID(di) & hdwId) == ENCODE_DEV_INFO_TO_HDW_ID(di)` is a "is `di`'s full 16-bit encoding a bitwise subset of `hdwId`" check, not a proper masked-field comparison. `IS_HARDWARE_IMX = ENCODE_HDW_ID(IS_HARDWARE_TYPE_IMX, -1, -1) = 0x0FFF` (type field `0b000011` OR'd with an all-`1`s major/minor wildcard sets every bit in the low 12 bits). Since `IS_HARDWARE_TYPE_UINS = 1 = 0b000001` and `IS_HARDWARE_TYPE_EVB = 2 = 0b000010` are both *bitwise subsets* of type `3 = 0b000011` (IMX), any uINS/EVB `devInfo` — regardless of its actual major/minor — passes the subset check too. `IS_HARDWARE_TYPE_GPX = 4 = 0b000100` isn't a subset of `0b000011`, so GPX is correctly excluded — which is presumably why this wasn't caught: GPX (the other real `DeviceFactory` in this codebase) happens not to trigger it, but uINS/EVB do.

This SDK has no dedicated uINS/EVB `DeviceFactory` (only `ImxDeviceFactory`/`GpxDeviceFactory` exist), so before this PR a uINS/EVB device would correctly go unclaimed by every factory; after this PR, `ImxDeviceFactory` wrongly claims it and instantiates an IMX-flavored `ISDevice` for hardware that isn't one.

**Suggested fix**: use a direct field comparison instead of the bitmask-subset macro, e.g. `devInfo.hardwareType == IS_HARDWARE_TYPE_IMX` — that gets the "any IMX version" behavior the commit message describes without the subset-check pitfall, since it's a plain equality on a single field rather than an all-bits-OR'd mask. Alternatively, reverting to the original two-way exact match (`ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_5_0 || ... == IS_HARDWARE_IMX_6_0`) — i.e. what the PR description actually says this PR does — would also be safe, just narrower (matches the currently-described "minimal fix" scope).

Hephaestus-Maximus (dismissed):
## Review: Fix ImxDeviceFactory rejecting IMX-6 devices (SN-8386)

Moderate-tier review (small diff, but it's the device-discovery/allocation gate for every IMX device this SDK connects to). Built the SDK, ran the full test suite, and wrote a small standalone program against the actual branch headers to empirically check the new hardware-matching logic across every current hardware type — not just the two the PR description names.

### What's correct
- Root-cause diagnosis of the original bug is accurate: `ImxDeviceFactory::allocateDevice()` did require an exact `== IS_HARDWARE_IMX_5_0`, silently rejecting IMX-6 even after full validation succeeded.
- Confirmed against real hardware per the description (IMX-6, SN 942756674) — not something I can re-verify without the device, but no reason to doubt it, and it explains the reported symptom precisely.
- Build + full test suite pass cleanly on this branch (459 tests, 455 passed, 4 pre-existing skips — identical to baseline).

### Blocking finding — real regression, verified empirically
The code that actually shipped (second commit, "Use `DEV_INFO_MATCHES_HDW_ID` wildcard check...") doesn't match the PR description. Instead of the described exact match against the two named hardware IDs, it does:
```cpp
if (DEV_INFO_MATCHES_HDW_ID(devInfo, IS_HARDWARE_IMX))
    return std::make_shared<ISDevice>(devInfo, port);
```
I compiled a small program against `data_sets.h` on this branch and checked every current `eIsHardwareType` against this exact call:
```
IMX-5.0            matches IS_HARDWARE_IMX=1  matches IMX_5_0=1  matches IMX_6_0=0
IMX-6.0            matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=1
IMX-7.0 (future)   matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=0
GPX-1.0            matches IS_HARDWARE_IMX=0  matches IMX_5_0=0  matches IMX_6_0=0
uINS-3.2           matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=0
EVB-2.0            matches IS_HARDWARE_IMX=1  matches IMX_5_0=0  matches IMX_6_0=1
```
**`uINS-3.2` and `EVB-2.0` both incorrectly match `IS_HARDWARE_IMX`** — `EVB-2.0` even matches `IS_HARDWARE_IMX_6_0` specifically. `ImxDeviceFactory` will now claim and instantiate an `ISDevice` for a uINS or EVB device, which it has no business doing.

Root cause: `DEV_INFO_MATCHES_HDW_ID` is a bitwise-subset check (`(full & hdwId) == full`), not a proper masked-field comparison. `IS_HARDWARE_IMX = ENCODE_HDW_ID(IS_HARDWARE_TYPE_IMX, -1, -1)` sets every bit in the low 12 bits (type=`3`=`0b000011` OR'd with an all-1s major/minor wildcard). `IS_HARDWARE_TYPE_UINS` (`1`=`0b000001`) and `IS_HARDWARE_TYPE_EVB` (`2`=`0b000010`) are both bitwise subsets of `3`, so they pass the check regardless of major/minor. `IS_HARDWARE_TYPE_GPX` (`4`=`0b000100`) isn't a subset, which is presumably why this wasn't noticed — the only other real factory in this codebase (`GpxDeviceFactory`) doesn't trigger it, but uINS/EVB do. This SDK has no dedicated uINS/EVB factory, so today those devices correctly go unclaimed by every factory; after this PR they get wrongly claimed by `ImxDeviceFactory` instead.

**Suggested fix**: compare the hardware-type field directly — `devInfo.hardwareType == IS_HARDWARE_TYPE_IMX` — which gets the "match any IMX version" behavior intended here without the bitmask-subset pitfall (it's a plain single-field equality, not an OR'd-bits mask). The originally-described minimal fix (exact match against `IS_HARDWARE_IMX_5_0` / `IS_HARDWARE_IMX_6_0`) would also be safe, just narrower. Full trace posted inline on the changed line.

### Non-blocking
The PR description and the Doxygen comments in this diff both describe the exact-match approach and explicitly defer type-based matching as a future follow-up — but the shipped code already does type-based matching. Appended an addendum to the description noting the drift; worth updating the doc comments too once the fix above lands, so they describe what the code actually checks.

**Suggestion**: no test exists for `ImxDeviceFactory::allocateDevice()` (or `GpxDeviceFactory`'s equivalent). A small table-driven test covering IMX-5, IMX-6, a hypothetical future IMX version, and explicit rejection of GPX/uINS/EVB would have caught this immediately and would guard against regressing it again.

### Verdict
Requesting changes on the regression above — @kylemallory's existing approval predates this specific check (his review comments were about Jira/branch-naming process, not this code path), so flagging directly in case it slipped past. Everything else here — the original bug diagnosis, the build, and the test suite — is solid.

tmcd249 (approved):

waltjohnson (approved):

src/DeviceFactory.h
Hephaestus-Maximus: For the record: the macro-based approach you suggested here (`(ENCODE_DEV_INFO_TO_HDW_ID(devInfo) & IS_HARDWARE_IMX) == IS_HARDWARE_IMX`) is what actually shipped in a later commit, and it's what triggered the uINS/EVB regression flagged below (the macro did a flat bitwise-subset check, not a proper per-field wildcard match). Fixed the macro itself in 88011415 so this pattern is now safe to use going forward -- it correctly wildcards only within each field (type/major/minor) instead of across the whole packed value.

src/DeviceFactory.h
Hephaestus-Maximus: Fixed the root cause directly in the macro (commit 88011415) rather than just working around it here -- DEV_INFO_MATCHES_HDW_ID now does a per-field wildcard-aware comparison instead of a flat bitwise-subset check across the whole packed ID, so uINS/EVB no longer match an IMX wildcard regardless of the call site. Added table-driven tests in tests/test_utils.cpp covering this exact regression. The direct devInfo.hardwareType comparison already in place here remains correct and unchanged.

src/DeviceFactory.h
```diff
@@ -191,13 +191,13 @@ class ImxDeviceFactory : public DeviceFactory {
     // ~ImxDeviceFactory() override = default;
 
     /**
-     * @brief Implements DeviceFactory::allocateDevice(); allocates an ISDevice only if devInfo resolves to an IMX-5 hardware Id.
+     * @brief Implements DeviceFactory::allocateDevice(); allocates an ISDevice only if devInfo resolves to any IMX hardware type (IMX-5, IMX-6, or future IMX revisions).
```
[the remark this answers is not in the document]
Hephaestus-Maximus: Reworded in f1d74487 to describe what the code actually checks (devInfo.hardwareType == IS_HARDWARE_TYPE_IMX) instead of promising forward compatibility with hardware types that don't exist yet.

src/DeviceFactory.h
```diff
@@ -191,13 +191,13 @@ class ImxDeviceFactory : public DeviceFactory {
...
      * @param port an associated port (optional) that this device should be bound to.
-     * @return a new ISDevice if devInfo is an IMX-5, otherwise nullptr.
+     * @return a new ISDevice if devInfo's hardware type is IMX, otherwise nullptr.
      */
     device_handle_t allocateDevice(const dev_info_t &devInfo, port_handle_t port) override {
-        if (ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_5_0)
+        if (devInfo.hardwareType == IS_HARDWARE_TYPE_IMX)
             return std::make_shared<ISDevice>(devInfo, port);
 
         return nullptr;
```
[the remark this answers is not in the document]
Hephaestus-Maximus: Added braces in f1d74487.

Hephaestus-Maximus (commented):
## Follow-up on my earlier review

The regression I flagged earlier (`DEV_INFO_MATCHES_HDW_ID(devInfo, IS_HARDWARE_IMX)` wrongly matching uINS/EVB) is fixed at the root now:

- `88011415` — fixed `DEV_INFO_MATCHES_HDW_ID` itself in `src/data_sets.h`. It was doing a flat bitwise-subset check across the whole packed hardware ID; reworked it to compare each field (type/major/minor) independently, only wildcarding a field when it's all-1s within its own width. Added table-driven tests in `tests/test_utils.cpp` covering the exact regression scenario, exact-match behavior, and the `IS_HARDWARE_ANY` full wildcard.
- `f1d74487` — addressed both Copilot comments: reworded the `allocateDevice()` doc comment to describe what the code actually checks rather than promising forward compatibility, and added braces around the `if` body.

Full `IS-SDK_unit-tests` suite passes (469/473, 4 pre-existing skips, 0 failures) after both commits. The `devInfo.hardwareType == IS_HARDWARE_TYPE_IMX` comparison already in `ImxDeviceFactory::allocateDevice()` remains correct and unchanged — this PR doesn't need to switch back to the macro to benefit from the fix, though it (and any other current/future caller) is now safe to use it.

Marking my blocking finding resolved. I'm not the merge decision-maker here — leaving that to a human reviewer — but wanted to reflect that the concern I raised is addressed.

kylemallory (approved):

PR #1253, opened 2026-07-29
SN-8390 Fix NTRIP connect + reconnect with non-blocking TCP

kylemallory:
## Summary

Fixes NTRIP corrections failing on 3.1.0-rc. The caster is healthy — the failures are SDK regressions/gaps in the NTRIP connect path around the non-blocking-connect change. The non-blocking connect itself is **retained**.

Jira: [SN-8390](https://inertialsense.atlassian.net/browse/SN-8390)

Two related bugs (the second found during discovery of the first):

### Bug 1 — initial connect writes before the socket is connected

`SN-8274` (`72b9c298`) made `tcpPortOpen()` use a **non-blocking** connect (good — avoids stalling the caller for the full kernel SYN timeout). `portOpen()` now returns `PORT_ERROR__NONE` while the handshake is still in flight ("pending"), **without** setting `PORT_FLAG__OPENED`; the caller must poll until `portIsOpened()`.

`NtripCorrectionService::connect()` called `portOpen()` once, saw `NONE`, and immediately `portWrite()` the NTRIP GET → `EAGAIN` → the reported `Error submitting NTRIP connection request`. `portOpenRetry()` had the same latent bug (returned on the pending `0`).

**Fix:**
- `base_port.c portOpenRetry()`: keep polling until `portIsOpened()` instead of returning on the first `PORT_ERROR__NONE`; only a negative result is a hard failure. (No existing callers; synchronous/blocking opens still return immediately via a post-`portOpen()` `portIsOpened()` check.)
- `NtripCorrectionService::connect()`: use `portOpenRetry()` (5 s) so the TCP connection is established before the GET is written.

### Bug 2 — corrections never resume after a reconnect

`CorrectionService::step()` reopens the source TCP socket after a drop but never re-sends the NTRIP mount-point request. Since an NTRIP caster only streams after it receives that request, a reconnected socket sits connected-but-silent.

**Fix:**
- Make `CorrectionService::step()` `virtual` (and `lastConnAttemptTs` protected) so protocol subclasses can extend (re)connect handling.
- Factor the mount-point negotiation out of `connect()` into `negotiateMountPoint()`, and cache the built request (`requestMsg`) for replay.
- Override `NtripCorrectionService::step()`: when the source isn't open, advance the non-blocking TCP connect and, once the socket reopens, re-run `negotiateMountPoint()` before delegating to `CorrectionService::step()`. Reconnect stays non-blocking (portOpen polled across `step()` calls); only the bounded mount-point negotiation blocks briefly, once per reconnect.

`tcpPortOpen()`'s non-blocking connect is intentionally unchanged.

## Validation

Against the **live caster** with SDK harnesses using the real `TcpPortFactory` / `portOpen` / `portWrite` / `NtripCorrectionService`:

- Caster healthy: sourcetable `SOURCETABLE 200 OK`; mountpoint `GET /GNSS-VRS-NAD83-RTCM32` → `ICY 200 OK`.
- **Bug 1 before:** single `portOpen()` → `portIsOpened()`=0, immediate `portWrite()` = `-11` (`-EAGAIN`) → fails. `portOpenRetry()` returns 0 with `portIsOpened()`=0.
- **Bug 1 after:** `portOpenRetry()` returns 0 with `portIsOpened()`=1; the exact new `connect()` sequence writes all bytes and gets `ICY 200 OK`.
- **Bug 2 after:** construct `NtripCorrectionService` (connects, `ICY 200 OK`); force-close the source socket (`isConnected()`→0); pump `step()` → reconnects and re-negotiates within ~3 iterations (`ICY 200 OK` again), `isConnected()`→1.

Builds clean on Linux/GCC (SDK lib + full EvalTool). MSVC via CI (the `WSAEWOULDBLOCK`/`WSAEINPROGRESS` pending branch in `tcpPortOpen()` mirrors the POSIX path, so `portOpenRetry()` polls identically on Windows).

[SN-8390]: https://inertialsense.atlassian.net/browse/SN-8390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/CorrectionService.h
+++ b/src/CorrectionService.h
@@ -179,10 +179,13 @@ public:
     void removeRTCM3PacketListeners(uint32_t id);
 
     /**
-     * Checks the source port for data and forwards it to all devices
+     * Checks the source port for data and forwards it to all devices.
+     * Also (re)opens the source port if it is not currently open. Virtual so that protocol-specific
+     * subclasses (e.g. NtripCorrectionService) can extend the (re)connection handling -- NTRIP, for
+     * example, must re-negotiate its mount-point request after every (re)connect, not just reopen the socket.
      * @return 0 if no packets processed, positive number representing number of packets processed, negative number representing errno
      */
-    int step();
+    virtual int step();
 
     /**
      * @return the age, in milliseconds, of the last received RTCM3 message.
@@ -209,6 +212,7 @@ public:
 protected:
     port_handle_t source {};                  //!< The bound source port from which correction data is read
     std::vector<port_handle_t> ports;          //!< Downstream ports to which correction data is forwarded
+    uint32_t lastConnAttemptTs = 0;            //!< timestamp (ms) after which the next source (re)connection attempt is allowed (reconnect backoff)
 
 private:
     inline static const std::vector<PortFactory*>& nullFactories = {};    //!< Empty factory list, used as the default argument for the portName-based constructor
@@ -217,7 +221,6 @@ private:
     is_comm_instance_t packetParser = {};                                 //!< SDK comm-protocol parser instance bound to the source port
     uint32_t rtcm3PacketsProcessed = 0;                                 //!< total number of RTCM3 packets that have been processed
     uint32_t rtcm3PacketLastMs = 0;                                     //!< timestamp in ms, since the last RTCM3 packet was seen
-    uint32_t lastConnAttemptTs = 0;                                     //!< timestamp (ms) of the last (re)connection attempt to the source
     bool localSrcPort = false;                                          //!< true if the source port was locally instantiated rather than passed in the constructor
     PortFactory* srcPortFactory = nullptr;                               //!< the factory used to create the source port (if localSrcPort is true)
 
--- a/src/NtripCorrectionService.cpp
+++ b/src/NtripCorrectionService.cpp
@@ -12,6 +12,9 @@
 #include "TcpPortFactory.h"
 #include "util/util.h"
 
+// Max time to wait for the (non-blocking) TCP connection to the caster to complete before giving up.
+#define NTRIP_CONNECT_TIMEOUT_MS 5000
+
 bool NtripCorrectionService::connect(const std::string& connectUrl, std::string userAgent) {
     // parse the URL; NTRIP defaults to port 2101 when the caster URL omits one
     const utils::UriParts uri = utils::parseUri(connectUrl, "ntrip://:2101");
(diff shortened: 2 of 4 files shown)


tmcd249 (approved):

aabogdanov (approved):

Hephaestus-Maximus (approved):
## Review: NTRIP non-blocking connect/reconnect fix (SN-8390)

Did an independent line-by-line trace of the diff against the pre-existing base-class/port behavior (not just reading the patch in isolation):

- **`base_port.c: portOpenRetry()`** — the success check now correctly uses `portIsOpened(port)` instead of `lastResult == PORT_ERROR__NONE`. This matters because a non-blocking `tcpPort` connect returns `PORT_ERROR__NONE` while the handshake is still in flight, without ever setting `PORT_FLAG__OPENED` — the old check would report success before the socket was actually usable. The pre-existing retry-on-negative-result loop structure is otherwise unchanged.
  - Minor (non-blocking) observation: if `portOpen()` returns `0` on every poll all the way to timeout (perpetually "still pending," never negative, never opened), `lastResult` stays `0` and the final `return portIsOpened(port) ? PORT_ERROR__NONE : lastResult;` degenerates to returning `PORT_ERROR__NONE` even though the port never actually opened. Traced the only caller (`negotiateMountPoint()`) and confirmed it has its own `portIsOpened(source)` guard immediately after, so this can't actually produce a false-positive connection — just a mildly misleading return value in a very narrow edge case. Worth a one-line comment for future maintainers but not blocking.

- **`NtripCorrectionService::connect()`** — replacing the one-shot `portOpen()` check with `portOpenRetry(source, NTRIP_CONNECT_TIMEOUT_MS, 2)` is the correct fix for the root cause (non-blocking connect returning "success" before the handshake completes). The 2ms poll interval matches the ticket's own empirical validation (~12ms observed for 6 polls against a live caster).

- **`negotiateMountPoint()`** — clean extraction of the existing request/response logic into its own method; traced through and confirmed no logic was dropped or reordered.

- **New `step()` override** — implements the reconnect/renegotiate state machine (single non-blocking `portOpen()` poll per call, backoff only on hard negative failure, immediate renegotiation attempt once `portIsOpened()`, `portClose()` + backoff if renegotiation fails). Verified against the unmodified base `CorrectionService::step()` (`src/CorrectionService.cpp:96-97`) that this mirrors the same `lastConnAttemptTs`-as-next-allowed-time convention already established there — no semantic drift between base and derived class. `portClose()` confirmed as an established, widely used API elsewhere in the codebase (`DeviceManager.cpp`, `ISBootloaderBase.cpp`).

- **`CorrectionService.h`** — `lastConnAttemptTs` moved `private`→`protected` and `step()` made `virtual` purely to support the derived-class override; doc comment update accurately describes existing semantics.

This also folds in the second bug noted in the Jira ticket (reconnect never re-sent the NTRIP mount-point request) — the new `step()` correctly re-runs `negotiateMountPoint()` on every successful reconnect, not just the initial connect.

CI: both Linux and Windows are now green (Windows was pending at time of initial review, confirmed passing on recheck). Two prior human approvals already in place.

No blocking issues found. Approving.

PR #1254, opened 2026-07-29
SN-8269 Example Project fix for Minimal_ISDevice weak pointer run problem

shefft:
The develop branch Minimal_ISDevice project did not run (whether given a command line argument with a port name of an IS device channel or not), with error
```
terminate called after throwing an instance of 'std::bad_weak_ptr'
  what():  bad_weak_ptr
Aborted (core dumped)

``` 

Was found to come from using a raw ISDevice pointer instead of shared, the latter necessitated by updates to SDK ISDevice class, and some Example Projects needed to pick up the appropriate new usage.

This change was tested, and the project now runs rather than throwing the bad_weak_ptr exception.  This is the exact change to the two lines, taken from the `Airgap` branch commit: https://github.com/inertialsense/inertial-sense-sdk/commit/e88743903e944e4ff4b33edcb4bd343f33f80246, just the changes associated with this ISDevice* only.  Kyle and I determined to go ahead and get this small particular change done quicker independently.

### DIFF
--- a/ExampleProjects/Minimal_ISDevice/ISDeviceExample1.cpp
+++ b/ExampleProjects/Minimal_ISDevice/ISDeviceExample1.cpp
@@ -65,8 +65,11 @@ int main_explained(const char* portStr) {
     }
 
     // With a valid, opened port, we can instance an ISDevice - in this case, an IMX-5.0 and associate the port to it.
-    ISDevice* device = new ISDevice(IS_HARDWARE_IMX_5_0, port);
-
+    // NOTE: ISDevice derives from std::enable_shared_from_this, and several of its methods (connect/validate/step)
+    // call shared_from_this() internally. It MUST therefore be owned by a std::shared_ptr - allocating it with a raw
+    // 'new' would leave it without a control block and throw std::bad_weak_ptr at the first such call.
+    std::shared_ptr<ISDevice> device = std::make_shared<ISDevice>(IS_HARDWARE_IMX_5_0, port);
+    
     if (!device->isConnected())
         exit(3); // this is another way we can confirm the connected status of the device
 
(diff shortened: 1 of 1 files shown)


Hephaestus-Maximus (approved):
## Review: Minimal_ISDevice weak_ptr fix (SN-8269)

Trivial-tier change (2 functionally-identical lines, well-explained, directly traceable) — verified by reasoning + independent code inspection rather than a full build/test cycle, per the review skill's right-sizing guidance.

**Root cause confirmed independently:**
- `ISDevice : public std::enable_shared_from_this<ISDevice>` (`src/ISDevice.h:55`).
- `shared_from_this()` is called inside `ISDevice::connect()` (`src/ISDevice.cpp:1745`) and `ISDevice::step()` (`src/ISDevice.cpp:118`) — both of which are on the hot path of *every* iteration of both `main_explained()` and `main_minimal()` in this example (the `while (portIsOpened(...)) { device->step(); ... }` loop, plus the initial `connect()`/`validate()` call). An object constructed with a raw `new` (no shared_ptr control block) throws `std::bad_weak_ptr` the moment `shared_from_this()` is invoked on it — exactly the reported crash.

**Fix verified correct and consistent with the rest of the codebase:**
- `std::make_shared<ISDevice>(...)` is the established construction pattern used everywhere else `ISDevice` is instantiated: `src/DeviceManager.cpp:103`, `src/DeviceFactory.h:62,201,231`, `src/DeviceFactory.cpp:98`. This example project was simply out of sync with that convention.
- Confirmed the two changed lines are an *exact* match to the referenced Airgap-branch commit (`e8874390`), isolating just the pointer-type fix from that larger unrelated commit, as the description claims.
- Checked the rest of the file for fallout from the type change (explicit `delete device`, passing `device` as a raw `ISDevice*` elsewhere): none — `device` is only ever used via `->` member calls in both functions, which work identically for a `shared_ptr`.
- The trailing "no newline at end of file" fix is harmless and unrelated to the substantive change.

**CI:** Linux + Windows both pass.

**Description:** 5/5 — clearly states the symptom, root cause, exact provenance of the fix, and confirms it was tested and now runs correctly. No edit needed.

No issues found. Approving.

tmcd249 (approved):

kylemallory (approved):

PR #1256, opened 2026-07-30
SN-8402: Redefine GPX RX-traffic status bits as detected (breaking)

tmcd249:
## Summary — REVISED per Kyle's correction

Earlier revisions of this PR used a display-layer `boolInvert` flag to translate the negatively-worded bits for the ribbon chart, while leaving the underlying `GPX_STATUS_COM0/1/2/USB_RX_TRAFFIC_NOT_DETECTED` bit definition untouched. Kyle's correction: that's the wrong shape of fix — the underlying bit definition should carry the correct meaning; the rendering should just reflect it, not translate it.

## This version

**BREAKING:** renamed the four enum constants in `data_sets.h` (same mask, bits 4-7 of `DID_GPX_STATUS.status`, **inverted meaning**):
```
GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED -> GPX_STATUS_COM0_RX_TRAFFIC_DETECTED
GPX_STATUS_COM1_RX_TRAFFIC_NOT_DETECTED -> GPX_STATUS_COM1_RX_TRAFFIC_DETECTED
GPX_STATUS_COM2_RX_TRAFFIC_NOT_DETECTED -> GPX_STATUS_COM2_RX_TRAFFIC_DETECTED
GPX_STATUS_USB_RX_TRAFFIC_NOT_DETECTED  -> GPX_STATUS_USB_RX_TRAFFIC_DETECTED
```
The bit is now **1 when traffic HAS been heard in the last 30s**, previously 1 when it had NOT.

- `ISStatusDecode.{h,cpp}`: removed the `boolInvert` schema field (dead once the bit itself is correct); `buildGpxStatusDecode()` uses the new names via the existing `gerr()` helper — no special-casing needed. `isError` stays `false` (status, not fault). `legacyText` flipped to positive phrasing ("...RX traffic detected...").
- `data_sets.h`: also fixed the `eGpxStatus` block doc comment, which still said "per-port Rx-traffic-not-detected flags" (caught by Copilot review).
- `python/inertialsense/logInspector/logPlotter.py`: relabeled the four GPX-status plot rows to match (same raw bit; meaning flips).
- `tests/test_ISStatusDecode.cpp`: updated the legacy-oracle text to the new polarity; replaced the `boolInvert`-specific test with one asserting the new positive-polarity, non-fault subfields directly.

**Compatibility note (raised in review):** this is an intentional, documented breaking change to the wire meaning of bits 4-7 of `DID_GPX_STATUS.status`. Two concrete consequences:
1. Any external tooling reading these raw bits directly (not through `ISDataMappings`/`ISStatusDecode`) will misinterpret them against firmware built from this change vs. any prior release.
2. **Historical logs recorded with pre-fix firmware, replayed through a post-fix build of this SDK** (cltool/EvalTool/Logalyzer), will show the new positive-polarity label/legend applied to the old bit semantics — i.e. an old log where the bit meant "not detected" will now display as if it meant "detected". There's no data-format flag distinguishing pre- and post-fix firmware for this field, so this is a real, accepted limitation, not an oversight — flagging it explicitly rather than leaving it implicit.

## Companion PRs

- `inertialsense/imx#1643` — the actual firmware fix (where the bit is set).
- `inertialsense/is-gpx#178` — submodule bump so shipped GPX-1 3.1.0 firmware picks it up.
- `inertialsense/logalyzer#84` — where the bug was first observed; now needs no special handling at all since the bit itself is correct.

## Test plan

- [x] Full SDK suite: 460 tests, 456 pass / 4 pre-existing skips, 0 failures.
- [x] `GpxStatus_RxTrafficSubfields_PositivePolarityNotFault` (new) — asserts name/kind/isError/legacyText for all four subfields.
- [x] Existing `GpxStatus_RoundTrip_*` (every-bit, fatal-resets, random-sweep) all still pass against the updated legacy oracle.

### DIFF
--- a/src/ISStatusDecode.cpp
+++ b/src/ISStatusDecode.cpp
@@ -27,7 +27,7 @@ uint32_t maskShift(uint32_t mask)
 }
 
 /** @brief Build a single on/off-bit sub-field descriptor. */
-status_subfield_t bitField(const char* name, uint32_t mask, bool isError, const char* legacy)
+status_subfield_t bitField(const char* name, uint32_t mask, bool isError, const char* legacy, bool boolInvert = false)
 {
     status_subfield_t s;
     s.name       = name;
@@ -35,6 +35,7 @@ status_subfield_t bitField(const char* name, uint32_t mask, bool isError, const 
     s.mask       = mask;
     s.shift      = 0;
     s.isError    = isError;
+    s.boolInvert = boolInvert;
     s.gateMask   = 0;
     s.legacyText = legacy;
     return s;
@@ -488,21 +489,21 @@ status_field_decode_t buildGpxStatusDecode()
     d.fieldName = "status";   // registry key is "gpxStatus"
     d.errorMask = (uint32_t)GPX_STATUS_GENERAL_FAULT_MASK;
 
-    auto gerr = [](const char* name, uint32_t mask, const char* legacy) {
-        return bitField(name, mask, (mask & (uint32_t)GPX_STATUS_GENERAL_FAULT_MASK) != 0, legacy);
-    };
-    // Inverted variant: bit set means "OK" (e.g. traffic detected), not fault.
-    auto gerrInv = [](const char* name, uint32_t mask, const char* legacy) {
-        return bitField(name, mask, (mask & (uint32_t)GPX_STATUS_GENERAL_FAULT_MASK) == 0, legacy);
+    auto gerr = [](const char* name, uint32_t mask, const char* legacy, bool boolInvert = false) {
+        return bitField(name, mask, (mask & (uint32_t)GPX_STATUS_GENERAL_FAULT_MASK) != 0, legacy, boolInvert);
     };
 
     // Parse-error count is rendered by the legacy code as a presence flag (any of the low nibble),
     // not a number — model it as a Bit on the count mask.
     d.subfields.push_back(gerr("COM parse errors", GPX_STATUS_COM_PARSE_ERR_COUNT_MASK, "0x0000000F - Communications parse error count"));
-    d.subfields.push_back(gerrInv("COM0 RX traffic detected", GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED, "0x00000010 - COM0 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerrInv("COM1 RX traffic detected", GPX_STATUS_COM1_RX_TRAFFIC_NOT_DETECTED, "0x00000020 - COM1 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerrInv("COM2 RX traffic detected", GPX_STATUS_COM2_RX_TRAFFIC_NOT_DETECTED, "0x00000040 - COM2 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerrInv("USB RX traffic detected", GPX_STATUS_USB_RX_TRAFFIC_NOT_DETECTED, "0x00000080 - USB RX traffic not detected in last 30 seconds."));
+    // These four are a status, not a fault (deliberately outside GPX_STATUS_GENERAL_FAULT_MASK, so
+    // isError stays false via gerr()) -- positively-worded name + boolInvert=true so a boolean-style
+    // consumer (the status-ribbon chart) shows "RX traffic detected" = true when the underlying
+    // NOT_DETECTED bit is clear. legacyText (and RenderStatusFromDecode's line output) is untouched.
+    d.subfields.push_back(gerr("COM0 RX traffic detected", GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED, "0x00000010 - COM0 RX traffic not detected in last 30 seconds.", /*boolInvert=*/true));
+    d.subfields.push_back(gerr("COM1 RX traffic detected", GPX_STATUS_COM1_RX_TRAFFIC_NOT_DETECTED, "0x00000020 - COM1 RX traffic not detected in last 30 seconds.", /*boolInvert=*/true));
+    d.subfields.push_back(gerr("COM2 RX traffic detected", GPX_STATUS_COM2_RX_TRAFFIC_NOT_DETECTED, "0x00000040 - COM2 RX traffic not detected in last 30 seconds.", /*boolInvert=*/true));
+    d.subfields.push_back(gerr("USB RX traffic detected", GPX_STATUS_USB_RX_TRAFFIC_NOT_DETECTED, "0x00000080 - USB RX traffic not detected in last 30 seconds.", /*boolInvert=*/true));
     d.subfields.push_back(gerr("Firmware image confirmed", GPX_STATUS_UPDATE_CONFIRMED, "0x00000100 - Update confirmed."));
     d.subfields.push_back(gerr("RTK buffer overflow", GPX_STATUS_FAULT_RTK_QUEUE_LIMITED, "0x00010000 - RTK buffer overflow."));
     d.subfields.push_back(gerr("GNSS receiver time fault", GPX_STATUS_FAULT_GNSS_RCVR_TIME, "0x00100000 - GNSS receiver time fault"));
--- a/src/ISStatusDecode.cpp
+++ b/src/ISStatusDecode.cpp
@@ -27,7 +27,7 @@ uint32_t maskShift(uint32_t mask)
 }
 
 /** @brief Build a single on/off-bit sub-field descriptor. */
-status_subfield_t bitField(const char* name, uint32_t mask, bool isError, const char* legacy, bool boolInvert = false)
+status_subfield_t bitField(const char* name, uint32_t mask, bool isError, const char* legacy)
 {
     status_subfield_t s;
     s.name       = name;
@@ -35,7 +35,6 @@ status_subfield_t bitField(const char* name, uint32_t mask, bool isError, const 
     s.mask       = mask;
     s.shift      = 0;
     s.isError    = isError;
-    s.boolInvert = boolInvert;
     s.gateMask   = 0;
     s.legacyText = legacy;
     return s;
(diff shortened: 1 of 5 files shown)


kylemallory (approved):

src/ISStatusDecode.cpp
```diff
@@ -491,14 +491,18 @@ status_field_decode_t buildGpxStatusDecode()
...
+    };
 
     // Parse-error count is rendered by the legacy code as a presence flag (any of the low nibble),
     // not a number — model it as a Bit on the count mask.
     d.subfields.push_back(gerr("COM parse errors", GPX_STATUS_COM_PARSE_ERR_COUNT_MASK, "0x0000000F - Communications parse error count"));
-    d.subfields.push_back(gerr("COM0 RX traffic lost", GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED, "0x00000010 - COM0 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerr("COM1 RX traffic lost", GPX_STATUS_COM1_RX_TRAFFIC_NOT_DETECTED, "0x00000020 - COM1 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerr("COM2 RX traffic lost", GPX_STATUS_COM2_RX_TRAFFIC_NOT_DETECTED, "0x00000040 - COM2 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerr("USB RX traffic lost", GPX_STATUS_USB_RX_TRAFFIC_NOT_DETECTED, "0x00000080 - USB RX traffic not detected in last 30 seconds."));
+    d.subfields.push_back(gerrInv("COM0 RX traffic detected", GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED, "0x00000010 - COM0 RX traffic not detected in last 30 seconds."));
```
Hephaestus-Maximus: **This label inversion looks backwards, and it recreates the exact class of bug this PR is fixing — just moved from the color to the title.**

I traced the actual consumer: Logalyzer's `StatusRibbonModel::resolveSegment()` (Bit case) does:
```cpp
if (sf.isError) {
    outLabel = sv ? "FAULT" : "OK";
    outColor = sv ? policy.error : policy.nominal;
}
```
where `sv = (raw & mask) ? 1 : 0`. For `GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED`, `sv == 1` means traffic was **NOT** detected (a real fault — that's literally the bit's name). With `isError` now `true` (the correct part of this fix), that state renders as a red **"FAULT"** segment. Good.

But `sf.name` is used verbatim as the **static row title** (`row.label = QString::fromStdString(sf.name);` in `StatusRibbonModel.cpp`), shown regardless of the current bit value. Renaming it from `"COM0 RX traffic lost"` to `"COM0 RX traffic detected"` means the ribbon row will read **"COM0 RX traffic detected"** while showing a red **FAULT** segment at the exact moment traffic is *not* detected — the title and the fault indicator directly contradict each other.

Every other `isError=true` bit row in this file names the *fault* condition (e.g. `"RTK buffer overflow"`, `"GNSS receiver time fault"` a few lines below) — that's the established convention this table follows, and it's what makes the row title consistent with the red "FAULT" label. I'd suggest reverting just the name back to `"COM0/1/2 RX traffic lost"` / `"USB RX traffic lost"` while keeping the `isError` fix (the actual, correct part of this change).

src/ISStatusDecode.cpp
```diff
@@ -491,14 +491,18 @@ status_field_decode_t buildGpxStatusDecode()
     auto gerr = [](const char* name, uint32_t mask, const char* legacy) {
         return bitField(name, mask, (mask & (uint32_t)GPX_STATUS_GENERAL_FAULT_MASK) != 0, legacy);
     };
+    // Inverted variant: bit set means "OK" (e.g. traffic detected), not fault.
+    auto gerrInv = [](const char* name, uint32_t mask, const char* legacy) {
```
Hephaestus-Maximus: Minor clarity nitpick on the comment/helper itself (non-blocking): this says "bit set means OK ... not fault," but for every mask this is actually called with (`GPX_STATUS_COM{0,1,2}_RX_TRAFFIC_NOT_DETECTED`, `GPX_STATUS_USB_RX_TRAFFIC_NOT_DETECTED`), bit set means the *opposite* — traffic was **not** detected, which is the fault condition. The resulting `isError=true` is actually correct (since none of these masks overlap `GPX_STATUS_GENERAL_FAULT_MASK`, `(mask & FAULT_MASK) == 0` is `true` for all of them) — but that's an accident of these specific masks always falling outside `FAULT_MASK`, not a general "inverted polarity" helper. If someone later reuses `gerrInv` for a mask that *does* overlap `FAULT_MASK` expecting the "bit-set-means-OK" semantics the comment describes, they'd get `isError=false` for what the comment implies should still map correctly — worth either fixing the comment to describe what this helper actually computes, or just inlining `bitField(name, mask, true, legacy)` directly at each call site instead of a generically-named-but-narrowly-correct helper.

Hephaestus-Maximus (changes requested):
## Review: GPX status labeling fix (SN-8402)

Moderate-tier review — self-contained logic change in one function, checked out and built the SDK unit-test suite (`tests/IS-SDK_unit-tests`) against the PR head, and cross-referenced the actual downstream consumer (`inertialsense/logalyzer`) since this table's whole purpose is to feed that consumer.

### What's correct
- Root-cause diagnosis is right: `gerr()` derives `isError` from `(mask & GPX_STATUS_GENERAL_FAULT_MASK) != 0`, and the four RX-traffic-not-detected bits (`0x10`/`0x20`/`0x40`/`0x80`) fall outside `GPX_STATUS_GENERAL_FAULT_MASK` (`0xFFFF0000`), so they were always classified `isError=false` despite being genuine comms-loss faults.
- Traced the actual runtime effect in `logalyzer/src/charts/StatusRibbonModel.cpp::resolveSegment()`: for a Bit sub-field, `isError=false` makes the segment render as neutral `"TRUE"/"FALSE"` in `policy.nominal`/`policy.off` colors — i.e. losing RX traffic was previously never shown in the ribbon's error/red state. Flipping `isError` to `true` (via the new `gerrInv` helper) fixes this: segments now render `"FAULT"`/`"OK"` in `policy.error`/`policy.nominal`, which is the correct alerting behavior.
- Ran the existing `ISStatusDecode.*` test suite (40 tests) against the PR head — all pass (round-trip fidelity is preserved since `legacyText` output is unchanged).
- CI: Windows passes. Linux shows one failure (`TruncatedLogTest.MidPacketTruncation_StopsAtLastValidRecord`) — verified this is unrelated to this diff (touches log-truncation code, not `ISStatusDecode.cpp`) and passed 3/3 on a local re-run against this exact branch, so it's a pre-existing flake, not something this PR introduced.

### Blocking finding
This PR also renames the four sub-fields from `"COM0/1/2 RX traffic lost"` / `"USB RX traffic lost"` to `"... RX traffic detected"`. That name is used verbatim as the **static row title** in Logalyzer's ribbon chart (`row.label = QString::fromStdString(sf.name)`), shown regardless of the bit's current value — it's not conditioned on set/clear. Since the bit (`GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED`, etc.) is 1 precisely when traffic is **not** detected, the ribbon will now show a row titled **"COM0 RX traffic detected"** displaying a red **"FAULT"** segment at the exact moment traffic is lost — title and fault indicator directly contradict each other. Every other `isError=true` bit row in this same table names the fault condition itself (`"RTK buffer overflow"`, `"GNSS receiver time fault"`, ...), so this also breaks the table's own established naming convention. See inline comment for the full trace. Suggested fix: keep the `isError` change, revert just the `name` strings back to the "lost" framing.

### Nitpick (non-blocking)
The `gerrInv` helper's doc comment ("bit set means OK ... not fault") describes the opposite of what these specific bits mean, and only produces the correct `isError=true` here because none of the masks it's called with overlap `GPX_STATUS_GENERAL_FAULT_MASK` — see inline comment for detail. Worth tightening the comment or just inlining `bitField(name, mask, true, legacy)` so a future reuse of `gerrInv` on a different mask doesn't get surprised.

### Suggestion
No test was added for this fix, despite this same file having a dense existing pattern for exactly this (e.g. `EXPECT_TRUE(sf.isError) << sf.name;` in `test_ISStatusDecode.cpp`). A targeted assertion on `isError` (and ideally `name`, given the finding above) for these four sub-fields would have caught the title/fault contradiction before it shipped.

### Description
Graded 1/5 (a single line restating the commit title, no root cause or downstream-impact explanation) — replaced with a summary derived from the diff + `logalyzer` cross-reference; original preserved in a collapsed footer.

Requesting changes on the row-title inversion; everything else here is solid.

waltjohnson (approved):

src/ISStatusDecode.cpp
shefft: If this is true, seems worth addressing.

src/ISStatusDecode.cpp
shefft: I agree it looks inverted.

shefft (changes requested):

aabogdanov (approved):

Hephaestus-Maximus:
Rebased this branch off `3.1.0-rc` instead of `develop` (the PR's base was already `3.1.0-rc`, but the branch content was based on `develop`, which had diverged by 8 unrelated commits — producing a bloated 52-file/10-commit diff). Confirmed `3.1.0-rc`'s copy of the three touched files was byte-identical to the original pre-fix state, so both commits cherry-picked cleanly with zero conflicts. Now: 2 commits, 3 files, 47/-7. Full SDK test suite re-run and green (460 pass) on the new base.

waltjohnson (approved):

src/ISStatusDecode.cpp
Hephaestus-Maximus: Resolved by rewrite: `gerrInv` (and the classification-flip trick it implemented) is gone entirely in the current version of this PR. The fix no longer reclassifies these bits as errors at all — it redefines the bit itself (`GPX_STATUS_COM0/1/2/USB_RX_TRAFFIC_DETECTED`, same mask, inverted wire meaning), and `buildGpxStatusDecode()` now uses the plain `gerr()` helper with the new names, no special-casing. `isError` stays `false` throughout — these are a status, not a fault, per Kyle's explicit direction. So the specific concern here (a narrowly-correct-by-accident helper being misleading for future reuse) no longer applies since the helper is gone.

src/ISStatusDecode.cpp
```diff
@@ -495,10 +495,15 @@ status_field_decode_t buildGpxStatusDecode()
...
-    d.subfields.push_back(gerr("COM0 RX traffic lost", GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED, "0x00000010 - COM0 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerr("COM1 RX traffic lost", GPX_STATUS_COM1_RX_TRAFFIC_NOT_DETECTED, "0x00000020 - COM1 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerr("COM2 RX traffic lost", GPX_STATUS_COM2_RX_TRAFFIC_NOT_DETECTED, "0x00000040 - COM2 RX traffic not detected in last 30 seconds."));
-    d.subfields.push_back(gerr("USB RX traffic lost", GPX_STATUS_USB_RX_TRAFFIC_NOT_DETECTED, "0x00000080 - USB RX traffic not detected in last 30 seconds."));
+    // SN-8402: these four are a status, not a fault (deliberately outside GPX_STATUS_GENERAL_FAULT_MASK,
+    // isError stays false). The bit itself was redefined (GPX_STATUS_COM*_RX_TRAFFIC_NOT_DETECTED ->
+    // GPX_STATUS_COM*_RX_TRAFFIC_DETECTED, same mask, inverted meaning) so the wire value now directly
+    // matches its positive name -- no display-side inversion needed. Firmware that sets this bit
+    // (communications.cpp) was updated to match.
```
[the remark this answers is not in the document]
Hephaestus-Maximus: Good catch on the description/code mismatch — that was a stale PR description from an intermediate revision; fixed now (see the current PR description, which no longer mentions `boolInvert` at all).

To be explicit about the actual current state: yes, this hunk intentionally redefines the wire bit itself. That's not accidental — it's the reworked approach after Kyle (the ticket owner) explicitly directed that the correct fix inverts the underlying bit definition rather than adding a display-side translation layer (the earlier `boolInvert` approach). `ISStatusDecode.h` has no `boolInvert` field anymore; I removed it in this rework since it's unused once the bit itself carries the right polarity.

Re: `communications.cpp` — that's correctly out of scope for *this* repo. The bit is actually set in `is-common`'s `cpp/hdw-src/hw-libs/families/gpx/communications/communications.cpp` (inertialsense/imx#1643), which this PR's description links as a companion PR — this repo (`inertial-sense-sdk`) only owns the enum definition + decode table, not the firmware that sets the bit.

Re: mis-decoding older firmware/logs — this is a real, valid concern and I've added it explicitly to the PR description as a documented, accepted limitation: replaying a log recorded with pre-fix firmware through a post-fix build of this SDK will show the new positive-polarity label applied to the old bit semantics, since there's no data-format flag distinguishing the two. This is intentional (matches how the org has handled other breaking wire-format renames, e.g. the GPS->GNSS identifier rename), not an oversight.

Hephaestus-Maximus:
Addressing Copilot's two "suppressed (low confidence)" comments from its latest review, which weren't posted as resolvable inline threads:

**`src/data_sets.h:4299` — stale doc comment / deprecated-alias suggestion.**
The stale-doc-comment part was a real catch: the `@brief` above `enum eGpxStatus` still said "per-port Rx-traffic-not-detected flags" after the rename. Fixed in `0aacde2c`.

The deprecated-alias suggestion (keep the old `_NOT_DETECTED` names defined pointing at the same mask) I'm declining, deliberately: unlike a same-meaning rename, this bit's *meaning* inverted. A deprecated alias here wouldn't just be a naming convenience — it would let old code keep compiling against `GPX_STATUS_COM0_RX_TRAFFIC_NOT_DETECTED` while silently getting the opposite of what that name says, which is worse than a compile error forcing a look at the call site. This is the same reasoning the org already applied to the GPS->GNSS identifier rename (no aliases kept there either).

**`tests/test_ISStatusDecode.cpp:265` — oracle text change.**
Yes, intentional — the legacy-oracle string is *supposed* to change here, since the actual on-wire bit meaning changed (see the compatibility note added to the PR description). The oracle exists to pin "what `RenderStatusFromDecode` should output for this decode table," not to preserve byte-identical output forever regardless of intentional behavior changes — the other `GpxStatus_RoundTrip_*` tests (every-bit, fatal-resets, random-sweep) all still pass against the updated oracle, confirming the renderer and the oracle agree post-change.

Hephaestus-Maximus (approved):
## Re-review: GPX RX-traffic status bits redefined as "detected" (SN-8402, revised)

This supersedes the version I reviewed on 2026-07-30 (Request Changes). That review flagged a row-title/polarity mismatch in the ribbon-chart display caused by relabeling the subfield name without changing what the bit actually meant. This revision takes a fundamentally better approach per Kyle's direction: redefine the bit itself (breaking change) rather than patch around it at the display layer. I re-verified from scratch rather than assuming the prior round's context still applies.

### Verified
- **Clean rename, no dangling references.** Grepped the whole repo post-checkout for `RX_TRAFFIC_NOT_DETECTED` and `boolInvert` (the now-removed display-layer flag from the intermediate revision) — the only hit is a comment describing the rename history, not live code.
- **My original concern is now actually resolved, not just patched.** The bit itself now carries positive polarity (1 = detected), so the row title ("COM0 RX traffic detected") and the runtime value agree — no more title/fault-color contradiction. `isError` staying `false` is a deliberate, documented product call (these are status bits, not faults — not every port is expected to have traffic), not an oversight; I'm deferring to that judgment rather than re-litigating it.
- **Test suite reproduced exactly**: built and ran the full suite — `460 tests, 456 passed, 4 skipped` (identical skip list), matching the description precisely. The new `GpxStatus_RxTrafficSubfields_PositivePolarityNotFault` test passes and is well-targeted (asserts name, `isError==false`, positive-polarity legacy text, and full round-trip render equality).
- **Cross-checked the companion firmware fix** (`imx#1643`, currently open/unreviewed): `communications.cpp`'s `setGPXStatus()` correctly inverts the condition for COM0/1/2 — old: `if (idle > 30000) status |= ..._NOT_DETECTED`; new: `if (idle <= 30000) status |= ..._DETECTED`. That's the correct logical negation, not just a rename with unchanged logic. Confirms the SDK-side bit redefinition and the firmware that actually sets the bit will agree once both land.
- CI green (Linux + Windows).

### Non-blocking notes
- `GPX_STATUS_USB_RX_TRAFFIC_DETECTED` isn't actually set anywhere in `imx` (grepped the whole repo, old and new names both return zero hits outside the SDK's own decode/test code). This predates this PR — not a regression — but worth knowing the USB variant of this field has apparently never been wired up on real hardware.
- Copilot's latest comment suggests keeping the old `_NOT_DETECTED` names as deprecated aliases for source compatibility. Reasonable in the abstract, but the PR description already explicitly documents this as an accepted, deliberate breaking change (with the specific external-tooling and historical-log-replay consequences spelled out) rather than an oversight — I don't think aliasing is necessary here given that explicit tradeoff was already made with eyes open.
- `imx#1643`'s SDK submodule pin points at an intermediate commit of this branch (`7194c49f`), not the current HEAD (`0aacde2c`, which only adds a doc-comment fix on top) — just flagging so whoever merges remembers to bump that submodule pointer once this PR's final SHA is in.

### Description
5/5 — the revision explicitly documents the breaking change, why the prior approach was wrong, and the concrete compatibility consequences. No edit needed.

Approving.

aabogdanov (approved):

kylemallory (approved):

shefft (approved):

PR #1258, opened 2026-08-02
Open up MAX_DATASET_SIZE

waltjohnson:
### Context

Copilot flagged on the SN-8405 PR that `MAX_PKT_BODY_SIZE` (and its derived macros) compute packet capacity by halving `PKT_BUF_SIZE`, based on a comment assuming "packet encoding doubles size." That assumption dates to the pre-2023 v1 ISB protocol, which escaped reserved bytes via `PSC_RESERVED_KEY` (see the old `encodeByteAddToBuffer()`). The v2 packet format rewrite (SN-3585, #456) removed all byte-stuffing — the wire format is now purely length-prefixed binary — but never updated this math, so the macros understated real packet capacity by ~2x ever since (even before SN-8405 raised `MAX_DATASET_SIZE`, `MAX_PKT_BODY_SIZE` was already inconsistent with it).

### Changes

- **`PKT_OVERHEAD_SIZE`**: now `ISB_MIN_PACKET_SIZE` (`sizeof(packet_hdr_t) + 2`) instead of a hardcoded `8` with a comment describing v1-era fields (`START_BYTE`, `COUNTER_BYTE`, three checksum bytes) that no longer exist.
- **`MAX_PKT_OVERHEAD_SIZE`**: dropped the `PKT_OVERHEAD_SIZE * 2 - 2` doubling; now equals `PKT_OVERHEAD_SIZE` directly.
- **`MAX_PKT_BODY_SIZE`**: dropped the erroneous `/ 2`. Now `2040` bytes (`PKT_BUF_SIZE - PKT_OVERHEAD_SIZE`, rounded down to even).
- **`MAX_DATASET_SIZE`**: now `MAX_PKT_BODY_SIZE` directly — **2040 bytes** (was a hardcoded `1280`, and briefly `MAX_P_DATA_BODY_SIZE` / 2035 in an earlier revision of this PR — see "Follow-up" below). Since a dataset must fit in a single ISB packet — `ISB_FLAGS_EXTENDED_PAYLOAD` is defined but has no reassembly implementation anywhere in the parser — tying it to actual packet capacity means it no longer needs manual bumping every time a DID grows.
- Reordered these macros to sit after `packet_hdr_t`/`p_data_hdr_t`/`p_ack_hdr_t` are declared, since they're now computed from `sizeof()` those structs.
- **`ISComm.c`**: removed the `DID_CAL_TEMP_COMP_GYR/ACC/MAG` bypasses in the `PKT_TYPE_SET_DATA`/`PKT_TYPE_DATA`/`PKT_TYPE_GET_DATA` size validation. Those three DIDs carry `sensor_tcal_{gyr,acc,mag}_group_t` (1620 bytes for gyr/acc, 324 for mag — the per-sensor-type subset, not the combined 3564-byte `sensor_tcal_group_t` the DID comment names) and were hardcoded exceptions because they exceeded the old 1280-byte ceiling. They fit comfortably under the new 2040-byte ceiling, so the exemption — which let those IDs bypass the size check for *any* size — is no longer needed.

### Follow-up: why `MAX_PKT_BODY_SIZE` and not `MAX_P_DATA_BODY_SIZE`

An earlier revision of this PR defined `MAX_DATASET_SIZE` as `MAX_P_DATA_BODY_SIZE` (`MAX_PKT_BODY_SIZE - sizeof(p_data_hdr_t)`, 2035 bytes), reasoning that a `p_data_hdr_t` (id + size + offset) rides in the payload alongside the data. It doesn't. `packet_t`'s union aliases `dataHdr.id`/`dataHdr.size` directly onto `packet_hdr_t.id`/`packet_hdr_t.payloadSize` — fields already inside the 6-byte header that `PKT_OVERHEAD_SIZE` accounts for; they're never serialized a second time. Only `dataHdr.offset` (2 bytes) is a real extra wire byte, and it's optional (`ISB_FLAGS_PAYLOAD_W_OFFSET`) — and `processIsbPkt` already computes `pkt->data.size = payloadSize - 2` when it's present, netting it out before anything compares against `MAX_DATASET_SIZE`. So no further subtraction is needed: `MAX_DATASET_SIZE = MAX_PKT_BODY_SIZE` is the correct, non-double-counted ceiling.

### Why it's safe

In `is-imx`, `CMakeLists.txt` builds the firmware from an explicit source list (`ISComm.c`, `com_manager.cpp`, `data_sets.c`, etc.) that excludes `DataJSON.cpp`, `DataCSV.cpp`, `ISDisplay.cpp`, and `ISLogger.cpp` — the only files where `MAX_DATASET_SIZE`/`MAX_P_DATA_BODY_SIZE` size a buffer rather than a scalar comparison. Those files are host-only (SDK/EvalTool/NavPostProcess), so growing the constant costs no embedded RAM there. Other consumers of this shared SDK header (`is-gpx`, host tool builds) may compile more broadly; this hasn't been independently re-verified for every consumer, but wherever the constant only backs a scalar bound check (as it does in `ISComm.c` itself) the change is free regardless of what else gets compiled alongside it.

### Testing

- `gcc -fsyntax-only` on `ISComm.c` — clean.
- Built the full SDK from source and ran the complete unit test suite: **459 tests, 455 passed, 4 skipped** (pre-existing skips: missing live-fixture log files and an opt-in E2E soak test — unrelated to this change).
- Targeted filter (`*ISComm*:*ComManager*:*com_manager*:*Protocol*:*Logger*:*DeviceCal*`): 42/42 passed, including `test_ISDeviceCal.cpp`, which directly exercises `sensor_tcal_group_t` — confirms the calibration-DID bypass removal doesn't break calibration upload/download.

### DIFF
--- a/src/ISComm.c
+++ b/src/ISComm.c
@@ -516,10 +516,7 @@ static protocol_type_t processIsbPkt(void* v)
     case PKT_TYPE_SET_DATA:
     case PKT_TYPE_DATA:
         // Validate data size
-        if (pkt->data.size <= MAX_DATASET_SIZE ||
-            pkt->hdr.id == DID_CAL_TEMP_COMP_GYR ||
-            pkt->hdr.id == DID_CAL_TEMP_COMP_ACC ||
-            pkt->hdr.id == DID_CAL_TEMP_COMP_MAG)
+        if (pkt->data.size <= MAX_DATASET_SIZE)
         {
             if (ptype==PKT_TYPE_SET_DATA)
             {   // acknowledge valid data received
@@ -538,10 +535,7 @@ static protocol_type_t processIsbPkt(void* v)
         {
             p_data_get_t *get = (p_data_get_t*)&(isbPkt->payload.data);
             // Validate data size
-            if (get->size <= MAX_DATASET_SIZE ||
-                get->id == DID_CAL_TEMP_COMP_GYR ||
-                get->id == DID_CAL_TEMP_COMP_ACC ||
-                get->id == DID_CAL_TEMP_COMP_MAG)
+            if (get->size <= MAX_DATASET_SIZE)
             {   // Update data pointer
                 return _PTYPE_INERTIAL_SENSE_CMD;
             }
--- a/src/ISComm.h
+++ b/src/ISComm.h
@@ -131,15 +131,6 @@ typedef enum
 /** Default protocol enable mask used by is_comm_init() when no explicit mask is set. */
 #define DEFAULT_PROTO_MASK (ENABLE_PROTOCOL_ISB | ENABLE_PROTOCOL_NMEA | ENABLE_PROTOCOL_UBLOX | ENABLE_PROTOCOL_RTCM3)
 
-/** The maximum allowable dataset size. SN-8405: raised from 1024 to cover gnss_sig_t's new
- *  max size (1128 bytes at MAX_NUM_SAT_SIGNALS=160) with headroom; a single ISB packet can
- *  carry far more than this (see is_comm_write_isb_precomp_to_port's PKT_BUF_SIZE check), so
- *  this does not require any change to PKT_BUF_SIZE. */
-#define MAX_DATASET_SIZE        1280
-
-/** The decoded overhead involved in sending a packet - 4 bytes for header, 4 bytes for footer */
-#define PKT_OVERHEAD_SIZE       8       // = START_BYTE + INFO_BYTE + COUNTER_BYTE + FLAGS_BYTE + CHECKSUM_BYTE_1 + CHECKSUM_BYTE_2 + CHECKSUM_BYTE_3 + END_BYTE
-
 /** The maximum buffer space that is used for sending and receiving packets */
 #ifndef PKT_BUF_SIZE
 #define PKT_BUF_SIZE            2048
@@ -148,17 +139,9 @@ typedef enum
 /** The maximum time between received data that will reset in the parser */
 #define MAX_PARSER_GAP_TIME_MS  100
 
-/** The maximum encoded overhead size in sending a packet (7 bytes for header, 7 bytes for footer). The packet start and end bytes are never encoded. */
-#define MAX_PKT_OVERHEAD_SIZE   (PKT_OVERHEAD_SIZE + PKT_OVERHEAD_SIZE - 2)  // worst case for packet encoding header / footer
-
-/** The maximum size of an decoded packet body */
-#define MAX_PKT_BODY_SIZE       (((PKT_BUF_SIZE - MAX_PKT_OVERHEAD_SIZE) / 2) & 0xFFFFFFFE) // worst case for packet encoding body, rounded down to even number
-
-/** The maximum size of decoded data in a packet body */
-#define MAX_P_DATA_BODY_SIZE    (MAX_PKT_BODY_SIZE-sizeof(p_data_hdr_t))    // Data size limit
-
-/** The maximum size of a decoded ACK message */
-#define MAX_P_ACK_BODY_SIZE     (MAX_PKT_BODY_SIZE-sizeof(p_ack_hdr_t))     // Ack data size
+// MAX_DATASET_SIZE, PKT_OVERHEAD_SIZE, MAX_PKT_OVERHEAD_SIZE, MAX_PKT_BODY_SIZE, and
+// MAX_P_DATA_BODY_SIZE are defined below, once packet_hdr_t and p_data_hdr_t exist, since they're
+// computed from sizeof() those structs. MAX_P_ACK_BODY_SIZE is defined further below, after p_ack_hdr_t.
 
 /** Binary checksum start value */
 #define CHECKSUM_SEED 0x00AAAAAA
@@ -418,6 +401,30 @@ typedef struct
 #define ISB_MIN_PACKET_SIZE             (sizeof(packet_hdr_t) + 2)                                      //!< Minimum ISB packet size: header + 2-byte checksum, no payload
 #define ISB_HDR_TO_PACKET_SIZE(hdr)     ((hdr).size + ISB_MIN_PACKET_SIZE + ((hdr).offset ? 2 : 0))     //!< Compute total ISB packet byte size from a packet_hdr_t
 
+/** The overhead involved in sending a packet: @ref packet_hdr_t header + 2-byte checksum footer.
+ *  Same computation as @ref ISB_MIN_PACKET_SIZE (a packet with zero-length payload is pure overhead). */
+#define PKT_OVERHEAD_SIZE       ISB_MIN_PACKET_SIZE
+
+/** The maximum overhead size in sending a packet. Equal to @ref PKT_OVERHEAD_SIZE: protocol 2.x
+ *  (see PROTOCOL_VERSION_CHAR0) is a length-prefixed binary format with no byte-stuffing/escaping,
+ *  so unlike the old v1 protocol's PSC_RESERVED_KEY escaping, there is no worst-case encoding growth
+ *  to account for here. */
+#define MAX_PKT_OVERHEAD_SIZE   PKT_OVERHEAD_SIZE
+
+/** The maximum size of a decoded packet body: full buffer minus header/footer overhead, rounded down to an even number. */
+#define MAX_PKT_BODY_SIZE       ((PKT_BUF_SIZE - MAX_PKT_OVERHEAD_SIZE) & 0xFFFFFFFE)
+
+/** The maximum size of decoded data in a packet body */
+#define MAX_P_DATA_BODY_SIZE    (MAX_PKT_BODY_SIZE-sizeof(p_data_hdr_t))    // Data size limit
+
+/** The maximum allowable dataset size: tied directly to @ref MAX_P_DATA_BODY_SIZE, the actual
+ *  per-packet data capacity, since a dataset must fit in a single (non-fragmented) ISB packet --
+ *  ISB_FLAGS_EXTENDED_PAYLOAD exists as a flag bit but has no reassembly implementation anywhere
+ *  in this parser. SN-8405 raised this from a hardcoded 1024 to cover gnss_sig_t's new max size
+ *  (1128 bytes at MAX_NUM_SAT_SIGNALS=160); deriving it from MAX_P_DATA_BODY_SIZE instead means
+ *  it no longer needs manual bumping as datasets grow, as long as they still fit in one packet. */
+#define MAX_DATASET_SIZE        MAX_P_DATA_BODY_SIZE
+
 /** Represents a packet header and body */
 typedef struct
 {
@@ -524,6 +531,9 @@ typedef struct
     /** Packet counter of the received packet */
     uint16_t            pktCounter;
 } p_ack_hdr_t;
+
+/** The maximum size of a decoded ACK message */
+#define MAX_P_ACK_BODY_SIZE     (MAX_PKT_BODY_SIZE-sizeof(p_ack_hdr_t))     // Ack data size
 
 /** Represents the entire body of an ACK or NACK packet */
 typedef struct
--- a/src/ISComm.h
+++ b/src/ISComm.h
@@ -417,13 +417,10 @@ typedef struct
 /** The maximum size of decoded data in a packet body */
 #define MAX_P_DATA_BODY_SIZE    (MAX_PKT_BODY_SIZE-sizeof(p_data_hdr_t))    // Data size limit
 
-/** The maximum allowable dataset size: tied directly to @ref MAX_P_DATA_BODY_SIZE, the actual
- *  per-packet data capacity, since a dataset must fit in a single (non-fragmented) ISB packet --
- *  ISB_FLAGS_EXTENDED_PAYLOAD exists as a flag bit but has no reassembly implementation anywhere
- *  in this parser. SN-8405 raised this from a hardcoded 1024 to cover gnss_sig_t's new max size
- *  (1128 bytes at MAX_NUM_SAT_SIGNALS=160); deriving it from MAX_P_DATA_BODY_SIZE instead means
- *  it no longer needs manual bumping as datasets grow, as long as they still fit in one packet. */
-#define MAX_DATASET_SIZE        MAX_P_DATA_BODY_SIZE
+/** The maximum allowable dataset size: tied directly to the maximum packet payload capacity.
+ *  Note: for packets with ISB_FLAGS_PAYLOAD_W_OFFSET set, pkt->data.size already excludes the 2-byte offset.
+ *  This limit intentionally bounds pkt->data.size (dataset bytes), not on-wire payloadSize. */
+#define MAX_DATASET_SIZE        MAX_PKT_BODY_SIZE
 
 /** Represents a packet header and body */
 typedef struct


tmcd249 (approved):

kylemallory (approved):

src/ISComm.h
```diff
@@ -418,6 +401,30 @@ typedef struct
...
+/** The maximum size of decoded data in a packet body */
+#define MAX_P_DATA_BODY_SIZE    (MAX_PKT_BODY_SIZE-sizeof(p_data_hdr_t))    // Data size limit
+
+/** The maximum allowable dataset size: tied directly to @ref MAX_P_DATA_BODY_SIZE, the actual
+ *  per-packet data capacity, since a dataset must fit in a single (non-fragmented) ISB packet --
+ *  ISB_FLAGS_EXTENDED_PAYLOAD exists as a flag bit but has no reassembly implementation anywhere
+ *  in this parser. SN-8405 raised this from a hardcoded 1024 to cover gnss_sig_t's new max size
+ *  (1128 bytes at MAX_NUM_SAT_SIGNALS=160); deriving it from MAX_P_DATA_BODY_SIZE instead means
+ *  it no longer needs manual bumping as datasets grow, as long as they still fit in one packet. */
+#define MAX_DATASET_SIZE        MAX_P_DATA_BODY_SIZE
```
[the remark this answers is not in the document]
waltjohnson: Fixed

Hephaestus-Maximus: **Confirming and expanding on Copilot's comment above (which got an "Fixed" reply, but the diff is unchanged since then — same head commit `f080379`).**

Traced this against the actual wire-validation code in `ISComm.c`:
```c
case PKT_TYPE_SET_DATA:
case PKT_TYPE_DATA:
    if (pkt->data.size <= MAX_DATASET_SIZE)   // ISComm.c:519
```
and a few lines up in the same function:
```c
pkt->data.size = payloadSize;   // (or payloadSize - 2 if the offset flag is set)
```
So for `PKT_TYPE_DATA`/`PKT_TYPE_SET_DATA`, `pkt->data.size` is the *raw* payload byte count — there's no separate `p_data_hdr_t` consuming any of those bytes on the wire for this packet type (that struct is only parsed out explicitly for `PKT_TYPE_GET_DATA`'s `p_data_get_t`, a different, small fixed struct). The true per-packet ceiling for `pkt->data.size` is `MAX_PKT_BODY_SIZE` (2040), not `MAX_P_DATA_BODY_SIZE`/`MAX_DATASET_SIZE` (2035) — the `-sizeof(p_data_hdr_t)` subtraction only makes sense for `p_data_buf_t` (`hdr` + `buf[MAX_DATASET_SIZE]`, `ISComm.h:502`), where the header genuinely does share the same packet body as the buffer.

Net effect: a dataset sized 2036–2040 bytes would legitimately fit in one ISB packet, but `MAX_DATASET_SIZE` would reject it at both the sender's `pkt->data.size` check here and the `get->size` check at line 538 — a narrow (5-byte) but real gap in exactly the invariant this PR sets out to establish ("a dataset must fit in a single ISB packet" / "no longer needs manual bumping"). No *currently* existing dataset lands in that window (largest checked: `sensor_tcal_gyr_group_t` at 1620 bytes, `gnss_sig_t` at 1128 bytes), so this has zero live impact today — but it's the exact kind of edge a future DID growth spurt could hit, silently reintroducing the manual-bump problem this PR is removing.

Suggest either bounding the wire-level `PKT_TYPE_DATA`/`SET_DATA`/`GET_DATA` checks against `MAX_PKT_BODY_SIZE` directly instead of `MAX_DATASET_SIZE`, or adding a second constant for the two different use cases if the header-reserving margin is still wanted somewhere for consistency. Given @waltjohnson's "Fixed" reply above, this may already be in progress elsewhere — flagging since I don't see it reflected in this diff yet.

Hephaestus-Maximus (changes requested):
## Review: Open up MAX_DATASET_SIZE (fix stale byte-stuffing assumptions)

Substantial-tier review (protocol-level sizing constants + a receive-side validation change, shared by every consumer of this SDK). Built the SDK from source, wrote a small standalone program to empirically check every macro/struct size claimed in the description, ran the full test suite, and cross-checked the "why it's safe" firmware-RAM argument against the actual imx build graph.

### Independently verified — all correct
- **Byte-math**: compiled a tiny program against this branch's headers. Every number in the description checks out exactly: `sizeof(packet_hdr_t)=6`, `sizeof(p_data_hdr_t)=5` (packed), `PKT_OVERHEAD_SIZE=8`, `MAX_PKT_BODY_SIZE=2040`, `MAX_P_DATA_BODY_SIZE`/`MAX_DATASET_SIZE=2035`. Confirms the old macro really was understating capacity by ~2x (was 1016/1280 with the erroneous `/2`).
- **Calibration DID bypass removal**: initially worried `sensor_tcal_group_t` (the type named in the DID's doc-comment) is 3564 bytes on this build target (v1.4, host/IMX-6), which would *exceed* the new 2035-byte ceiling. Dug further: the DIDs actually carry the smaller "one sensor type at a time" sub-structs — `sensor_tcal_gyr_group_t`/`sensor_tcal_acc_group_t` measured at exactly 1620 bytes each (matching the description precisely), `sensor_tcal_mag_group_t` at 324. All comfortably under 2035. The bypass removal is safe.
- **Testing claims reproduced exactly**: built and ran the full suite — `459 tests, 455 passed, 4 skipped` (identical skip list) — and the targeted filter — `42/42 passed`. Both match the PR description number-for-number.
- **Rename/reorder**: macros correctly reference `sizeof(packet_hdr_t)`/`sizeof(p_data_hdr_t)` after those types are declared; no leftover references to the old hardcoded values found anywhere in this repo.
- **Base-branch dependency** (`feature/SN-8405-increase-max-satellites-signals`, PR #1257): open but already `APPROVED` by Kyle — not a blocker, just noting the stacking for whoever merges.

### One clarification on the "why it's safe" RAM argument
The description says the firmware build "only compiles `ISComm.c`/`com_manager.cpp`/`data_sets.c` etc. from this SDK." Checked the actual build graph (imx repo): the IMX-5 firmware target links against the same `InertialSenseSDK` static library built from the SDK's root `CMakeLists.txt`, which globs `src/*.cpp` broadly — so `DataJSON.cpp`/`DataCSV.cpp`/`ISDisplay.cpp` *are* compiled into that library, contrary to the literal claim. The safety conclusion still holds, just via a different mechanism than stated: IMX-5's own firmware source (`IS_IMX/src`) never references any DataJSON/DataCSV/ISDisplay symbols, and the firmware `Makefile` uses `-ffunction-sections -fdata-sections` + `-Wl,--gc-sections`, so those unreferenced translation units get stripped at link time regardless of what got compiled. Net effect (zero firmware RAM growth) is the same — just flagging that the mechanism description is slightly off, in case that reasoning gets reused/cited elsewhere later.

### Blocking finding
Copilot already flagged this (see the existing inline thread) and got a "Fixed" reply from @waltjohnson — but the diff is unchanged since then (same head commit). Confirmed the underlying claim is real: `MAX_DATASET_SIZE` is derived as `MAX_PKT_BODY_SIZE - sizeof(p_data_hdr_t)`, but the receive-side check for `PKT_TYPE_DATA`/`PKT_TYPE_SET_DATA` (`ISComm.c:519`) validates `pkt->data.size`, which is the *raw* payload byte count with no `p_data_hdr_t` consuming any of it for that packet type (confirmed by reading `pkt->data.size = payloadSize` a few lines above). So the check is unnecessarily conservative by `sizeof(p_data_hdr_t)` (5 bytes) versus the actual per-packet capacity (`MAX_PKT_BODY_SIZE`), for both the `PKT_TYPE_DATA`/`SET_DATA` size check and the `PKT_TYPE_GET_DATA` `get->size` check at line 538. No dataset in this repo currently lands in that 5-byte window, so there's zero live impact — but it directly undercuts the PR's own stated goal ("no longer needs manual bumping ... as long as it still fits in one packet"), since a future dataset in that narrow range would hit exactly the problem this PR is meant to eliminate. See inline comment for the specific trace and a suggested fix. Requesting changes on this one point — everything else here is solid and independently verified.

### Description
5/5 — exceptionally thorough; explains the root cause, the fix, the safety audit, and testing performed, with specific byte counts that (per the above) all checked out. No edit needed.

src/ISComm.h
```diff
@@ -418,6 +401,27 @@ typedef struct
...
+/** The maximum size of a decoded packet body: full buffer minus header/footer overhead, rounded down to an even number. */
+#define MAX_PKT_BODY_SIZE       ((PKT_BUF_SIZE - MAX_PKT_OVERHEAD_SIZE) & 0xFFFFFFFE)
+
+/** The maximum size of decoded data in a packet body */
+#define MAX_P_DATA_BODY_SIZE    (MAX_PKT_BODY_SIZE-sizeof(p_data_hdr_t))    // Data size limit
+
+/** The maximum allowable dataset size: tied directly to the maximum packet payload capacity.
+ *  Note: for packets with ISB_FLAGS_PAYLOAD_W_OFFSET set, pkt->data.size already excludes the 2-byte offset.
+ *  This limit intentionally bounds pkt->data.size (dataset bytes), not on-wire payloadSize. */
+#define MAX_DATASET_SIZE        MAX_PKT_BODY_SIZE
```
Hephaestus-Maximus, line 423: Independently verified this correction (not just trusting Copilot's flag or the "Fixed" reply): read `is_comm_encode_hdr()` directly. `p_data_hdr_t`'s `id`/`size` fields are **not** separately serialized on the wire — `id` comes from `packet_hdr_t.id`, and `size` is derived from `payloadSize`, never transmitted as its own field. Only the optional 2-byte offset is actually present in the payload (`pkt->hdr.payloadSize += 2` when `offset` is set). Confirmed via `pkt->size = payloadSize + sizeof(packet_hdr_t) + 2` — no `p_data_hdr_t`-sized gap anywhere in the actual framing.

So `MAX_PKT_BODY_SIZE` (2040) is the correct ceiling for the no-offset case (the common one), and stays safe for the with-offset case too (true max there is 2038, and the encode-side check in `is_comm_write_isb_precomp_to_port` already independently caps `data.size` before this validation ever runs) — being 2 bytes permissive on this receive-side sanity/corruption check has no overflow or security consequence, since nothing sized by `MAX_DATASET_SIZE` (`p_data_buf_t::buf[]`, etc.) can be written past what was actually encoded.

Rebuilt + ran the full suite against this exact commit (not just the pre-fix one): 459 tests, 455 pass, 4 pre-existing skips, 0 failures — identical to the earlier commit's numbers, which makes sense since nothing in the current suite specifically probes the 2035-vs-2040 boundary (see my summary review for a concrete suggestion there).

src/ISComm.c
```diff
@@ -516,10 +516,7 @@ static protocol_type_t processIsbPkt(void* v)
     case PKT_TYPE_SET_DATA:
     case PKT_TYPE_DATA:
         // Validate data size
-        if (pkt->data.size <= MAX_DATASET_SIZE ||
-            pkt->hdr.id == DID_CAL_TEMP_COMP_GYR ||
-            pkt->hdr.id == DID_CAL_TEMP_COMP_ACC ||
-            pkt->hdr.id == DID_CAL_TEMP_COMP_MAG)
+        if (pkt->data.size <= MAX_DATASET_SIZE)
```
Hephaestus-Maximus, line 519: Verified this removal is safe, but via independent arithmetic rather than the cited test (see description addendum for why `test_ISDeviceCal.cpp` doesn't actually reach this code path). The three exempted DIDs each transmit only their own sensor-group slice, not the full `sensor_tcal_group_t` — confirmed via the actual `comManagerSendData(port, ..., DID_CAL_TEMP_COMP_GYR, SIZE_OF_SENSOR_TCAL_GYR)`-style call sites in `ISDeviceCal.cpp`. Worst case across all three DIDs and both hardware versions is 1620 bytes (`SIZE_OF_SENSOR_TCAL_GYR_V1P4` / `_ACC_V1P4`, 5 IMU devices × `sizeof(nvm_sensor_tcal_3axis_t)`=324B) — comfortably under the new 2040-byte ceiling. (For the record: the full combined `sensor_tcal_group_t`, summing gyr+acc+mag together, is 2592–3564 bytes depending on version — well over the ceiling. I initially computed against that and got confused for a bit until I found the actual per-DID send sizes.)

**Suggestion (non-blocking):** there's no test that actually exercises `processIsbPkt`'s validation gate for these three DIDs, or that probes the exact `MAX_DATASET_SIZE` boundary (2040 accepted, 2041 rejected) end-to-end. `test_ISComm.cpp` already has the right pattern for this (`is_comm_data(...)` to encode + `is_comm_parse(&comm)` to decode/validate, e.g. `TxRxWithOffsetTest`) — worth a follow-up test using that pattern with a `DID_CAL_TEMP_COMP_GYR`-sized payload and a boundary-sized payload, so a future change to either the calibration struct or these macros gets a real regression signal instead of relying on hand-verified arithmetic.

Hephaestus-Maximus (approved):
## Review summary

**Effort tier: Substantial/risky** — touches core protocol validation macros shared by every firmware target (IMX-5, IMX-6, GPX-1) plus all host tooling, and removes a compatibility bypass for three DIDs. Full checkout/build/test treatment.

**Verified independently, not taken on faith:**
- The core claim (v1 byte-stuffing removed in the v2 rewrite, so the old `/2` margin is vestigial): confirmed `PSC_RESERVED_KEY` and `encodeByteAddToBuffer()` exist nowhere in the current codebase, and the actual encode path (`is_comm_write_isb_precomp_to_port`) does a plain `memcpy` with zero byte substitution.
- `ISB_FLAGS_EXTENDED_PAYLOAD` ("has no reassembly implementation anywhere"): confirmed — the flag bit is defined and documented but never read/set anywhere else in the parser or encoder.
- **Caught that the PR head moved mid-review** (`11aec7d0`, "Apply copilot feedback") and re-verified against the actual current commit rather than the one I'd originally checked out. That commit is a real fix, not a nit: Copilot correctly identified that the original `MAX_DATASET_SIZE = MAX_P_DATA_BODY_SIZE` (2035) over-subtracted — `p_data_hdr_t`'s `id`/`size` aren't separately serialized on the wire (id comes from `packet_hdr_t.id`; size is derived from `payloadSize`). I independently re-derived this from `is_comm_encode_hdr()` rather than trusting the "Fixed" reply — confirmed `MAX_PKT_BODY_SIZE` (2040) is the mathematically correct ceiling, and is safe even in the with-offset case (true max 2038) since the encode-side check independently bounds `data.size` before this validation ever runs.
- Calibration-DID bypass removal: independently verified via the actual `comManagerSendData(..., SIZE_OF_SENSOR_TCAL_*)` call sites in `ISDeviceCal.cpp` — worst case across all three DIDs/both hardware versions is 1620 bytes, comfortably under 2040. (Initially miscalculated this against the full combined `sensor_tcal_group_t` — 2592-3564 bytes — before finding the actual per-DID transmitted sizes; flagging that dead end so the real verification path is clear.)
- Repo-wide sweep for every consumer of these macros: matches the PR's own audit exactly — every buffer-sized use (`p_data_buf_t::buf[]`, `ISDisplay.h`, `DataCSV`/`DataJSON` scratch buffers) is host-only; zero matches anywhere in `cpp/hdw-src/` (firmware source). RAM-neutral claim holds.
- Test suite: rebuilt and ran twice — once against the pre-fix commit (459/455/4-skip/0-fail, matching the description exactly) and again against the actual current head after the Copilot fix (identical 459/455/4-skip/0-fail — expected, since nothing in the current suite specifically probes the 2035-vs-2040 boundary). Targeted filter also reproduced exactly: 42/42.

**Findings (both left inline, both non-blocking):**
1. The `MAX_PKT_BODY_SIZE`/Copilot-fix correction is real and correct, but the PR description is now stale — still says "2035 bytes" throughout with no mention of the fix. Addressed via description addendum (graded 4/5, appended).
2. The bypass-removal safety is real, but the test cited in the description (`test_ISDeviceCal.cpp`) doesn't actually exercise the code path being changed (`processIsbPkt` never runs since `port=nullptr` short-circuits `comManagerSendData` first). Suggested a concrete follow-up test using the existing `is_comm_data`/`is_comm_parse` pattern in `test_ISComm.cpp`.

## Verdict: APPROVE

The core change is correct, and I verified the two most consequential claims (byte-stuffing removal, calibration-DID safety) through independent derivation rather than trusting the description — both hold, including after re-checking against a commit that moved mid-review. The two findings above are description-accuracy and test-coverage gaps, not functional defects; nothing here should block merge.

PR #1263, opened 2026-08-06
SN-8328: Update platform config and io presets to fit the 2026 hardware selection

tmcd249:
Update platform config and io presets to fit the 2026 hardware selection


src/data_sets.h
```diff
@@ -3156,7 +3156,13 @@ enum ePlatformConfig
     PLATFORM_CFG_TYPE_MASK                                                          = (int)0x0000003F,  //!< Mask for platform/carrier-board type field
     PLATFORM_CFG_TYPE_FROM_MANF_OTP                                                 = (int)0x00000080,  //!< Type is overwritten from manufacturing OTP memory.  Write protection, prevents direct change of platformType in flashConfig.
     PLATFORM_CFG_TYPE_NONE                                                          = (int)0,   //!< No/unknown carrier board (IMX-5 default)
-    PLATFORM_CFG_TYPE_RUG3_G0                                                       = (int)8,   //!< PCB RUG-3.x.         PPS disabled
+    PLATFORM_CFG_TYPE_BRK_GPX                                                       = (int)1,   //!< IS-IMX-GPX-DEV, external GPX-1 module:    PPS1 on G15 (pin 20), PPS2 on G13 (pin 14)
```
Hephaestus-Maximus, line 3171: **Flash-persisted breaking change, needs confirmation.** `platformType` value `1` was previously unassigned; value `21` used to mean `PLATFORM_CFG_TYPE_IMX_BRK_1` and now means `PLATFORM_CFG_TYPE_IG2_1` (see the `21` -> `IG2_1` line below). `platformConfig` is stored in device flash and survives firmware updates.

I traced the consumer (`imxPlatformConfigToFlashCfgIoConfig`): for *this* function specifically, old-BRK-GPX (21) and IG2_1 land in the same `case` block and produce identical io-config output, so there's no functional regression there. But `platformType` also looks like a hardware-identity field that's likely read elsewhere (telemetry, EvalTool, cloud dashboards, manufacturing records) — anything that keys off the raw integer `21` to mean "this is a BRK-GPX-DEV-1" would now silently read it as an IG2_1 unit after this firmware ships, with no error and no way to detect the mismatch from the field alone.

This release cycle already has precedent for exactly this situation (changelog: "(IMX) Flash config key incremented to cause flash config factory reset"). Has any BRK-GPX-DEV-1 hardware shipped/deployed with `platformType==21` in flash? If so, should this change bump the flash-config version key (or otherwise force reconfiguration) the same way?

src/data_sets.h
```diff
@@ -3169,10 +3175,8 @@ enum ePlatformConfig
...
     PLATFORM_CFG_TYPE_TBED2_G2_W_LAMBDA                                             = (int)20,  //!< Enable UBX input from Lambda
-    PLATFORM_CFG_TYPE_IMX_BRK_1                                                     = (int)21,  //!< IS-IMX-GPX-DEV-1:    PPS1 on G15 (pin 20), PPS2 on G13 (pin 14)
-    PLATFORM_CFG_TYPE_RUG4_G2                                                       = (int)22,  //!< PCB RUG-4.x:         PPS1 on G15 (pin 20), PPS2 on G11 (pin 16)
-    PLATFORM_CFG_TYPE_IG2_1                                                         = (int)23,  //!< IG-2.1 and later:    PPS1 on G15 (pin 20), PPS2 on G13 (pin 14)
-    PLATFORM_CFG_TYPE_COUNT                                                         = (int)24,  //!< Number of defined platform/carrier-board types
+    PLATFORM_CFG_TYPE_IG2_1                                                         = (int)21,  //!< IG-2.1 and later:    PPS1 on G15 (pin 20), PPS2 on G13 (pin 14)
+    PLATFORM_CFG_TYPE_COUNT                                                         = (int)22,  //!< Number of defined platform/carrier-board types
```
Hephaestus-Maximus, line 3191: **This one I could verify causes an actual functional regression, not just an identity mismatch.** Old value `22` meant `PLATFORM_CFG_TYPE_RUG4_G2`; it's now unassigned (split into `RUG4_SG5`=6 / `RUG4_GPX`=7), and `PLATFORM_CFG_TYPE_COUNT` dropped from 24 to 22.

Tracing what happens to a device that still has `platformType==22` in flash after this update:
- `imxPlatformConfigErrorCheck()` only resets `type` to `NONE` when `type > PLATFORM_CFG_TYPE_COUNT`. `22 > 22` is false, so `type==22` passes through unclamped (this off-by-one — should probably be `>=` — predates this PR, I checked `3.1.0-rc` before this branch touched anything, but this renumbering is what makes it newly reachable/consequential).
- `imxPlatformConfigToFlashCfgIoConfig()`'s `switch(type)` no longer has a `case PLATFORM_CFG_TYPE_RUG4_G2:` (removed), so `type==22` falls into the `default:` case, which is the "No GPS" branch — it disables both GNSS1 and GNSS2 sources/types.

Net effect: an existing RUG-4.x device flashed with the old `RUG4_G2` value would silently lose its GNSS routing on boot after this firmware update, with no error surfaced, until someone notices and re-flashes `platformType` to `6` or `7`. Same likely applies to old value `23` (previous `IG2_1`), except that one *does* get caught by the `> COUNT` check (23 > 22) and reset to `NONE` — also a config loss, but at least a clean, detectable one rather than a silent default fallthrough.

Given this is a real, reachable functional regression for any already-deployed RUG4_G2 (and IG2_1) hardware, I think this needs the same treatment as the BRK-GPX/IG2_1 case above: either confirm no such hardware exists in the field, or force a flash-config factory reset via the version-key bump this release already uses elsewhere for this exact class of problem.

src/imx_defaults.c
```diff
@@ -60,58 +93,118 @@ void imxPlatformConfigToRug3FlashCfgIoConfig(uint32_t *ioConfig, uint32_t platfo
...
-        break;
+        case PLATFORM_CFG_TYPE_RUG_G0:
+            SET_IO_CFG_GNSS1_TYPE(*ioConfig, IO_CONFIG_GNSS_TYPE_NONE);
+            SET_IO_CFG_GNSS2_TYPE(*ioConfig, IO_CONFIG_GNSS_TYPE_NONE);
+            break;
+        case PLATFORM_CFG_TYPE_RUG3_G1:
+            SET_IO_CFG_GNSS1_TYPE(*ioConfig, IO_CONFIG_GNSS_TYPE_UBLOX);
+            SET_IO_CFG_GNSS2_TYPE(*ioConfig, IO_CONFIG_GNSS_TYPE_NONE);
+            break ;
+        default:    // RUG_G0, RUG3_G1, RUG3_G2, RUG4_X20: onboard u-blox receiver (ZED-F9P / X20)
```
Hephaestus-Maximus: Minor: this `default` comment lists `RUG4_X20`, but this function (`imxPlatformConfigToRug3FlashCfgIoConfig`) early-returns at the top for any `type` other than `RUG_G0`/`RUG3_G1`/`RUG3_G2` (line 49-51), so `RUG4_X20` can never actually reach this `default` case. Looks like a copy-paste from the sibling `imxPlatformConfigToRug4FlashCfgIoConfig` below, which has the mirror-image problem at line 203 — its own `default` comment lists `RUG_G0, RUG3_G1, RUG3_G2` even though that function's guard only lets `RUG4_X20`/`RUG4_SG5`/`RUG4_GPX` through (and `RUG_G0` is handled by its own explicit `case` above, not `default`). Purely a comment accuracy nit, no functional effect.

Hephaestus-Maximus (dismissed):
## Summary

Reviewed the `ePlatformConfig` enum renumbering and the corresponding `imx_defaults.c` logic (`src/data_sets.h`, `src/imx_defaults.c`). Built the full SDK test suite against this branch (via the `tests/` CMake path) and ran it: 456 pass / 4 skip — identical to baseline, no regressions in existing coverage. `PLATFORM_CFG`/`imxPlatformConfig*` logic currently has zero test coverage of its own, so the concerns below weren't caught by any existing test and wouldn't be caught by CI either way.

## Main concern: this renumbers values persisted in device flash, with no visible migration path

`platformConfig` (including the `platformType` sub-field) is written to non-volatile flash and survives firmware updates. This PR reassigns several `platformType` integer values:

- `21`: was `IMX_BRK_1` (BRK-GPX-DEV-1) → now `IG2_1`
- `22`: was `RUG4_G2` → now unassigned (split into `RUG4_SG5`=6 / `RUG4_GPX`=7)
- `23`: was `IG2_1` → now unassigned (`PLATFORM_CFG_TYPE_COUNT` dropped 24→22)

I traced both cases through the actual consumers (left as inline comments on the specific lines) rather than just flagging the renumber in the abstract:

- **Value 21** happens to be a functional no-op in `imxPlatformConfigToFlashCfgIoConfig` (old `IMX_BRK_1` and new `IG2_1` land in the same `case` block, same io-config output) — but `platformType` also looks like it's read as a hardware-identity field elsewhere (telemetry/EvalTool/manufacturing records), where a silent `BRK-GPX-DEV-1 → IG2_1` relabel wouldn't be a no-op.
- **Value 22** is a real, verified functional regression: it's no longer caught by `imxPlatformConfigErrorCheck()`'s bounds check (pre-existing off-by-one: `type > COUNT` should probably be `>= COUNT`, so `22 > 22` is false) and falls into the `default:` ("No GPS") case in the io-config switch, since the old `RUG4_G2` case was removed. A device still holding `platformType==22` in flash would silently lose its GNSS routing on boot, with no error surfaced.

This same release cycle already has precedent for exactly this class of problem (changelog: "(IMX) Flash config key incremented to cause flash config factory reset"). This PR doesn't appear to include an equivalent safeguard.

**This is the part I can't resolve from source code alone and think needs a human answer:** has any BRK-GPX-DEV-1 or RUG4_G2 hardware actually shipped/deployed with these old values in flash? If yes, I'd suggest bumping the flash-config version key (or another explicit migration) the same way this release already does elsewhere. If no such hardware exists in the field, this is a non-issue and I'd be happy to re-review and approve once confirmed.

## Ticket reference looks mismatched

The PR title references `SN-8328`, but that ticket (per Jira) is an already-closed, unrelated logging-test-suite item — it doesn't describe platform-config/io-preset work at all. The only other PR referencing `SN-8328` (#1619, merged) is also unrelated. Worth double-checking this is the intended ticket number before merge, since it'll make the change hard to trace back later.

## Minor

- `src/imx_defaults.c:118` and `:203` — copy-pasted `default:` comments each list a carrier type that can't actually reach that `default` case in that function (see inline comments for specifics). No functional effect, just misleading if someone reads the comment as documentation of reachable states.
- No test coverage exists for `imxPlatformConfigErrorCheck()` / `imxPlatformConfigToFlashCfgIoConfig()` / `imxPlatformConfigTypeValid()`. Given how easy it is to introduce exactly this class of silent-flash-reinterpretation bug in this code, a handful of unit tests asserting old-and-new type/preset values through these functions would make future renumbers much safer to review.

## CI

Windows passed. Linux shows as "fail" in `gh pr checks`, but the underlying job (`92704649923`) actually completed with conclusion `cancelled` and recorded zero steps after sitting for ~15 minutes — that looks like a runner/infra issue unrelated to this diff, not a real test failure. Worth a re-run to get a clean signal, but I'm not blocking on it.

## Verdict

Requesting changes — not because the code is wrong on its own terms (the split/rename logic itself is consistent and the switch statements are exhaustive for the new values), but because the flash-persisted breaking-change risk above is real and verifiable, and I don't have visibility into IMX/RUG4/BRK field-deployment history to know whether it's actually reachable. If someone with that visibility confirms it's a non-issue (or a migration path is added), I'm happy to take another pass.

waltjohnson (approved):

kylemallory (approved):

shefft (approved):

Hephaestus-Maximus (commented):
## Re-review (branch updated since my last pass)

Diffed the new head (`d984f898`) against the commit I last reviewed (`de749c81`) rather than re-reading the whole PR. Two things changed in the files this PR actually touches:

1. **`RTK_CFG_BITS_MULTIPATH_MITIGATION` restored** in `src/data_sets.h` (via a merge of `3.1.0-rc`). Good catch/fix — I independently hit this exact missing symbol as a hard build break while reviewing a downstream consumer of this branch (inertial-sense-sdk's `imx` repo, PR #1650, which points its SDK submodule here); confirmed there that bumping to a commit at-or-past this merge resolves it. This branch is now caught up with `3.1.0-rc` on that front.
2. **Stale comment fixed** in `imxPlatformConfigToRug3FlashCfgIoConfig`'s `default` case (`src/imx_defaults.c`) — no longer lists `RUG4_X20`, which addressed the copy-paste nitpick I flagged inline.

## Still open, unaddressed

My core concern from the last review — the flash-persisted `platformType` renumbering (old `21`/`22`/`23` reassigned to different meanings, with a verified functional regression for old `22`/`RUG4_G2` falling into an unhandled default) — is **unchanged** in this diff, and neither my inline comments nor the question I asked (does BRK-GPX-DEV-1 or RUG4-G2 hardware exist in the field with these old values in flash?) have a reply. For what it's worth, Copilot's automated review independently flagged the identical underlying concern (enum renumbering + flash/OTP persistence = breaking change) on the same lines, for whatever that corroboration is worth.

I don't have a way to independently answer the "does this hardware exist in the field" question from source code alone — it's a manufacturing/deployment fact, not something I can verify or refute myself. Three reviewers have approved since my request-changes; I can't tell from the PR whether that's because the question was considered and judged a non-issue, or because it wasn't the focus of those reviews. I'm not re-requesting changes over an unresolved judgment call that's already had human eyes on it, but I want the open question on record and visible rather than silently dropped: **please confirm explicitly (in a comment or the Jira ticket) whether any device has shipped with `platformType` 21, 22, or 23 before this merges**, so whoever signs off is doing so knowingly rather than by omission.

Also still open: the PR title/description still reference `SN-8328`, which (per Jira) is an unrelated, already-closed logging-test ticket — worth a quick check that this is the ticket you meant to link.

## Verdict

**Comment, no verdict change from me.** The two things I flagged last time that *were* code-fixable are fixed. The one substantive risk left is a factual/business question outside what I can verify, already visible to three approving reviewers — I'm surfacing it explicitly one more time rather than guessing at an answer or blocking indefinitely on a question only the team can answer.

Hephaestus-Maximus (commented):
Re-checked after this re-request. Diffed the new head (`516b4dbf`) against the commit I last reviewed (`d984f898`) — the only change is a merge of `3.1.0-rc` into this branch (pulling in an already-merged, unrelated commit, SN-8445/#1265 on GPX log hardware-ID derivation). `src/data_sets.h` and `src/imx_defaults.c` — the actual substance of this PR — are byte-identical to what I last reviewed.

Nothing new for me to add. My open question from the last pass still stands and remains unanswered: please confirm whether any device has shipped with `platformType` 21, 22, or 23 before this merges. Submitting this to clear the re-requested-review state; no change to my prior assessment.

Hephaestus-Maximus (commented):
Re-checked again after this re-request. Same pattern as last time: the new head (`89125a5e`) differs from what I last reviewed only by a chain of `3.1.0-rc` merges (three already-landed, unrelated PRs — #1267, #1247, #1268). `src/imx_defaults.c` is untouched; the one `data_sets.h` change is from #1247's `DEV_INFO_MATCHES_HDW_ID` fix, unrelated to this PR's platform-config work.

This is now the third time I've re-reviewed a merge-only update with nothing new to assess. Submitting this to clear the re-requested-review state again, but flagging directly: my open question (whether any BRK-GPX-DEV-1/RUG4-G2 hardware has shipped with `platformType` 21/22 in flash) has been outstanding since 2026-08-06 with no reply. If the answer is "no such hardware exists, this is a non-issue," a one-line confirmation here or in SN-8328 would let me actually close this out instead of re-flagging it on every base-branch sync.

PR #1264, opened 2026-08-06
SN-8261 Advanced fault detection for multiple IMUs

aabogdanov:
## What

SDK-side support for SN-8261 (multi-IMU fault detection), companion to imx#1601:

- **`src/filters.cpp`/`.h`**: simplifies `multiToSingleImu`'s per-device status-mask computation (behavior-preserving — combines two shift operations into one), and removes `multiToSingleImuExc`/`multiToSingleImuAxis`, which the new parity-space fault detector in imx#1601 no longer calls (it re-implements exclusion-aware averaging inline).
- **`src/data_sets.h`**: adds `IMUS_STATUS_ALL_GYR_OK_MASK`/`IMUS_STATUS_ALL_ACC_OK_MASK` convenience masks covering all `MAX_IMU_DEVICES` (5) devices' gyro/accel OK bits.
- **`src/ISDataMappings.cpp`**: extends the `DID_BIT` field-description strings to cover the new IMU-fault BIT test commands/modes (`BIT_CMD_IMU_REJECT*`, `BIT_TEST_MODE_IMU_*`).
- **`python/inertialsense/logInspector/*`**: adds an "IMUs Status" plot (per-device, per-axis OK/saturation bits over time) and reworks the device-chooser dialog into a table showing hardware/firmware/build/protocol/commit/build-date per device, with an optional "More info" expansion.

## Note for reviewers

This PR's head is pinned as the SDK submodule commit in imx#1601 — the two are a matched pair. **Do not merge this independently of imx#1601 landing** (or without otherwise updating imx's current fault-detection code first): imx's `3.1.0-rc`/`develop` branches still call the now-removed `multiToSingleImuAxis` in their pre-#1601 implementation, and imx does frequent independent "SDK bump" merges — one of those picking up this commit before #1601 merges would break the imx build.


src/filters.cpp
```diff
@@ -220,88 +208,6 @@ void multiToSingleImu(imu_t *result, const imus_t *imus, const int numDevices)
...
-            result->status &= ~(base << iaxis);  // No valid data
-        }
-        res[iaxis] = mean;
-    }
-    result->time = di->time;
-    // result->status = di->status & IMUS_STATUS_SATURATION_MASK;
-}
-
-
 void singleToMultiImu(imus_t *result, imu_t *imu, const int numDevices)
```
Hephaestus-Maximus, line 211: **Coordination hazard, not a bug in this file — please confirm merge sequencing.** `multiToSingleImuExc` and `multiToSingleImuAxis` were removed right above this (in the same diff). imx's current `3.1.0-rc`/`develop` (i.e. everything except the not-yet-merged imx#1601) still calls `multiToSingleImuAxis` from the pre-#1601 fault-detection implementation in `NavINL_base.cpp`. imx does frequent independent "SDK bump" merges unrelated to this feature — the next one of those, after this PR merges but before imx#1601 also lands, would break the imx build. imx#1601 itself currently has blocking issues (see my review there) and isn't ready to merge yet, so there's a real window where that could happen. Suggest either merging both together, holding this until imx#1601 lands, or leaving the two functions as deprecated-but-present until the consumer is confirmed fully migrated.

src/data_sets.h
```diff
@@ -1004,6 +1004,8 @@ enum eImusStatus
     IMUS_STATUS_IMU_OK_MASK     = IMUS_STATUS_GYR_X_OK | IMUS_STATUS_GYR_Y_OK | IMUS_STATUS_GYR_Z_OK | IMUS_STATUS_ACC_X_OK | IMUS_STATUS_ACC_Y_OK | IMUS_STATUS_ACC_Z_OK,  //!< Mask of all IMU-valid bits
     IMUS_STATUS_GYR_OK_MASK     = IMUS_STATUS_GYR_X_OK | IMUS_STATUS_GYR_Y_OK | IMUS_STATUS_GYR_Z_OK,  //!< Mask of all gyro-valid bits
     IMUS_STATUS_ACC_OK_MASK     = IMUS_STATUS_ACC_X_OK | IMUS_STATUS_ACC_Y_OK | IMUS_STATUS_ACC_Z_OK,  //!< Mask of all accelerometer-valid bits
+    IMUS_STATUS_ALL_GYR_OK_MASK = IMUS_STATUS_GYR_OK_MASK<<0 | IMUS_STATUS_GYR_OK_MASK<<6 | IMUS_STATUS_GYR_OK_MASK<<12 | IMUS_STATUS_GYR_OK_MASK<<18 | IMUS_STATUS_GYR_OK_MASK<<24,  //!< Mask of all gyro-valid bits
```
Hephaestus-Maximus, line 1013: Nitpick: these two masks are hand-unrolled for 5 devices (matching today's `MAX_IMU_DEVICES`) rather than derived from that constant. Correct today, but if `MAX_IMU_DEVICES` is ever increased this would silently under-cover the new devices instead of failing to compile. A `static_assert(MAX_IMU_DEVICES == 5, ...)` next to these two lines (or deriving them programmatically) would make a future mismatch loud instead of silent.

Hephaestus-Maximus (changes requested):
Reviewed independently: cloned at head (`7fde9afe`), read every changed line in full context, and cross-checked against my concurrent review of the companion PR (imx#1601, same Jira ticket SN-8261) since this SDK PR's head is exactly the SDK submodule commit imx#1601 currently pins to.

The PR had no description, so I've added one summarizing the diff and, more importantly, flagging the coordination note below — please double check it captures the intent correctly.

## Requesting changes for one reason: a real cross-repo merge-ordering hazard, undisclosed

`src/filters.cpp`/`.h` deletes `multiToSingleImuExc` and `multiToSingleImuAxis` entirely. I checked: imx's **current** `3.1.0-rc`/`develop` branches (i.e. everything except the not-yet-merged imx#1601) still call `multiToSingleImuAxis` from `cpp/libs/nav-base/NavINL_base.cpp`'s pre-#1601 fault-detection implementation. imx#1601's new implementation no longer needs either function — I read through the whole new pipeline in that review and confirmed it re-implements exclusion-aware averaging inline rather than delegating to these helpers.

So the two PRs are consistent as a *pair* (confirmed: imx#1601's SDK submodule pointer is this exact commit), but merging this SDK PR to `3.1.0-rc` on its own, before imx#1601 also lands, creates a real hazard: imx does frequent independent "SDK bump" commits (I count over a dozen in the last month of history) unrelated to any specific feature work. The very next routine SDK-bump merge on `3.1.0-rc`/`develop` after this PR merges would break the imx build, since the currently-shipping fault-detection code still calls a function this PR removes.

This is easy to fix procedurally (merge together, or hold this until imx#1601 lands, or leave the two old functions in place as deprecated-but-present until the consumer is confirmed gone) — I'm requesting changes mainly so this gets an explicit decision rather than landing silently, especially since **imx#1601 itself currently has several blocking issues I found in its own review and isn't ready to merge yet** — meaning there's a real window where this SDK change could land on `3.1.0-rc` well before its consumer does.

## Everything else checked and found correct

- **`multiToSingleImu`'s refactor is behavior-preserving.** The old code computed the per-device status mask via two chained shifts (`(baseS << iaxis) << (idev * IMUS_STATUS_IMU_OK_BITSIZE)`); the new code computes the same total shift in one step (`baseS << (iaxis + idev * IMUS_STATUS_IMU_OK_BITSIZE)`). Shift-associativity holds here (no overflow risk — total shift stays well under 32 bits), so this is a pure simplification, not a behavior change.
- **New `IMUS_STATUS_ALL_GYR_OK_MASK`/`ALL_ACC_OK_MASK` constants are correct** for the current `MAX_IMU_DEVICES=5`/`IMUS_STATUS_IMU_OK_BITSIZE=6` (5 unrolled terms × 6 bits = 30 bits, leaving bits 30/31 for the two saturation flags — matches exactly). Minor nitpick: these are hand-unrolled for 5 devices rather than derived from `MAX_IMU_DEVICES`, so if that constant is ever increased, these masks would silently under-cover the new devices rather than failing to compile. Not urgent, but a `static_assert` tying them together would be cheap insurance.
- **New `DID_BIT` field descriptions match real enum values** — spot-checked every `BIT_CMD_IMU_*`/`BIT_TEST_MODE_IMU_*` constant referenced against its definition in `data_sets.h`; all exist and match. (Incidentally, `BIT_TEST_MODE_IMU_INVALID_DATA`'s own doc comment says NaN injection should be "detected and rejected" — worth knowing that's exactly the capability I found broken in the imx#1601 review, where the validity-check calls that would catch this are currently commented out. Not this PR's bug, just relevant context for whoever picks up BIT test coverage.)
- **New "IMUs Status" log-inspector plot is correct.** Bit ordering (`X/Y/Z Gyr OK, X/Y/Z Acc OK`, 6 bits per device) and the two saturation-bit masks (`0x40000000`/`0x80000000`) match `eImusStatus` exactly.
- **Device-chooser dialog table**: build-number decoding (`(buildNumber >> 12) & 0xFFFFF`, `buildNumber & 0xFFF`) matches the existing convention in `ISDevice.cpp`'s device-string formatter. UI-only change, no functional risk.

No build attempted for this one (it's header/data-mapping/Python tooling, not a runnable test target) — the changes are straightforward enough that a read-through plus the cross-file symbol check above covers it; happy to build-verify if there's a preferred target.

shefft (approved):

tmcd249 (approved):

kylemallory (approved):

Hephaestus-Maximus (approved):
Reviewed the full diff (7 files, +205/-135), the companion SDK-side PR to imx#1601 (which I just reviewed/approved).

**`filters.cpp`/`.h`**: verified the `multiToSingleImu` mask-computation simplification is exactly behavior-preserving, not just claimed to be — the old code computed `mask = (baseS << iaxis) << (idev * BITSIZE)`; the new single-shift form `baseS << (iaxis + idev * BITSIZE)` is the same value algebraically (shifting twice by a then b equals shifting once by a+b), so this is a pure identity, not a behavior change. Also confirmed the sensor-value read was moved inside the `if (status & mask)` branch rather than always executing before the conditional — harmless since it was a pure read with no side effects either way, but worth confirming since it's exactly the kind of reordering that occasionally hides a real change. Grepped the whole repo for `multiToSingleImuExc`/`multiToSingleImuAxis`: zero remaining references anywhere, so the removal is safe within this repo.

**`data_sets.h`**: the new `IMUS_STATUS_ALL_GYR_OK_MASK`/`IMUS_STATUS_ALL_ACC_OK_MASK` (shifts of 0/6/12/18/24) checked against `MAX_IMU_DEVICES=5` and `IMUS_STATUS_IMU_OK_BITSIZE=6` — correct, covers exactly 5 devices, no overflow (max shift 24+6=30 bits, within a 32-bit mask).

**`ISDataMappings.cpp`**: the new BIT command/test-mode description strings reference `BIT_CMD_IMU_REJECT*`/`BIT_TEST_MODE_IMU_*` — confirmed all of those already exist as real enum values in `data_sets.h` (pre-existing, not new to this PR), so this is a correct, low-risk documentation-string addition with no logic behind it.

**Python `logInspector`**: the new `imusStatus()` plot's bit-unpacking (`shift = imuIdx * 6`, bits 0-5 for X/Y/Z Gyr/Acc OK, `0x40000000`/`0x80000000` for saturation) correctly mirrors the C-side bit layout I just verified in `data_sets.h`.

**One real (though minor, cosmetic) issue** in the new device-chooser dialog (`logInspectorInternal.py`): `deviceInfoStrings()` unconditionally formats `hardwareStr = 'IMX-%d.%d.%d' % (hv[0], hv[1], hv[2])`, regardless of `devInfo['hardwareType']`. `dev_info_t.hardwareType` distinguishes IMX (3) from GPX (4) explicitly (per its comment in `data_sets.h`), and GPX-1 is a real, separate top-level product this same log-inspector tool supports — a GPX-1 log would show "IMX-x.y.z" in this new table. Worth noting this is a step backward from the existing convention elsewhere in this same tool (`logInspector.py`'s pre-existing device list just shows `'H:' + verArrayToString(...)` with no type-specific prefix at all) — the new code added a label that happens to be wrong for a whole product line the codebase otherwise treats as first-class. Not blocking given this is a diagnostic-tool label, not firmware behavior, but worth a quick fix (branch on `hardwareType`, or drop the prefix like the existing convention does).

**On the sequencing warning**: independently verified this is a real, current risk, not just a defensive disclaimer — cloned imx's actual `3.1.0-rc` branch and confirmed `NavINL_base.cpp` there still calls `multiToSingleImuAxis`. So the PR body's warning is accurate: merging this SDK PR before imx#1601 lands (or before imx's `3.1.0-rc`/`develop` is otherwise updated) would break the imx build on the next independent SDK-bump. Since imx#1601 is now approved on my end too, this is close to being safe to sequence, but the ordering still matters and isn't something a code review resolves on its own.

CI is green on both platforms. Approving — the one finding is cosmetic and non-blocking.

PR #1266, opened 2026-08-08
SN-8317: rename external-aiding DIDs, register in ISDataMappings, add GRMC bits

kylemallory:
Part of the SN-8317 / SN-8472 external-aiding work. Targets `SN-8318_external_aiding` so the whole feature integrates once.

## What this does

**Rename (no wire change).** `DID_EXT_POS`/`DID_EXT_VEL` → `DID_EXT_AIDING_POS`/`DID_EXT_AIDING_VEL`, `ext_pos_t`/`ext_vel_t` → `ext_aiding_pos_t`/`ext_aiding_vel_t`, `eExternalAidingStatus` → `eExtAidingFrame`, `EP_STATUS_FRAME_*` → `EXT_AIDING_FRAME_*`. DID numbers (106/107) and struct layout are untouched.

**Register the data sets.** ISDataMappings name-table entries plus `PopulateMapExtAidingPos/Vel`; `offsetsExtAidingPos` in data_sets.c so the three doubles in `pos[]` byte-flip on big-endian hosts.

**GRMC bits** for the GPX producer: positions 28/29, `GRMC_BIT_POS_COUNT` → 30, masks, `g_gpxDidToGrmcBit` and `g_gpxGRMCPresetLookup` entries.

## Decisions worth a look

- **Both DIDs are left writable** (no `DATA_FLAGS_READ_ONLY`) because they are inputs to the EKF. Side benefit: an observation can be injected with `cltool -set "{DID_EXT_AIDING_POS: {...}}"`, which makes the SN-8318 EKF path testable with no GPX and no external hardware.
- **Two GRMC bits, not one shared bit.** `g_gpxDidToGrmcBit` is one-to-one and `getPeriodMultiPos()` derives the period slot from the DID, so a shared bit would collapse both data sets into one period slot with shared auto-disable.
- **Not added to `GRMC_PRESET_GPX_IMX`.** An IMX consuming these from the same GPX it already uses as its GNSS receiver would fuse one solution twice, so this stays opt-in.
- `DefaultPeriodMultiple()` intentionally unchanged — unlisted DIDs default to 1, correct for epoch-rate data.

## Documentation corrections

The `status` comments claimed `0=ECEF, 1=NED` while the enum defines `ECEF=1, NED=2, BODY=3` with 0 invalid, so anyone coding to the comment would have sent an invalid frame. The frame enum also referenced a `DID_EXTERNAL_AIDING` that does not exist; that commented-out define is now an explicit RESERVED note pointing at the consolidated SID-based message proposed in SN-7740. Also recorded two previously-undocumented, load-bearing facts for producers: `var[]` is consumed as NED covariance even though `pos[]`/`vel[]` are ECEF, and a zero `var` silently discards the observation.

## Heads-up on flash layout

`GRMC_BIT_POS_COUNT` 28 → 30 grows `grmci_t` by 4 bytes and `grmci_sysCfg_t` from 376 to 392. `GPXFlashCfg_readConfig()` gates on `size == sizeof(grmci_sysCfg_t)`, so **any persisted GRMC config self-invalidates to defaults on upgrade** — needs a release note. Verified on hardware that persistence round-trips correctly at the new size.

## Verification

cltool builds; IMX-5 firmware builds. Both DIDs appear in `cltool -dids`. Mapper field resolution tested with positive and negative controls: all five field names resolve on each DID, an unknown name is rejected, and `pos` is correctly rejected on the VEL DID, confirming the two mappings are distinct. Full hardware results on SN-8472.

## Merge order

This first, then `inertialsense/imx` (is-common), then `inertialsense/is-gpx`.

### DIFF
--- a/src/data_sets.h
+++ b/src/data_sets.h
@@ -172,9 +172,11 @@ typedef uint32_t eDataIDs;
 #define DID_CAL_MOTION_GYR              (eDataIDs)103   /**< INTERNAL USE ONLY (sensor_mcal_group_t) */
 #define DID_CAL_MOTION_ACC              (eDataIDs)104   /**< INTERNAL USE ONLY (sensor_mcal_group_t) */
 #define DID_CAL_MOTION_MAG              (eDataIDs)105   /**< INTERNAL USE ONLY (sensor_mcal_group_t) */
-#define DID_EXT_POS                     (eDataIDs)106   /**< (ext_pos_t) External position observation */
-#define DID_EXT_VEL                     (eDataIDs)107   /**< (ext_vel_t) External velocity observation */
-// #define DID_EXTERNAL_AIDING             (eDataIDs)106   /**< (external_aiding_u) External aiding information. */
+#define DID_EXT_AIDING_POS              (eDataIDs)106   /**< (ext_aiding_pos_t) External aiding position observation, input to the INS/EKF */
+#define DID_EXT_AIDING_VEL              (eDataIDs)107   /**< (ext_aiding_vel_t) External aiding velocity observation, input to the INS/EKF */
+// RESERVED: a future consolidated external-aiding message (DID + sub-ID + SID-dependent payload,
+// covering position/velocity/speed/heading/attitude/IMU/config with full covariance) was proposed
+// under SN-7740. The two flat DIDs above are the initial subset; see SN-7740 before adding more.
 
 #define DID_EVENT                       (eDataIDs)119   /**< INTERNAL USE ONLY (did_event_t)*/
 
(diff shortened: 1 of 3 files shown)


Hephaestus-Maximus (approved):
## Summary

Reviewed independently rather than taking the (unusually thorough) PR description and Jira write-up at face value. Checked out the branch, ran `git submodule update --init --recursive`, and verified each of the three pieces this PR claims to do.

## What I verified

**Rename is complete — no orphaned references.** Grepped the entire repo (not just the diff) for `DID_EXT_POS`, `DID_EXT_VEL`, `ext_pos_t`, `ext_vel_t`, `eExternalAidingStatus`, `EP_STATUS_FRAME_*` — zero hits outside this PR's own changes. A partial rename is the easiest way for this class of change to silently break a build; it didn't happen here.

**Build + full test suite, independently run.** Built `tests/IS-SDK_unit-tests` from source (via `tests/CMakeLists.txt`, working around the known self-referential `include_is_sdk_find_library.cmake` quirk in this sandbox — configuring under `tests/build2` mirrors a build dir at the repo root and the top-level `all` target can't descend into it, but building the `IS-SDK_unit-tests` target directly works fine and the underlying `InertialSenseSDK` static lib already links correctly). Full run: **461 passed, 4 skipped (pre-existing, environment-gated — live-capture/E2E-soak tests, unrelated to this diff), 0 failed.** `ISDataMappings`/`DidMappingTest` suites specifically all green.

**Mapper registration is correct.** Checked `DataMapper::AddArray`'s actual signature (`name, member, type, arraySize, units, description, flags, conversion`) against both new calls — argument order, types (`DATA_TYPE_F64` for the `double pos[3]`, `DATA_TYPE_F32` for `vel`/`offset`/`var`), and flags (`DATA_FLAGS_FIXED_DECIMAL_*`, `DATA_FLAGS_DISPLAY_HEX`) all resolve to real, correctly-used definitions elsewhere in the file. `offsetsExtAidingPos` byte-flip offsets are correct for the `PACKED` struct layout (two `uint32_t` then three `double`s → offsets 8/16/24).

**GRMC bits are structurally sound.** New positions 28/29 are the next sequential values with no gaps or collisions (`GRMC_BIT_POS_COUNT` auto-follows to 30), `g_gpxDidToGrmcBit`/`g_gpxGRMCPresetLookup` entries are additive with no pre-existing entries overwritten, and `GRMC_PRESET_GPX_IMX` genuinely does not reference either new bit — confirmed the opt-in claim by reading the preset macro directly rather than trusting the description.

**Documentation corrections check out.** The old `status` field comments really did say `0=ECEF, 1=NED` while `eExtAidingFrame`/`eExternalAidingStatus` defines `ECEF=1, NED=2, BODY=3` — confirmed against the enum both before and after. The fix is accurate.

**Flash-layout disclosure is exactly the right shape.** `GRMC_BIT_POS_COUNT` sizes `grmci_t.periodMultiple[]` directly (confirmed via grep), so the 28→30 bump is a real, if modest, size change. This PR calls it out explicitly as a release-note item rather than leaving it implicit — the right way to handle this class of change (a related SDK PR I reviewed a couple of days ago had exactly the opposite problem: a persisted-config renumbering with no callout at all).

## Nitpick (non-blocking)

`status` is mapped with `DATA_FLAGS_DISPLAY_HEX` for a 3-valued enum (1/2/3) — hex vs. decimal makes no visual difference for single-digit values, so this is purely cosmetic and not worth holding up the PR over.

## Description / ticket

5/5 — the PR description matches the code precisely, including the two design decisions (two GRMC bits not one; not in the IMX preset) and the flash-layout heads-up. No edit needed. Jira (SN-8317) matches, target branch (`SN-8318_external_aiding`) matches the stated merge-order plan across all three repos.

## Verdict

**Approve.** Clean, well-documented, independently verified (grep for orphaned renames, full test suite green, structural checks against actual template/enum definitions rather than trusting the diff or description). No blocking findings.

shefft (approved):

src/data_sets.h
```diff
@@ -172,9 +172,11 @@ typedef uint32_t eDataIDs;
 #define DID_CAL_MOTION_GYR              (eDataIDs)103   /**< INTERNAL USE ONLY (sensor_mcal_group_t) */
 #define DID_CAL_MOTION_ACC              (eDataIDs)104   /**< INTERNAL USE ONLY (sensor_mcal_group_t) */
 #define DID_CAL_MOTION_MAG              (eDataIDs)105   /**< INTERNAL USE ONLY (sensor_mcal_group_t) */
-#define DID_EXT_POS                     (eDataIDs)106   /**< (ext_pos_t) External position observation */
-#define DID_EXT_VEL                     (eDataIDs)107   /**< (ext_vel_t) External velocity observation */
-// #define DID_EXTERNAL_AIDING             (eDataIDs)106   /**< (external_aiding_u) External aiding information. */
+#define DID_EXT_AIDING_POS              (eDataIDs)106   /**< (ext_aiding_pos_t) External aiding position observation, input to the INS/EKF */
```
aabogdanov, line 175: Do you guys feel like having AIDING in the DID? I named if originally DID_EXT_POS to be consistent with DID_GNSS_POS and DID_GNSS_VEL
In my mind I had this: *DID* *SOURCE* *OBSERVATION_TYPE*
So I though of source as "external"

src/data_sets.h
```diff
@@ -3313,34 +3319,34 @@ typedef struct PACKED
...
-/** @brief External aiding status bitflags, used with ext_vel_t.status (DID_EXTERNAL_AIDING). Reports the frame of measurement for the external aiding sensor. */
-enum eExternalAidingStatus
+/** @brief Frame of measurement for an external aiding observation, held in the low nibble of ext_aiding_pos_t.status / ext_aiding_vel_t.status (DID_EXT_AIDING_POS, DID_EXT_AIDING_VEL). Note 0 is not a valid frame. */
+enum eExtAidingFrame
 {
-    EP_STATUS_FRAME_MASK    = 0x0000000F,  //!< Mask for the frame of measurement
-    EP_STATUS_FRAME_ECEF    = 1,           //!< ECEF frame
-    EP_STATUS_FRAME_NED     = 2,           //!< NED frame
-    EP_STATUS_FRAME_BODY    = 3            //!< Body frame
+    EXT_AIDING_FRAME_MASK    = 0x0000000F,  //!< Mask for the frame of measurement
```
aabogdanov, line 3325: I think status meant to have other information besides frame in the future

src/data_sets.h
kylemallory: Yeah, I think the feeling was that even though these are generally position messages, they are primarily AIDING messages, and in the case of POS, you're AIDING POS.  So, by itself its not an EXT POS.  But, we can discuss more on monday if you like.

tmcd249 (approved):

PR #1270, opened 2026-08-12
SN-3918: fwUpdate flow control, adaptive chunk pacing, and u-blox target/instance handling

kylemallory:
Paired with imx#1656, which carries the driver and engine. Held as draft so the protocol addition is reviewed alongside its only consumer.

Jira: SN-3918.

## What

**1. Device-driven flow control** (`FirmwareUpdate.{h,cpp}`) — the substantive addition.

A device that buffers image chunks can run out of room while it works through what it already holds.
The only answer the protocol gave it was to refuse each arriving chunk with a retryable status, which
the host resends immediately — so a device that is merely *busy* gets hammered instead of given room,
starving the very work that would free it.

- `REASON_TOO_FAST` joins `resend_reason_e`; `CHUNK_ID_PAUSE` (`0xFFFF`) is the sentinel chunk id.
- A `REQ_RESEND_CHUNK` carrying both suspends sending indefinitely. Any later request naming a real
  chunk resumes from there. `REASON_TOO_FAST` with a real chunk id is the milder form — send that
  chunk next, after `FLOW_CONTROL_DELAY_MS`.
- Device side: `fwUpdate_sendRetryFrom()`, `fwUpdate_requestChunkPause()`, `fwUpdate_resumeChunks()`.
- Host side: `chunks_paused` / `chunks_hold_until`, honoured in `fwUpdate_sendNextChunk()`.

**No wire-format change.** These are new semantics on an existing message and an existing enum, so a
host that predates them treats a pause request as an ordinary resend.

Measured on a ZED-F9P behind an IMX-5 — without this, a 2646-chunk image stalls at 15.3% in a resend
storm; with it the same transfer completes 2646/2646.

`fwUpdate_sendRetryFrom()` is also how a device asks to go **backwards**, to a chunk it has already
consumed but needs again. The host has always honoured whatever chunk id it is given
(`fwUpdate_processMessage()` assigns `next_chunk_id` directly); only the device-side helper hardcoded
`last_chunk_id + 1`. Nothing in-tree needs the rewind yet, but it removes the reason a device would
otherwise retain consumed image bytes in RAM.

**2. Model-agnostic fwUpdate target.** Renames `TARGET_UBLOX_F9P*` to `TARGET_UBLOX*`, value `0x110`
unchanged. One target covers every u-blox receiver; the driver distinguishes models at runtime from
UBX-MON-VER rather than allocating a target id per model.

- `fwUpdate_getTargetName()` — reports `u-blox` with `.1`/`.2` instance suffixes, mirroring CXD5610.
- `data_sets.h` — generalizes the `IS_HDW_GNSS_UBLOX` comment, which described the type as F9P-only.

**3. `UBX1` / `UBX2` manifest tokens** (`ISFirmwareUpdater.cpp`). `TARGET_UBLOX__1` and `__2` existed
but no manifest token reached them, so a package could only address the base target. On a product with
two u-blox receivers — an IG1-G2 carries two ZED-F9Ps — both driver instances then answer the same
target and the communications fan-out decides which one wins, leaving the second receiver unreachable.
Mirrors the existing `CXD1`/`CXD2`. Bare `UBX` resolves to instance 1, which is the receiver it reached
in practice anyway.

Confirmed on hardware: `UBX1` returns GNSS1's version, `UBX2` updates GNSS2.

**4. Report u-blox receivers by model designation.** `ISDevice::getName()` renders u-blox
`hardwareVer[]` as the designation an operator recognizes — `UBX-X20D`, `UBX-F9P`, `UBX-M8` — instead
of a numeric triple. u-blox receivers expose no meaningful hardware version, so the driver uses those
bytes to carry model identity: `[0]` = product-line letter as `('X' - 'A')`, `[1]` = model number,
`[2]` = variant letter. Self-describing, so new models need no table. Non-u-blox hardware types keep
the existing numeric formatting untouched.

```
Received device version: SN0 (UBX-X20D), fw2.0.0.1 b00000.1
```

## An hdwId limitation worth knowing about

This identity lives only in the **byte** form. The packed 16-bit `hdwId` cannot express it:
`HDW_MAJOR__MASK` (`0x03C0`) is 4 bits, so a product-line letter past `'P'` truncates —
`('X' - 'A') = 23` becomes 7 and reports `UBX-7.20`. Observed on hardware during development.

- There is deliberately **no** packed `IS_HDW_UBLOX_X20` constant; it cannot be represented.
- Callers discriminating u-blox models must compare `hardwareVer[]`, not the packed id.
- `IS_HDW_UBLOX_F9P` is declared but never consumed, so nothing depends on the packed form today.

If the packed form should represent these, the field widths need revisiting — flagging, not deciding.

## On the target rename

Audited before renaming: the old names had **no functional consumers** — only the enum itself and one
name string. `grep -rn 'TARGET_UBLOX_F9P'` is empty after this change. The numeric value is unchanged,
so the wire protocol is unaffected; only a source-level reference to the old identifier would need
updating.

**5. Adaptive chunk pacing, and three fixes in the same handler.**

`chunkDelay` was a fixed per-chunk sleep, raised from 5 ms to 25 ms by `initialize()` — which is
reached only from `processPackageManifest()`. So the same image took 5x longer through a package than
through `-uf-cmd`, for no reason connected to the device. That delay was also slow enough that no
device ever had to ask for a resend, which left the flow control above unused.

**How it works.** Start at 5 ms. On `REASON_TOO_FAST` the new pace is measured rather than guessed:
chunks sent since the target last complained, divided by elapsed time, is the ms/chunk it actually
sustained — pace 10% above that. Other resend reasons carry no measurement, so they raise the delay by
half again (older devices cannot send `TOO_FAST`, and a plain resend is the only rate signal they
have). 5% comes back off for every 100 chunks that go out clean. Bounded 5–50 ms, and reset per upload
so one slow target cannot tax the next.

**Performance** — same images, same hardware, before vs after:

| | chunks | before (fixed 25 ms) | after (adaptive) |
|---|---|---|---|
| ZED-F9P via IMX-5 | 2646 | 73.7 s | **22.3 s** |
| GPX-1 firmware | 505 | ~12.6 s | **3.3 s** |

After the change the F9P transfer costs 7 resend requests (0.26% redundant chunks); the GPX-1 needs
none, since `GPXFirmwareUpdater` blocks on flash and self-paces. ~22 s is the floor the F9P's flash
sustains — going faster means overrunning it and paying for re-sent chunks.

Also here, because the same handler owns them:

- **The give-up guard never worked.** `last_resent_chunk` was compared but never assigned, so it stayed
  0 and "too many resends of one chunk" could only ever fire for chunk 0. Now assigned; tolerance
  raised 10 → 25 since it can actually trigger.
- **`REASON_TOO_FAST` is excluded from that accounting.** A target may ask for room many times in one
  transfer, and the pause form carries the same chunk id every time — either would count a busy device
  toward a write-failure verdict.
- **`resendChunkDelay` removed.** It set `nextChunkSend` and then immediately called
  `fwUpdate_sendNextChunk()`, which overwrote it with `chunkDelay`. It had no effect.

**6. Progress messages were never NUL-terminated on the wire.** `msg_len` came from `vsnprintf()`'s
return, which excludes the terminator, so consumers reading those bytes as a C string ran past the
message (`...at 921600 baudKx`, `retries 0I␦␦Ib`). `msg_len` now counts the terminator, and both
readers bound with `strnlen()` — devices already in the field will keep sending unterminated strings,
and reading past the message is the real defect. The host also no longer passes device-supplied text
as a `printf` format string.

> **Note for whoever reviews the pacing change:** it is only safe because the CXD5610 and Septentrio
> drivers can now signal backpressure (imx#1656). Both previously relied on the fixed 25 ms, and the
> CXD in particular could not report a short write at all. Those two commits are prerequisites, not
> parallel improvements.
>
> **Resolved (SN-8549).** The prerequisite driver work has landed on this branch in `is-common`
> (`cpp/hdw-src/hw-libs/devices/gnss/Cxd5610.cpp`), and the CXD5610 hold is now validated on
> hardware. Measured on an IG2 (IMX-5 SN62797, GPX-1 SN928401213), forced update of all five CXD
> images on both receivers: 22 sequence errors to 0, 80.7s to 52.2s, and zero sequence errors across
> 90 consecutive package updates. This PR is no longer blocked from GPX-1 hardware.
>
> Note the driver lives in `is-common`, shared by `is-gpx` and `is-imx` — not in `is-gpx` as this
> note originally said.

## Review notes

- The flow-control gate returns the remaining chunk count rather than `0` when held, so a caller cannot
  read "paused for a moment" as "nothing left to send".
- A pause is not an error and does not end the session: `resend_count` is untouched by the pause form,
  since a device asking for room is not a device failing.
- `fwUpdate_resetEngine()` clears both flow-control fields, so a new session never inherits a hold.


aabogdanov (approved):

aabogdanov (approved):

shefft (approved):

tmcd249 (approved):

Hephaestus-Maximus (commented):
In-depth review of the full diff against `3.1.0-rc` (327/-31 across 6 files, 8 substantive commits). Cloned, checked out, traced the protocol-layer/app-layer split (`ISFirmwareUpdater` privately inherits `FirmwareUpdateHost`) end-to-end rather than reading the two files in isolation. CI is green (Linux + Windows). No unit tests were touched by this PR, so I read the existing `tests/test_ISFirmwareUpdate.cpp` harness to check what's actually exercised.

**Verified correct, traced through both sides of the wire:**
- Flow control (`REASON_TOO_FAST` / `CHUNK_ID_PAUSE`): confirmed the pause form leaves `next_chunk_id` untouched and doesn't touch `resend_count`; confirmed `fwUpdate_resumeChunks()` uses `REASON_NONE` so resuming doesn't also incur `FLOW_CONTROL_DELAY_MS`; confirmed `fwUpdate_sendNextChunk()` returns the remaining count (not 0) while held, so a caller can't misread a pause as "done."
- The give-up guard fix is real: grepped the whole file, `last_resent_chunk` had exactly one write site before this PR (its own `initialize()` reset to 0) and zero others, so it could never register a repeat — matches the PR body's claim precisely. It's now assigned in the "different chunk" branch, which is the correct fix.
- `REASON_TOO_FAST` is correctly excluded from that same give-up accounting (returns early, never reaches the `last_resent_chunk`/`resent_chunkid_count` block) — confirmed a busy-but-healthy target can't be counted toward `ERR_FLASH_WRITE_FAILURE`.
- Adaptive pacing math (`derivePaceFromWindow`, `creditChunkProgress`): worked through the rounding/floor/ceiling logic by hand, including the "percentage rounds to zero below ~20ms so force a 1ms step" case — internally consistent, and the documented trade-off (drift-then-correct, ~0.26% resend overhead) matches the measured numbers given in the PR body.
- `TARGET_UBLOX_F9P*` → `TARGET_UBLOX*` rename: grepped for the old names post-rename, none remain. `IS_HDW_UBLOX_F9P` (packed form): confirmed genuinely unreferenced anywhere else, so the claim "nothing depends on the packed form today" checks out.
- u-blox model-designation rendering in `ISDevice::getName()`: hand-decoded all three examples from the PR body (`UBX-X20D`, `UBX-F9P`, `UBX-M8`) against the `('A'+hardwareVer[0])` / `hardwareVer[1]` / `('A'+hardwareVer[2])` encoding — matches exactly, and it operates on the raw `hardwareVer[]` bytes rather than the packed `hdwId`, so the 4-bit-truncation caveat flagged in the PR body doesn't actually apply to this code path (that caveat is about a different, not-yet-written packed-encoding path) — worth being explicit that these are two different concerns so a future reader doesn't conflate them.
- Nice incidental hardening, not called out in the PR body: the progress-message logging change is a fix for two separate defects at once. Before, `LOG_FWUPDATE_STATUS(level, message)` passed device-supplied text as the *format string* — a corrupted or malicious device could put `%n`/`%s` etc. in a progress message and have the host interpret it. Now it's passed as an argument to `"%.*s"`. Worth mentioning explicitly since it reads like a side effect of the NUL-termination fix rather than a deliberate hardening, and someone re-reading history later should get credit for (and not accidentally revert) both defects at once.

**Findings:**

1. **(Worth fixing) Give-up-guard counters aren't reset per-upload, unlike the pacing state.** `initializeUpload()` now resets `chunkDelay`/`chunksSinceResend`/`paceWindowStartMs`/`paceWindowChunks` specifically so "one slow target cannot tax the next" (your words, in the header comment) when a package updates multiple targets in sequence. `last_resent_chunk`, `resent_chunkid_count`, and `resent_chunkid_time` are session-scoped by the same logic, but they're only reset in `initialize()` — which runs once per *package*, not once per *target's upload*. So if target A's session ends with `resent_chunkid_count` sitting a few away from `MAX_SAME_CHUNK_RESENDS`, and target B's first resend happens to land on the same `chunk_id` target A last resent, target B inherits progress toward the give-up threshold it didn't earn. Narrow (needs a same-package multi-target run plus a coincidental chunk-id match) and fails safe (aborts a session, doesn't corrupt data), but it's the same class of bug the pacing reset was just added to fix, in the same function, one paragraph away. Suggest adding those three fields to the same reset block in `initializeUpload()`.

2. **(Minor edge case, adjacent to finding 1's territory) `last_resent_chunk` defaults to 0, and 0 is a valid chunk id.** If the very first resend of a session happens to be for chunk 0, `chunk_id == last_resent_chunk` is true by coincidence (not because it's actually a repeat), so `resent_chunkid_count` starts at 1 via the wrong branch instead of 0 via the right one. Self-corrects on the next distinct resend, so it's not worth blocking on — just flagging since it's exactly adjacent to the bug this PR just fixed, and a sentinel like `-1`/`UINT16_MAX` (unpaused chunk ids don't reach that range in an 8-chunk test, but could in principle) would remove the ambiguity if you're touching this code again for finding 1.

3. **(Test coverage) None of the new logic has test coverage, and the harness already has what's needed to add it cheaply.** `REASON_TOO_FAST`/`CHUNK_ID_PAUSE`, the pacing derivation, and the give-up-guard fix have zero references in `tests/test_ISFirmwareUpdate.cpp`. `exchange__req_resend` already demonstrates the exact pattern needed — it mucks with the exchange buffer to relabel a chunk id and verifies the resulting `REQ_RESEND_CHUNK` handling — the same trick would let a test inject a `REASON_TOO_FAST`/`CHUNK_ID_PAUSE` response and assert `fwUpdate_sendNextChunk()` returns non-zero-but-doesn't-send, then a resume that lifts it. Separately, `pack_unpack__progress` sets `msg_len` directly rather than going through the changed `fwUpdate_sendProgressFormatted()`, so the actual terminator/format-string fix has no regression test either. Given this is firmware-update reliability logic (a regression here risks stalled or wrongly-aborted field updates), I'd treat this as the most valuable follow-up before/soon after merge.

4. **(Nit)** `cmd_SetTarget`'s doc comment and its "Invalid Target" error message both list `UBX1, UBX2` but not bare `UBX`, even though the code accepts it (`else if (targetName == "UBX") ...`). Someone hitting the error message won't discover the alias exists.

Not blocking on any of the above — CI is green and I didn't find a hard correctness break, just one real-but-narrow state-carryover gap (1), a benign edge case (2), and a coverage gap I think is worth closing (3) given the safety profile of this code path. The PR body is already unusually clear about the two things it explicitly isn't claiming (CXD5610 hold unvalidated on hardware, Septentrio/CXD driver defects tracked on `is-gpx`) — appreciate that this is called out rather than implied.

tmcd249 (approved):

aabogdanov (approved):

Hephaestus-Maximus:
Addressed all four findings from the review above:

1. **Fixed** (`0cab6ef9`) — `last_resent_chunk`/`resent_chunkid_count`/`resent_chunkid_time` are now reset in `initializeUpload()`, same as the pacing state, so one target's near-miss count can no longer carry into the next target's give-up threshold.
2. **Fixed** (`0cab6ef9`) — added an explicit `hasResentChunk` flag rather than reusing a sentinel value. Worth noting: the natural `uint16_t` sentinel (`0xFFFF`) turned out to already be claimed by `CHUNK_ID_PAUSE`, so overloading `last_resent_chunk` itself would've traded one ambiguity for another — a flag avoids that entirely.
3. **Partially addressed** (`7e18feac`) — added `exchange__flow_control_pause_resume`, covering the `REASON_TOO_FAST`/`CHUNK_ID_PAUSE` pause-then-resume mechanism end to end (asserts a pause leaves `next_chunk_id` untouched and sends nothing, and a resume both lifts it and rewinds correctly). The give-up-guard and pacing-derivation logic (finding 1/2's fields) remain untested: they live in `ISFirmwareUpdater`, which this test file's harness doesn't exercise — `ISFirmwareUpdateTestHost` subclasses `fwUpdate::FirmwareUpdateHost` (the protocol layer) directly, with its own separate handler stubs, not `ISFirmwareUpdater` (the app layer). Closing that gap needs a new `ISFirmwareUpdater`-level test harness (mocking a `port_handle_t`, since it implements the protocol callbacks itself rather than deferring to a subclass) — a materially bigger lift than extending this file, so flagging it explicitly rather than claiming coverage this change doesn't add.
4. **Fixed** (`0cab6ef9`) — doc comment and the "Invalid Target" error message now both list `UBX` alongside `UBX1`/`UBX2`.

`IS-SDK_unit-tests`: 586 passed, 4 pre-existing skips, 0 failures.

kylemallory:
## SN-8549: the pacing change is now validated on GPX-1, and one interaction it needed

Adds `edecc884f`. The blocker note in the description is cleared — the prerequisite driver work landed in `is-common` on this branch (imx#1656), and the CXD5610 hold is validated on hardware.

### The interaction that had to be fixed

Once the CXD5610 holds at every frame boundary — which it must, since a frame's last bytes and the next frame's first bytes arrive in the *same* chunk — the pace estimator turns against itself:

`fwUpdate_handleResendChunk()` fed **every** `REASON_TOO_FAST` to `derivePaceFromWindow()`, including the `CHUNK_ID_PAUSE` form. The hold's wait is folded into the measured ms/chunk and then added back on top as the new delay, so each hold raised the pace and the next hold measured the raised pace. It ratchets to `CHUNK_DELAY_MAX_MS` within ~70 chunks and never decays, because holds arrive far more often than `CHUNK_DECAY_CHUNKS`. Measured: 5 ms drifting to 18 ms — **slower than the fixed 25 ms this PR replaces.**

The pause form carries no rate to learn from. A target that holds and then names a chunk to resume from is running its own flow control, not asking for a different steady-state rate. So it is now exempt.

### Second half: stragglers are not sequence faults

A hold does not take effect instantly — whatever the host already put on the wire arrives ahead of the chunk the device is waiting for. Answering those with `REASON_INVALID_SEQID` made the host slow down for a fault that never happened. `FirmwareUpdateDevice` now drops them (`chunks_held`), bounded by `MAX_DROPPED_STRAGGLERS` so a genuinely lost chunk is still re-requested rather than both ends falling silent.

`fwUpdate_sendRetry()` now delegates to `fwUpdate_sendRetryFrom()` so the record of what has been asked for lives in one place — **byte-identical wire output**.

### Also

- **Per-target starting delay _and_ decay floor** (`chunkDelayFloor`). Raising the start alone was undone by the decay walking it back to the global base over the first few hundred clean chunks. The CXD value derives from the frame (`startingChunkDelay(target, chunk_size)`), so it stays correct if the negotiated chunk size ever changes.
- Holds/resumes moved to `log_more_debug`; a hold outstanding longer than 2x the progress period *is* reported, since that one is the reason progress has stopped. Progress-stream volume: 2528 lines → ~480.

### Compatibility

No new message types, no `payload_t` changes, no `msg_types_e` additions. `REASON_TOO_FAST` and `CHUNK_ID_PAUSE` do not exist before `64858eb32` (this branch), so firmware that predates them cannot send either and **none of the host-side branches touched here are reachable for it**. The only exchange-level change is a device sending *fewer* `REQ_RESEND_CHUNK` messages, and only in firmware that pauses.

Tests: 11/11 `ISFirmwareUpdate.*` pass, including the new `exchange__flow_control_pause_resume` from `7e18feac8`.

### One thing tried and reverted, with numbers

`FWUPDATE__MAX_CHUNK_SIZE` at 1024. It went 1024 → 512 in `f7257159e` ("Adds GPX Firmware Updates through IMX and SDK"), whose message says plainly *"Reduced MAX_CHUNK_SIZE to 512 (to play nice with IMX)"* — an accommodation, not a protocol limit. The transport allows it (`PKT_BUF_SIZE` 2048 vs a 1116-byte payload) and the IMX relay is not a constraint either, since `fwUpdate_unpackPayload()` copies only the fixed header and forwards the raw pointer.

But measured on hardware it is **worse**: 8 lost chunks in 1920 against none at 512, and the backoff each loss triggers costs more than halving the round trips saves — 66 s vs 52 s. Reverted, with the measurement recorded at the definition so the next person doesn't repeat it.

Useful byproduct: the chunk-size renegotiation was exercised for real. A host asking 1024 of 512-capped firmware got `ERR_MAX_CHUNK_SIZE`, halved, and completed normally — so `ISFirmwareUpdater.cpp:332` works as intended.

### Where the time actually goes, for anyone tuning this further

`FINALIZING → FINISHED` is **14.3 s** across the 10 images and identical in every run — the CXD writing its own NVM, untouchable from here. Transfer is bounded below by 21.9 s (2,022,528 bytes at 921600 baud). The ~16 s above that is host-path per-chunk overhead, ~2.6 ms on top of the nominal delay across 3950 chunks. That is why neither 5 ms nor 7 ms nor 1024-byte chunks moved the wall clock: the pacing is masked by the path, not by the receiver.

shefft (approved):

PR #1274, opened 2026-08-14
SN-8493: classify the NTRIP caster's response instead of assuming success

kylemallory:
## Problem

`NtripCorrectionService::negotiateMountPoint()` read the caster's response headers but never inspected the status line, returning `true` whenever bytes arrived without a timeout. Every client-visible refusal therefore presented as a healthy connection that simply never carried corrections — rejected credentials and an unknown mount point were indistinguishable from a working stream, and a caller had no reason to report because the reason was discarded here.

The old body also parsed `Content-Type:` into a variable that was never read, and drained the response by `Content-Length` — which meant that when a caster answered with its source table, the loop dutifully consumed the whole table and still returned success.

## Response handling

Covers both dialects, since we send an HTTP/1.1 request and a caster may legitimately answer in either:

| Response | Meaning |
|---|---|
| `ICY 200 OK` | Ntrip 1.0 success — **not** a valid HTTP status line, so a strict HTTP parser would wrongly reject a good link |
| `HTTP/1.x 200 OK` | Ntrip 2.0 success |
| `SOURCETABLE 200 OK` | mount point unavailable — the spec has a caster answer an unavailable mount point with its source table, not a 404 |
| 200 + `Content-Length`, or `gnss/sourcetable` | also a source table: a correction stream is endless and never carries a length, while a source table always does. **This is the case that actually bit us**, and it covers Ntrip 1.0 serving its table as the too-generic `text/plain` |
| `401` / `403` / `407`, and `ICY 401 Unauthorized` | credentials — some casters answer a poorly-formed v1 request with the ICY-prefixed form |
| `ERROR - Bad Password` | Ntrip 1.0's custom status code |
| `404`, other codes | passed through with the reason phrase |
| anything unrecognised | reported verbatim (truncated) rather than accepted, so a future caster quirk stays visible instead of silently succeeding |

The outcome is exposed via `getLastResult()` / `getLastResultString()` so a caller can explain the failure to a user. `classifyResponse()` is split out from the socket read, which keeps the decision — the part that has to be right across both dialects — directly testable.

## Also in this path

- **Credentials no longer reach any log line.** `casterUrl` retained the password in the clear and was passed to `log_debug`; a redacted `casterUrlSafe` is used instead.
- **Removed `printf("%s\n", buffer)`** on an uninitialised 512-byte array that `portReadLineTimeout()` does not NUL-terminate — a potential over-read, and it dumped caster response headers to stdout.
- The header loop now **stops at the blank line** rather than draining by `Content-Length`, so it cannot eat the leading bytes of a live stream.
- `connect()` **closes the socket when negotiation fails**, so it neither leaks the port nor leaves `isConnected()` claiming success.
- Replaced a local trim helper with the existing `utils::trim_copy()`.

## Tests

19 new cases in `tests/test_NtripCorrectionService.cpp` covering every response form above, plus two guards worth calling out:

- `text/plain` **without** a length must still be accepted — protects against over-eager source-table detection breaking a caster that mislabels a working feed
- an unrecognised reply is truncated in the message — a caster answering with a whole HTML error page must not paste it into a UI label

All 19 pass, re-verified after merging the latest `3.1.0-rc` (which touched `ISStatusDecode.cpp` and its tests).

## Verification

Beyond the unit tests, the full chain was exercised against a local fake caster: a `401` now surfaces as *"The caster rejected the credentials (401 Unauthorized)"* in EvalTool's Correction Input panel, where the old code reported a healthy connection. An accept-then-silence caster and a valid CRC24Q RTCM3 stream were also driven through to confirm the success path and the connected-but-no-data path.

## Notes

- The consuming EvalTool changes live on `evaltool-ui-improvements` in `is-common` and are **not** part of this PR. That branch pins `is-common/SDK` at this commit, so this needs to merge first, after which the submodule pointer gets re-bumped to the merged SHA.
- Unrelated and still open: `HDW_BIT_FAULT_EKF_NOT_INITIALIZED` has no entry in the `hdwBitStatus` decode table (SN-8502). Confirmed still absent after this merge; not touched here.

Jira: SN-8493


aabogdanov (approved):

Hephaestus-Maximus (approved):
Independently verified this end-to-end rather than relying on the existing approval:

**Logic trace.** Hand-traced every branch of `classifyResponse()` against each response form from SN-8493: `ICY 200 OK` / `HTTP/1.x 200 OK` success, explicit `SOURCETABLE` status line, a 200 whose `Content-Type` contains `sourcetable`, a 200 carrying any `Content-Length` (the actual incident trigger — a live correction stream never sends one), `ERROR - Bad Password` / 401 / 403 / 407 → unauthorized, 404 → mountpoint-unavailable, other codes → caster-error, and unparseable status lines → unrecognized (fails closed rather than assuming success). All branches check out; the `ICY 200 OK` literal-prefix special case is redundant with, but consistent with, the generic `ntripParseStatus` + `switch(200)` path.

**Credential handling.** `grep -n "casterUrl\b"` across the changed files confirms the raw (password-bearing) `casterUrl` is assigned once, in `connect()`, and never read, logged, or displayed anywhere else — `casterUrlSafe` (userinfo replaced with `[hidden]`) is what actually gets logged. `requestMsg` (base64 Basic-Auth) is only ever handed to `portIsOpened`/`portWrite`, never logged. No leak path.

**Response-head reading.** The rewritten loop in `negotiateMountPoint()` correctly treats line 0 as the status line and everything after as headers, extracting only `Content-Type`/`Content-Length`, bounded by `NTRIP_MAX_RESPONSE_LINES` and a per-line timeout — no unbounded read.

**Tests.** Built `IS-SDK_unit-tests` from a clean clone at the PR head (`b3ee13e`) and ran it:
- All 19 new `NtripResponse.*` cases pass, including the two that matter most for regression safety: `ContentLengthOnA200IsMountpointFailure` (the actual bug this PR fixes) and `PlainTextWithoutLengthIsNotRejected` (guards against over-eager detection).
- Full suite: 562 tests, 558 passed, 4 skipped (pre-existing hardware/live-capture fixtures and an opt-in soak test, unrelated to this change) — no regressions.

**Minor, non-blocking nitpick:** `setResult()`'s Doxygen comment (`NtripCorrectionService.h:162`) says "@return false always, so callers can 'return setResult(...)' on the failure paths," but the implementation returns `(result == NTRIP_RESULT_OK)` — true on success. The code is correct and used correctly at every call site; only the comment is inaccurate. Worth a follow-up doc fix, not worth blocking on.

No functional or logic defects found. PR description is accurate and unusually thorough (response-handling table, dialect notes, test count) — no changes needed there.

shefft (approved):

PR #1276, opened 2026-08-16
SN-8333: Fold ISComm_callback features into the ISComm example

Hephaestus-Maximus:
## Why

Rather than fix the standalone bug in `ExampleProjects/ISComm_callback` (a stale `serialPortOpen()` return-value contract — see SN-8333 for the root-cause trail), this folds `ISComm_callback`'s unique functionality into the primary `ISComm` example instead, so there's one comprehensive example rather than two overlapping/diverging ones.

## What changed

Both files are in `ExampleProjects/ISComm/`:

- **`ISCommExample.cpp`**:
  - Added a data-configuration example (Step 4): sets `DID_FLASH_CONFIG`'s `insRotation` via `is_comm_set_data()` — the specific snippet called out in the ticket.
  - Added a disabled-by-default (`#if 0`) example of saving persistent messages via `DID_SYS_CMD`/`SYS_CMD_SAVE_PERSISTENT_MESSAGES` (Step 7).
  - Fixed the port-validity check from `port == nullptr` to `portIsValid(port)`, matching the convention used by the newer example projects (`CustomVirtual`, etc.).
  - Restructured the `// STEP N` line comments into Doxygen-style `/** STEP N: ... */` blocks, and dropped an unused `protocol_nmea.h` include.
- **`README.md`**: substantially rewritten to match the newer documentation structure used elsewhere in `ExampleProjects` (Purpose/Design section, project-navigation guide, mermaid dependency/flow diagrams, Doxygen-comment note) rather than the older flat Step-by-step-only format.

## Scope note

`ExampleProjects/ISComm_callback/` itself has **not** been removed in this PR — it still exists on this branch and on `3.1.0-rc` unchanged. This PR only completes the "fold the functionality in" half of the ticket; removing the now-redundant callback project is a separate follow-up decision, not assumed here.

## Verified

- `git merge` (scratch, no conflicts) against the current `3.1.0-rc` tip — clean, no conflicts; the two touched files don't overlap with anything changed on `3.1.0-rc` since this branch forked.
- Built `InertialSenseSDK` from source and compiled `ISCommExample.cpp` directly against it (`g++ -std=c++17`) — clean compile, no warnings. Ran with no arguments — correct usage message, matches documented behavior.
- Have not tested against real hardware (no device attached in this environment) — the actual comms behavior (Step 4's rotation-config write, Step 8's receive loop) is unverified beyond compiling; whoever reviews with hardware on hand should exercise those paths.

Jira: SN-8333


shefft:

kylemallory (approved):
Looks great!  Nice work!

shefft (approved):

tmcd249 (approved):

PR #1278, opened 2026-08-19
SN-7604: GPX no longer write to USB port when port is closed.

tmcd249:
(GPX-1) (SN-7604, PR-1667) GPX no longer writes to the USB port when the port is closed.

### DIFF
--- a/src/core/base_port.h
+++ b/src/core/base_port.h
@@ -49,6 +49,8 @@
 #define PORT_ERROR__TIMEOUT             -7      //!< The port operation reported a timeout (could be read, write, open, etc)
 #define PORT_ERROR__INVALID_PARAMETER   -8      //!< The port was called with an invalid parameter
 
+#define EMBEDDED_USB_PORT_CLOSED        -99
+
 #define PORT_OP__READ               0x00        //!< A portLogger operation flag indicating a READ/RX was performed
 #define PORT_OP__WRITE              0x01        //!< A portLogger operation flag indicating a WRITE/TX was performed
 #define PORT_OP__OPEN               0x02        //!< A portLogger operation flag indicating a OPEN was performed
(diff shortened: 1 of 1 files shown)


Hephaestus-Maximus (approved):
Reviewed independently. This is the SDK half of the same SN-7604 fix in imx#1667 (GPX no longer writing to a closed USB port), which I already reviewed and approved this session — that review required tracing this exact `EMBEDDED_USB_PORT_CLOSED`/`portWrite()` change in full to confirm the fix was wired correctly end-to-end, so I'm carrying that verification over here rather than re-deriving it.

Note: diffing this PR against the raw tip of `origin/3.1.0-rc` initially showed a much larger, unrelated-looking diff (removal of `PORT_TYPE__CLASS_MASK` and `portClass()`) — that's because `3.1.0-rc` has moved forward since this branch's fork point via an already-merged, unrelated PR that added those. Diffing against the actual merge-base (`git diff $(git merge-base origin/3.1.0-rc pr1278) pr1278`) gives the true, minimal diff: exactly the 6 additions / 1 deletion GitHub reports.

**The actual change**, confirmed correct:
- Adds `#define EMBEDDED_USB_PORT_CLOSED -99` — checked it doesn't collide with any other `-99` return-code usage anywhere in `src/`.
- `portWrite()`'s stats bookkeeping now special-cases this sentinel: every other negative return still increments `txDataDrops`/`txBytesDropped` as before, but `EMBEDDED_USB_PORT_CLOSED` specifically does not — a closed port isn't a data-loss event, it's an expected "nothing to send to" state, so not counting it as a dropped-data fault is the right call.
- Cross-referenced against the actual consumer (GPX's `d_serial.c::serWrite()`, imx#1667): it returns exactly this constant when `zUsb_isOpen()` is false, before ever attempting `zUsb_write()`. The two PRs are a correctly-matched pair — this SDK PR's head is the exact SDK submodule commit imx#1667 pins to.

PR description correctly cross-references imx#1667's SN-7604/PR-1667 — no changes needed there.

Approving.

kylemallory (approved):

waltjohnson (approved):

aabogdanov (approved):

PR #1281, opened 2026-08-20
SN-8504: add imx detect define to global space

tmcd249:


Hephaestus-Maximus (approved):
Reviewed independently. Small, self-contained change (6 additions, 1 file) — part of SN-8504 (IMX-6 vs GPX-1 runtime hardware detection via the `GPX_nIMX_PIN` strap), which I'd previously reviewed and enriched on the Jira side (comment from 2026-08-18).

Adds `#define BOARD_IS_IMX 0` / `#define BOARD_IS_GPX 1` to the shared SDK header. Confirmed these exactly match the values of the same-named constants already defined locally in `cpp/hdw-src/hw-libs/mcu/STM32U5/startup/init.c` (imx repo) — this is a promotion of an existing local pair into shared/global scope, not a new or renumbered pair, so there's no value-shift risk for whatever currently reads `g_GPX_nIMX` against them.

Checked for naming collisions against the rest of `src/` — none. The added comment is a good, preemptive clarification: it explicitly distinguishes these from `eIsHardwareType`'s wire values (`IS_HARDWARE_TYPE_IMX=3`/`IS_HARDWARE_TYPE_GPX=4`), which is exactly the kind of mix-up a shared 0/1 pair named similarly to a wire-format enum could otherwise invite.

No description on the PR, but for a change this small and self-documenting (two `#define`s with an explanatory comment, linked to SN-8504 by title) I don't think one's needed here.

Approving.

kylemallory (approved):

shefft (approved):

aabogdanov (approved):

PR #1285, opened 2026-08-28
SN-6694 New example project with custom device, extending DeviceFactory

shefft:
Extending DeviceFactory - Demonstrates how to implement a new DeviceFactory to simplify discovery of custom ISDevice types.  This should build upon the Custom ISDevice example, by demonstrating how a custom DeviceFactory can be registered to look for a specific device or configuration, and then creating an extended ISDevice class (which I've called RobotDevice).

- Adds a new example project, ExampleProjects/RobotDevice, demonstrating how to subclass ISDevice and DeviceFactory for an application-specific device type (a "robot" device that caches and displays DID_INS_1 navigation data).
  - Uses the PortManager/DeviceManager auto-discovery pattern (as introduced in Simple_Discovery) rather than manual SerialPortFactory::bindPort(), so it also serves as a second, more elaborate discovery-based example alongside Simple_Discovery.
  - Ships with a full step-by-step README.md walkthrough that mirrors the STEP N comments left in the source files, intended as a tutorial for SDK consumers building their own custom device/factory pair.

New files
  - RobotDevice.h/.cpp — ISDevice subclass adding an ins_1_t insData member, a configure() method that stops existing broadcasts and subscribes to DID_INS_1, and an onIsbDataHandler() override that caches and pretty-prints incoming INS data via cInertialSenseDisplay.
  - RobotDeviceFactory.h/.cpp — DeviceFactory subclass (singleton, modeled on ImxDeviceFactory) that allocates a RobotDevice only for IMX-5 hardware IDs.
  - main.cpp — discovery loop using PortManager/DeviceManager, connects, configures, and steps the device in a loop until the port closes.
  - CMakeLists.txt, setup_debug.sh — standalone build/debug tooling for the example.
  - README.md — full tutorial covering project purpose, file layout, and an 8-step implementation walkthrough, plus Linux compile/run instructions.

Known gaps / follow-ups
  - Not yet wired into ExampleProjects/CMakeLists.txt — needs add_subdirectory(RobotDevice) added alongside the other examples, or it won't build as part of a full SDK/CI build. (Do we want this?)
  - Broadcast subscription in configure() uses a std::vector<std::function<bool()>> to demonstrate batching multiple subscribe calls, currently with only DID_INS_1 active; a second commented-out DID_SYS_PARAMS entry is left as a template for extending it.
  - onIsbAckHandler/onNmeaHandler overrides are stubbed out (commented) as extension points, not implemented in this pass.

  Test plan
  - [ ] Build standalone via the project's own CMakeLists.txt and confirm it links against a freshly-built libInertialSenseSDK.a
  - [ ] Run against a connected IMX-5 device with no arguments (auto port discovery) and confirm DID_INS_1 data prints continuously
  - [ ] Run with an explicit port argument (e.g. ./RobotDevice /dev/ttyACM0) and confirm the same behavior
  - [ ] Add add_subdirectory(RobotDevice) to ExampleProjects/CMakeLists.txt and verify it builds as part of the full ExampleProjects build  (Do we want this?)

-----------------------
EDIT 3 Sep
Renamed the project/files from CustomRobotDevice to RobotDevice, and RobotDeviceFactory etc


Hephaestus-Maximus (commented):
Reviewed this as the draft it is — since the PR body already lists its own known gaps and an unchecked test plan, I focused on giving genuinely useful feedback for finishing it up, not a merge-readiness verdict.

**Verified against the real SDK base classes, not just read for plausibility:**
- `CustomRobotDevice : public ISDevice` — `step()`, `onIsbDataHandler()`, and the virtual destructor all match `ISDevice`'s actual virtual signatures (`ISDevice.h`), so every `override` is legitimate.
- `CustomDeviceFactory::allocateDevice()` — signature matches `DeviceFactory`'s base exactly, and the `ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_5_0` exact-match pattern mirrors `GpxDeviceFactory`'s existing convention in `DeviceFactory.h` precisely (worth noting for your own understanding: this matches IMX-5 major.minor **exactly 5.0**, not "any IMX-5" — same narrowness as the shipped `GpxDeviceFactory`, so not a new problem, just worth knowing if a tester's unit is a different hardware revision).
- `BroadcastBinaryData`, `StopBroadcasts`/`SendRaw`'s `<0`-means-error convention, and `copyDataPToStructP`'s signature all check out against `ISDevice.h`/`ISComm.h`.
- I actually compiled the three new `.cpp` files (not just read them) via `-DIS_SDK_BUILD_FROM_SOURCE=ON` after pulling in the `src/libusb` submodule — all three compiled cleanly with no errors. (Final link failed on unrelated pre-existing `libusb` undefined references from `ISDFUFirmwareUpdater.cpp`, core SDK code untouched by this PR — that's a gap in my ad-hoc local build setup, not something in this diff.)

**Two small but real code issues, confirmed by inspection even though this build config didn't flag them:**
- `CustomRobotDevice::onIsbDataHandler()`: `uint32_t insStatus = insData.insStatus;` is assigned and never used afterward.
- `main.cpp`: `printf("...%lu...", dm.DeviceCount())` — `DeviceCount()` returns `size_t` (confirmed in `DeviceManager.h`), which is 32-bit under `%lu`'s expected `unsigned long` on 64-bit Windows/MSVC (LLP64) even though it's a 64-bit type there. Given the project's own `CMakeLists.txt` explicitly sets up an MSVC path, this is worth `%zu` (or a cast) rather than `%lu`.

**The bigger issue, and the main reason I'd hold off on marking this ready:** this reads like it was scaffolded from an existing "custom port factory" example template and not fully adapted — the tutorial layer (which the PR body calls out as the actual deliverable) has real leftover artifacts:
- `${BRIEF_DESC}` is left completely unfilled in the `@brief` line of all four new source files (`CustomRobotDevice.h/.cpp`, `CustomDeviceFactory.h/.cpp`) — will show up literally as `${BRIEF_DESC}` in any generated Doxygen docs.
- `CustomRobotDevice.cpp` and `CustomDeviceFactory.cpp` both have header-comment text that still says "Include the header file for child **port factory** class derived from parent **PortFactory.h**" — but neither file derives from `PortFactory`; they derive from `ISDevice`/`DeviceFactory`. `CustomDeviceFactory.cpp`'s own file-level comment even says "STEP 5: Implement the required functions for a **Port Factory**."
- In `main.cpp`, the doc-comment above STEP 7 (`device->connect()`, port connection) is a verbatim copy of the STEP 6 comment above it (singleton setup/registration) — so STEP 7 doesn't actually describe what STEP 7's code does. There's also a STEP-5 collision: `CustomDeviceFactory.cpp`'s file-level comment claims STEP 5, while `main.cpp` *also* has an internal "STEP 5" (setting the log level) — two unrelated things sharing a step number undermines the "guided numbered tour" the README explicitly promises.
- `README.md` still contains an entire struck-through paragraph (`~~Our custom PortFactory then binds...~~` through "...demonstrating the `base_port_t` interface") describing a different example's loopback-port content wholesale — it reads like someone flagged it for removal (hence the strikethrough) but never actually deleted it. There's also a stray `%%comment` line inside one of the mermaid diagrams, and the "Files" list is missing `CustomDeviceFactory.cpp` and `setup_debug.sh`.

**On your open question** ("wire into `ExampleProjects/CMakeLists.txt`?"): the two examples this PR explicitly says it builds upon — `Minimal_ISDevice` and `Simple_Discovery` — are both wired in there; `CustomPort` is a counter-example that isn't. Given this one is positioned alongside the wired-in pair, I'd lean toward adding it (mainly so a future SDK API change gets caught by CI instead of silently bit-rotting this example), but there's precedent either way, so it's genuinely your/the maintainers' call.

None of this is a correctness problem with the actual discovery/factory pattern being demonstrated — that part is sound and compiles. It's specifically the "STEP N walkthrough" framing, which is the thing the PR body says is the point of this example, that has the leftover-template issue running through it.

Hephaestus-Maximus (changes requested):
Re-reviewed after the latest round of commits. Good progress on the feedback from my last pass — but there's a real, verified build-breaking regression from this round that needs fixing before anything else.

**Build-breaking: `CMakeLists.txt` still references the deleted `CustomDeviceFactory.cpp`.** Commit `f7b43118` ("remove extra file, minimal code is in header") deleted `CustomDeviceFactory.cpp` (its content was always just includes/boilerplate, per my last review), but `ExampleProjects/CustomRobotDevice/CMakeLists.txt`'s `add_executable()` still lists it as a source. I didn't just spot this by reading — I reproduced it:
```
CMake Error at CMakeLists.txt:15 (add_executable):
  Cannot find source file:
    CustomDeviceFactory.cpp
CMake Error at CMakeLists.txt:15 (add_executable):
  No SOURCES given to target: CustomRobotDevice
```
This means the project can't even configure right now, let alone build — which directly contradicts the first line of your own test plan ("Build standalone via the project's own CMakeLists.txt"). Just drop the `CustomDeviceFactory.cpp` line from the `add_executable()` source list.

**Confirmed fixed from my last review:**
- The leftover "port factory"/`PortFactory.h` language in `CustomRobotDevice.cpp`'s header comments is gone — now correctly says "Include the new device header file for class derived from parent ISDevice.h."
- `main.cpp`'s STEP 7 comment no longer duplicates STEP 6's text — now correctly says "Connect and configure device," and a STEP 8 was added for the final read loop, so the walkthrough now actually reaches 8 steps as the PR body claims.
- The `printf("...%lu...", dm.DeviceCount())` format-string mismatch is fixed — now `%zu`, correctly matching `size_t`.
- The README's large struck-through leftover paragraph (describing a different loopback-port example) is gone.
- The STEP-5 numbering collision resolved itself as a side effect of deleting `CustomDeviceFactory.cpp` (it no longer competes with `main.cpp`'s own internal STEP 5).

**Still open, all minor:**
- `${BRIEF_DESC}` is still unfilled in the `@brief` line of all four remaining new source files (`CustomRobotDevice.h/.cpp`, `CustomDeviceFactory.h`, `main.cpp`).
- The README's Doxygen-note code block still says "...found throughout the code in this example to guide users through the process of creating and using custom **ports**" — same leftover terminology mismatch as before, just missed in this pass.
- A stray `%%comment` line is still sitting inside the first mermaid diagram.
- `CustomRobotDevice::onIsbDataHandler()`'s `uint32_t insStatus = insData.insStatus;` is still unused, though now has a "/** optionally do something with this data here */" comment after it that at least signals it's meant as an extension point rather than dead code — I'd still either use it or drop it, since as written it can trigger `-Wunused-but-set-variable` on stricter builds.

Once the CMakeLists fix lands I'd expect this to actually build and be ready for the real test-plan checks (device connection, standalone run) — right now those can't even be attempted.

Hephaestus-Maximus (approved):
Re-verified after the latest two commits — this time by actually rebuilding, not just reading.

**Build-breaking issue from my last review is fixed and confirmed working.** Rebuilt the whole thing standalone (recursive submodule init to pull in `src/libusb`, then `-DIS_SDK_BUILD_FROM_SOURCE=ON`): CMake configures cleanly now that the dangling `CustomDeviceFactory.cpp` entry is gone, and the full build succeeds end-to-end — `libInertialSenseSDK.a` links, and `CustomRobotDevice` links and produces a working executable. This satisfies the first line of your own test plan.

Also confirmed:
- The unused `insStatus`/`copyDataPToStructP` demo code is removed cleanly rather than just commented — no more dead read.
- The last leftover "creating and using custom ports" phrase in the README's Doxygen-note block is now "creating and using SDK class custom implementations," consistent with the rest of the terminology cleanup.
- The added `device->isDisplay.ExitProgram()` check in the STEP 8 loop is a nice touch for graceful ctrl+c exit.

Two purely cosmetic items still remain, not worth blocking on: `${BRIEF_DESC}` is still unfilled in the `@brief` line of all four source files (will show up literally in generated Doxygen docs), and there's still a stray `%%comment` line inside the first mermaid diagram in the README.

Approving.

Hephaestus-Maximus (approved):
Reviewed the new commit ("Brought the display back into main app as sink for all device data with callbacks") — a real architectural change this time, not just doc/build cleanup, so I read and rebuilt it fresh rather than assuming it's more polish.

**The design change itself is sound and an improvement.** Moving `ISDisplay` ownership out of `CustomRobotDevice` and into the application via a `std::function<void(const p_data_t*)> onDataReceived` hook on the device, wired through a new `CustomDeviceFactory::setDataCallback()` static setter that every allocated device gets assigned in `allocateDevice()`, is a more idiomatic pattern for a tutorial than baking a display object into the device class — it correctly demonstrates "the device shouldn't need to know how its data gets displayed." Verified the call ordering in `main.cpp`: `setDataCallback()` is called right after `dm.addDeviceFactory(...)` and before the discovery loop, so `s_dataCallback` is populated before any `allocateDevice()` call could read it. Also verified `onIsbDataHandler()` guards the call with `if (onDataReceived)`, so a factory that forgets to call `setDataCallback()` fails safe (silently no display) rather than crashing on an empty `std::function`.

Rebuilt the whole thing end-to-end again (recursive submodule init, `-DIS_SDK_BUILD_FROM_SOURCE=ON`) — compiles and links cleanly with the new `std::function`/lambda/static-member code, no warnings surfaced.

README, `main.cpp`'s STEP comments, and the actual code are all consistent with each other this time (Step 6 now covers instantiating the display and wiring the callback, Step 7 covers find/connect) — good follow-through on keeping the walkthrough in sync with the architecture change.

**One thing worth being explicit about, given this is a teaching example others will copy:** `CustomDeviceFactory::s_dataCallback` is a `static` (process-lifetime) member, but the callback stored in `main.cpp` captures `isDisplay` *by reference* (`[&isDisplay](...)`). That's safe in this exact program, because `main_discovery()` is only ever called once and `isDisplay` outlives every possible callback invocation within that single call. But it's a real lifetime hazard as a *pattern* to imitate: anyone extending this into an app that calls discovery more than once, or that lets the display go out of scope while a device (or the static callback) outlives it, gets a dangling reference with no warning. Not asking for a code change — just flagging it as a one-line caveat worth adding to the README's Step 6 section, since the whole point of this example is that people will copy it.

Approving.

Hephaestus-Maximus (approved):
Reviewed the four newest commits, which go well beyond just addressing my lifetime comment.

**The lifetime fix is real, not just documented.** `isDisplay` is now a function-local `static` (program lifetime) rather than a stack local captured by reference, and the lambda's capture list is empty (`[]`) — it references the static by name directly, which needs no capture and can never dangle. The comment in both `main.cpp` and the README correctly identifies the deeper reason this matters: it's not just "don't call `main_discovery()` twice," it's that devices retained by the `DeviceManager` singleton hold the callback (and thus a reference to `isDisplay`) for as long as `DeviceManager` keeps them, which isn't neatly scoped to any one call. This is the right fix, not a workaround.

**Caught several more real leftovers I'd missed:**
- The README's build instructions said `cd inertial-sense-sdk/ExampleProjects/CustomPort/CustomVirtual` — wrong project entirely, another leftover from the template this was scaffolded from. Now correctly says `CustomRobotDevice`.
- The README's SDK-file links were `../../../src/...` (one directory too deep) — I verified `../../src/...` is the correct relative path from this project's location; the old links would have resolved above the repo root.
- A logging-example reference to "a line in CustomVirtualPortFactory.cpp" — again the wrong file, now correctly "CustomRobotDevice.cpp".
- `exit(0)` inside `main_discovery()` on the no-device-found path replaced with `return 0` — `exit()` mid-function terminates the process immediately without unwinding the stack or running local destructors, which is a real (if narrow, single-executable-lifetime) correctness improvement, not just style.
- All `printf`-style output standardized to `std::cout`, which also incidentally eliminates any possibility of a format-specifier/argument-type mismatch (like the `%lu`/`size_t` one from a few rounds back) since there's no format string left at all.

Rebuilt end-to-end again — still compiles and links cleanly, and I confirmed the corrected `../../src/DeviceFactory.h` path actually resolves.

**One small leftover I did catch:** `CustomRobotDevice::insData` (`ins_1_t insData = {};`) is still declared, but the `copyDataPToStructP(&insData, ...)` call that populated it was removed in favor of forwarding the raw `p_data_t*` via `onDataReceived`. It's now dead — always default-initialized, never written. Not a bug (unused class member, no warning, no functional effect), just worth either removing it or updating its comment ("structures to hold the data we are interested in") since it no longer holds anything.

Approving — the substantive concerns are resolved and then some.

ExampleProjects/CustomRobotDevice/CustomDeviceFactory.h
kylemallory, line 1: I would renamed this CustomRobotDeviceFactory.  Honestly, you could probably drop "Custom" from all the files, and just go with RobotDevice and RobotDeviceFactory.  Or MyRobot and MyRobotFactory.  But I'd would be consistent that the "RobotDevice" is produced by the "RobotDeviceFactory".

ExampleProjects/CustomRobotDevice/CustomDeviceFactory.h
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * @brief Implements DeviceFactory::allocateDevice(); allocates an ISDevice only if devInfo resolves to an IMX-5 hardware Id.
+     * @param devInfo the device information uniquely identifying the specific device.
+     * @param port an associated port (optional) that this device should be bound to.
+     * @return a new ISDevice if devInfo is an IMX-5, otherwise nullptr.
+     */
+    device_handle_t allocateDevice(const dev_info_t &devInfo, port_handle_t port) override {
+
+        /** When we find IMX-5 dev info, we know we want to allocate a new custom device */
+        if (ENCODE_DEV_INFO_TO_HDW_ID(devInfo) == IS_HARDWARE_IMX_5_0) {
```
kylemallory: There is a better (newer?) mechanism/macro/function to do this comparison -- this one is often the source of bugs.  But, ultimately, we want to allow any IMX device, not just IMX-5 (particularly since we want these examples to be IMX-6 compatible).

ExampleProjects/CustomRobotDevice/CustomDeviceFactory.h
```diff
@@ -0,0 +1,77 @@
...
+            auto device = std::make_shared<CustomRobotDevice>(devInfo, port);
+            device->onDataReceived = s_dataCallback;   // every device shares same sink the application provides
```
kylemallory: This is okay, but the proper OO way would be to register a virtual override that calls into the specific RobotDevice instance's.  Ie, when data is received by this device's CustomRobotDevice's::onIsbDataHandler() the device should handle the data directly, rather than losing the context of the Device instance that held it.

To be clear, this isn't -wrong-... its just self-defeating.  Because if you have multiple CustomRobotDevice instances, that application-level handler loses the context of which specific device the data came from.

ExampleProjects/CustomRobotDevice/CustomRobotDevice.cpp
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    return true;
+}
+
+/**
+ * Steps the communications for this device
+ * @return the results of ISDevice step
+ */
+bool CustomRobotDevice::step() {
```
kylemallory: This shouldn't be necessary.  step() should be virtual and public inder ISDevice, so if you leave it un-implemented in the CustomRobotDevice class, you should just call ISDevice::step(), just like you have written here.

ExampleProjects/CustomRobotDevice/CustomRobotDevice.cpp
```diff
@@ -0,0 +1,92 @@
...
+ * @param port the port_handle_t that this data was received from
+ * @returns 0 if this message was successfully processed by a protocol-specific handler, and should not be further
+ * processed, otherwise return !0
+ */
+int CustomRobotDevice::onIsbDataHandler(p_data_t* data, port_handle_t port) {
+
+    if ( ISDevice::onIsbDataHandler(data, port) ) { // let ISDevice do its handling; we call it first for the integrity check it does
+    
+        if (data->hdr.id == DID_INS_1) {
+            if (onDataReceived)
```
kylemallory: The comment above about the "onDataReceived" really goes here.  Why not have the the CustomRobotDevice handle the data from it's physical device directly, here, right now, rather than deferrring it out the application level.  That's the point of OO design.  My Robot know how to do the "My Robot" thing, and the application side just asks "myRobot.getPosition()" to get that data in the way that CustomRobotDevice defines it.  Its a clean decoupling.

Also, in your class definition, you declare an 'insData' variable -- this should populate that variable, not pass it out of the device in a function call.

ExampleProjects/CustomRobotDevice/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+    IS_SET_LOG_LEVEL(IS_LOG_LEVEL_INFO);
+    
+    /** STEP 6: We need a singleton PortManager and SerialPortFactory, and DeviceManager and CustomDeviceFactory;
+     * we'll make local references, and we register the virtual port factory; also provide a way to display data
+     */    
+    PortManager& pm = PortManager::getInstance();
+    pm.addPortFactory(&SerialPortFactory::getInstance());   // tell the PortManager that we are interested in Serial Ports
+
+    DeviceManager& dm = DeviceManager::getInstance();
+    dm.addDeviceFactory(&CustomDeviceFactory::getInstance());  // tell the DeviceManager that we are interested in Custom Devices
```
kylemallory: Not just "Custom Devices" but for "CustomRobotDevice" specifically (via CustomRobotDeviceFactory).

ExampleProjects/RobotDevice/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+        SLEEP_MS(500);  // give a brief pause for new ports/devices to become available.
+    } //while
+
+    if ( !device ) {
+        // we failed to locate any devices - report and exit gracefully
+        std::cout << "No Inertial Sense devices were found after 3 attempts. Exiting." << std::endl;
+        return 0;    // this is NOT an error
+    }
+
+    if ( !device->connect() ) {
```
kylemallory, line 92: This is good, but as a general note -- discoverDevices's DISCOVER__* flags should have left this device already connected, so this could also be a "isConnected()" test.  This is probably the safer approach, since it will attempt to reconnect if for whaterver reason it wasn't previously.. but, it might be worth calling the behavior out in the comments for the reader.

ExampleProjects/CustomRobotDevice/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+        std::cout << "No Inertial Sense devices were found after 3 attempts. Exiting." << std::endl;
+        return 0;    // this is NOT an error
+    }
+
+    if ( !device->connect() ) {
+        std::cerr << "Could not connect to device on port " << device->getPortName() << std::endl;
+        return -2;
+    }
+
+    std::cout << "Found and connected " << dm.DeviceCount() << " IMX device(s) on port(s): " << std::endl;
```
kylemallory: This is an good pattern.  But, there is a device->getDescription() which returns a string that includes the portName, the hardware type & version, etc.  Up to you if you want to stick to just the "Port Name" as opposed to the full device details.

ExampleProjects/CustomRobotDevice/CustomRobotDevice.h
```diff
@@ -0,0 +1,53 @@
...
+    /** message handler callbacks, with extra commented-out prototype examples we are not using in this case */
+    int onIsbDataHandler(p_data_t* data, port_handle_t port) override;
+    //int onIsbAckHandler(p_ack_t* ack, unsigned char packetIdentifier, port_handle_t port) override;
+    //int onNmeaHandler(const unsigned char *msg, int msgSize, port_handle_t port) override;
+
+    /** an optional configuration function to set up our device and pick the data we want to see, etc */
+    bool configure();
+    
+    /** custom optional hook invoked each data message this device receives, to send data back to app */
+    std::function<void(const p_data_t* data)> onDataReceived;
```
kylemallory: I think I might be missing something -- this is setup as a public member function -- but that's not how its called, and I don't see where its defined anywhere.

If you don't intend to use it, you should remove it.  Otherwise, its confusing to the reader.

ExampleProjects/CustomRobotDevice/main.cpp
```diff
@@ -0,0 +1,161 @@
...
+    pm.addPortFactory(&SerialPortFactory::getInstance());   // tell the PortManager that we are interested in Serial Ports
+
+    DeviceManager& dm = DeviceManager::getInstance();
+    dm.addDeviceFactory(&CustomDeviceFactory::getInstance());  // tell the DeviceManager that we are interested in Custom Devices
+
+    /** One display instance, owned by application not any single device, shared by discovered devices;
+     * data sink for every device the CustomDeviceFactory allocates, devices retained by the DeviceManager
+     * singleton hold a reference to it via onDataReceived, so it must live at least as long as they might.
+     */
+    static cInertialSenseDisplay isDisplay(cInertialSenseDisplay::DMODE_PRETTY);
```
kylemallory: I think this is the part that is missing for me... as a new user of the SDK, this might be seen as "black-box magic".  As the reader, I see that it works -- but I don't really understand what its doing, or why.

I might be inclined to drop the use of isDisplay, and just manually extract the data that the RobotDevice cares about, and print directly.... but I do appreciate the convenience of using pre-existing code, and demonstrating that SDK capability to users as well.  You're call.

kylemallory (approved):
All in all a solid example!  Nice work, Tyler!

Since the README was last, I think most of my comments are addressed in one form or another after reading through it, but it might still be worth reviewing them, and then I'll leave it to you as to whether you want to change anything or not.

ExampleProjects/CustomRobotDevice/main.cpp
shefft: I would probably prefer to drop ISDisplay, simplify, and focus on a few pieces of data a robot device might care about and just show those, especially as ISDisplay is demonstrated in other projects.

ExampleProjects/CustomRobotDevice/CustomRobotDevice.h
shefft: It's re-directed to the lambda in main, so it is used, but if we squeeze out the use of ISDisplay this would all change.

ExampleProjects/CustomRobotDevice/main.cpp
shefft: I'll try getDescription(), I hadn't seen it.

ExampleProjects/CustomRobotDevice/CustomRobotDevice.cpp
shefft: OK, since we don't do anything custom in there that's fine and I will make that change.

ExampleProjects/CustomRobotDevice/CustomRobotDevice.cpp
shefft: I went back and forth on this and thought about putting the ISDisplay attached to the device, but ultimately went with wanting to serve multiple devices with the same display output, decoupling them from how exactly that was done.   I had code to use insData, then realized I didn't have any operations to do on it.

Now if we change it up by removing ISDisplay, I'll do the "handling" inside the device object itself.

ExampleProjects/CustomRobotDevice/CustomDeviceFactory.h
shefft: Sure, I'll make those class and file name and reference adjustments to drop 'Custom' and match the device and factory names.

ExampleProjects/CustomRobotDevice/main.cpp
shefft: Will be adjusted with forthcoming project/file name changes.

ExampleProjects/CustomRobotDevice/CustomDeviceFactory.h
shefft: That was one of my early-on questions about the approach to this.  I think the example makes it pretty straight-forward for the user to modify code for any desired target, so I don't think leaving it as one demo target is necessarily out of the question, but maybe we should do some work here to bake in multi-target support if that's your preference?

tmcd249 (approved):

ExampleProjects/CustomRobotDevice/CustomDeviceFactory.h
shefft: From an assumed interpretive design lens of an application that's just dumping the data from all devices and does not care about device specifics, this is legit, but if we are getting rid of ISDisplay and want to demonstrate the final destination of the data as being inside the device object then we should definitely make that change.

ExampleProjects/RobotDevice/main.cpp
shefft: This block was brought over from Simple_Discovery, so I'll look at the documentation there and see what it says.

Hephaestus-Maximus (changes requested):
## Review summary

Cloned the branch, diffed against `develop`, read through all 5 new source files, and — since the author's own test-plan checklist is unchecked and this project isn't wired into CI — built it myself: `InertialSenseSDK` builds clean, and `ExampleProjects/RobotDevice` configures and links against the freshly-built static lib without errors or warnings. Ran it with a bogus port argument and confirmed the "no device found after 3 attempts" path exits 0 as documented. I don't have hardware to exercise the actual DID_INS_1 streaming path, so that part of the test plan is still unverified by me.

### `RobotDevice::onIsbDataHandler()` breaks the dispatch contract it's built on (RobotDevice.cpp:68-89)
`ISDevice::processIsbMsgs()`'s doc comment (src/ISDevice.cpp:25-34) is explicit about the return-value contract: *"0 if this data packet WILL NOT BE processed again by other handlers; any other value indicates that the packet MAY BE processed by other handlers... Effectively, 0 = End-of-line."* The base `ISDevice::onIsbDataHandler()` follows this: it returns `0` only for a genuinely malformed packet (`size==0`/`ptr==NULL`), and `1` ("continue, others may still want this") for every other DID it recognizes (`DID_DEV_INFO`, `DID_SYS_PARAMS`, `DID_FLASH_CONFIG`, etc.) — even ones it fully handled itself.

`RobotDevice`'s override collapses this:
```cpp
if ( ISDevice::onIsbDataHandler(data, port) ) {
    if (data->hdr.id == DID_INS_1) { ... print ... }
    return 0;  // indicates ISDevice found valid message and we processed here
}
return 1;
```
The `return 0;` is outside the `DID_INS_1` check, so it fires for *any* valid message — not just the one DID this class actually cares about. A `DID_DEV_INFO` or `DID_SYS_PARAMS` packet that RobotDevice does nothing with still gets reported as "End-of-line," silently preventing any other handler on the same port from ever seeing it. This is the opposite of what the base class does with the exact same packets, and the class's own doc comment ("@returns 0 if this message was successfully processed... otherwise return !0") doesn't match what the code does for non-INS_1 DIDs either.

Since this is explicitly a tutorial meant to be copied ("intended as a tutorial for SDK consumers building their own custom device/factory pair"), the wrong pattern here will likely get propagated into real applications. The fix is straightforward — forward the base's return value except when this class actually consumes the message:
```cpp
int RobotDevice::onIsbDataHandler(p_data_t* data, port_handle_t port) {
    int ret = ISDevice::onIsbDataHandler(data, port);
    if (ret && data->hdr.id == DID_INS_1) {
        ...
        return 0;
    }
    return ret;
}
```

### Not wired into `ExampleProjects/CMakeLists.txt` — the PR body already asks "Do we want this?"
Confirmed this is still true on the current diff — `RobotDevice` doesn't appear in `ExampleProjects/CMakeLists.txt` alongside the other `add_subdirectory(...)` calls, and I verified CI's full build log has zero mentions of "RobotDevice," so this code has never actually been compiled by CI (only by me, manually, just now). Since the whole point is a maintained tutorial that stays buildable as the SDK evolves, I'd recommend wiring it in — otherwise there's no signal when a future SDK change silently breaks it.

### Minor
- All 4 new source files (`RobotDevice.h/.cpp`, `main.cpp`, `RobotDeviceFactory.h`) still have the literal template placeholder `@brief ${BRIEF_DESC}` in their file-header doxygen comment — looks like it was never filled in.
- The PR description refers to the new classes as `CustomRobotDevice`/`CustomDeviceFactory` and says the factory allocates "only for IMX-5 hardware IDs," but the shipped code is named `RobotDevice`/`RobotDeviceFactory` and `allocateDevice()` matches any `IS_HARDWARE_TYPE_IMX` device (any IMX version), per its own doc comment ("matching any IMX hardware version"). The description looks like it wasn't updated after a rename/behavior change — worth a quick pass before merge since this is also going in as the tutorial's own selling point.

Everything else — the `RobotDeviceFactory` singleton pattern, the `main.cpp` discovery/retry loop, the `CMakeLists.txt` standalone build wiring — looks correct and matches the SDK's existing example conventions (`ImxDeviceFactory`, `Simple_Discovery`).

Hephaestus-Maximus (approved):
## Follow-up review

Confirming my previous request-changes concern is resolved. Pulled the current head, re-read `RobotDevice.cpp`, and rebuilt both the SDK and the standalone example from scratch to re-verify (clean build, no warnings; smoke-tested the no-device path again, still exits 0 as documented).

**Dispatch-contract fix confirmed correct:**
```cpp
int ret = ISDevice::onIsbDataHandler(data, port);
if (ret && data->hdr.id == DID_INS_1) {
    ...
    return 0;
}
return ret;
```
This now forwards the base class's return value for every message this class doesn't specifically consume, and only collapses to `0` ("end-of-line") for the one DID it actually handles — matches `ISDevice::processIsbMsgs()`'s documented contract for every other DID, exactly the fix I'd suggested. No longer swallows unrelated messages from other potential handlers on the same port.

Two minor items from my earlier review remain open but are not blocking, given the project's own README/body already surfaces one of them as an explicit open question and both are cosmetic:
- `@brief ${BRIEF_DESC}` template placeholder is still present in the file-header comment of all 4 new source files.
- `ExampleProjects/CMakeLists.txt` still doesn't have `add_subdirectory(RobotDevice)` — the PR body's own "Do we want this?" question is still open. Worth a final decision before merge, but that's a team call, not a defect.

CI is green (Linux + Windows). Approving.

PR #1289, opened 2026-09-01
SN-8553: host-side support for updating a GNSS receiver behind a module

kylemallory:
The SDK half of SN-8553. Pairs with **inertialsense/imx#1677** and **inertialsense/is-imx#45**, and should merge first.

Five changes, each independently useful.

## `fwUpdate`: let a device ask the host to slow down, not just to stop

`fwUpdate_handleChunk()` could only emit `CHUNK_ID_PAUSE` + `REASON_TOO_FAST` (a hold) or `REASON_WRITE_ERROR` when refusing a chunk. There was no counterpart to `fwUpdate_requestChunkPause()` for the *delay* form, so a device whose constraint is a steady rate rather than an episodic wait could not express it — and falling back on `REASON_WRITE_ERROR` is actively harmful: it takes the crude `chunkDelay × 1.5` branch, forfeits decay progress, and **counts toward `MAX_SAME_CHUNK_RESENDS`**, so 25 refusals of one chunk inside 5 s fails the upload with `ERR_FLASH_WRITE_FAILURE`.

Adds `slowdown_requested`, a protected `fwUpdate_requestSlowdown()`, and a third arm in the refusal branch, symmetric with the existing pause flag.

## `fwUpdate`: apply the derived chunk pace

`nextChunkSend` was armed *after* the local-device early return in `fwUpdate_writeToWire()`, so for any in-process updater the send gate was permanently open and `chunkDelay` had no effect at all — the host ignored every pace it derived. Moved to the top of the function, before the early return.

The `REASON_TOO_FAST` handler also sent the next chunk immediately instead of scheduling it; it now schedules and returns, which is what lets the derivation converge. Measured settling after the fix: 5 requests, 13 → 25 → 33 → 41 → 48 ms, arriving at the analytic rate for the link.

## `fwUpdate`: stop defaulting to the obsolete alternate MD5

`IMG_FLAG_useAlternateMD5` was selected whenever the target's running version could not be read — "for safety" — but no current device computes that hash, so the transfer failed its integrity check at the end having delivered every byte correctly. A version probe that merely timed out was enough to trigger it.

Now requires what actually needs it: a GPX-1.0 running firmware at or below 2.0. Everything else uses the standard MD5.

## `ISDevice`: let a device be told never to validate

A receiver reached through a module, or an IMX in bridge mode, is a legitimate `ISDevice` bound to a port that never carries anything the ISB parser recognises. Validation could not be switched off, so such a device was refused before any command reached it.

Adds `doNotValidate` with `disableValidation()` / `validationDisabled()`, guarded at the top of **both** `validateAsync()` and `validate()`, and `InertialSense` binds an unvalidated device via `registerNewDevice()` when validation is disabled. `COMM_PORT_FLAG__EXPLICIT_READ` does not cover this — it only stops `is_comm_port_parse_messages()` reading the port, which is a different concern.

## `ISComm`: complete `ptype_names[]`

The table was short of `_PTYPE_SIZE`, so `SEPT_SBF` and `SEPT_REPLY` printed as garbage in parse-debug output — read as protocol corruption more than once.

## Validation

Exercised end to end by the two paired PRs: `cltool -uf-cmd target=SEPT,upload=…` installs a 2.95 MB image into a Septentrio mosaic-G5 through an IMX-6 and confirms it (`2951096/2951096` bytes, `FINISHED`, 81.4 s), and directly against a receiver on a USB-UART. 159/159 `is-common` unit tests pass.

`origin/3.1.0-rc` is merged in (`Remove eImx6ImuPopulationType enum definition`, `data_sets.h`), so this is current with its base.

### DIFF
--- a/src/ISComm.c
+++ b/src/ISComm.c
@@ -1385,7 +1385,7 @@ static inline void parse_messages(is_comm_instance_t* comm, port_handle_t port)
     protocol_type_t ptype;
 
 #if DEBUG_PARSE_MSG
-    static const char* ptype_names[_PTYPE_SIZE] = { "NONE", "ERROR", "ISB_ACK", "ISB_CMD", "ISB_DATA", "NMEA", "UBLOX", "RTCM3", "SPARTN", "SONY" };
+    static const char* ptype_names[_PTYPE_SIZE] = { "NONE", "ERROR", "ISB_ACK", "ISB_CMD", "ISB_DATA", "NMEA", "UBLOX", "RTCM3", "SPARTN", "SONY", "SEPT_SBF", "SEPT_REPLY" };
     static char log_msg[512];
     int log_msg_pos = sprintf(log_msg, "parse_messages() parsed:");
 #endif
--- a/src/ISDevice.cpp
+++ b/src/ISDevice.cpp
@@ -458,6 +458,12 @@ bool ISDevice::queryDeviceInfoISbl(uint32_t timeout) {
 
 
 bool ISDevice::validate(uint32_t timeout) {
+    // validate() is its own blocking implementation rather than a wrapper around validateAsync(), so
+    // it needs the guard too -- see disableValidation(). Silent, because a caller that disabled
+    // validation asked for exactly this and does not need telling each time.
+    if (doNotValidate)
+        return false;
+
     if (!isConnected()) {
         // INFO, not debug: validating a closed port is a caller mistake in the same class as querying
         // one, and not a hard failure -- but the caller should hear about it. This return also precedes
@@ -575,6 +581,13 @@ bool ISDevice::validate(uint32_t timeout) {
  *         ASYNC_STATE__SUCCESS if the device successfully validated.
  */
 int ISDevice::validateAsync(uint32_t timeout) {
+    // Guarded here rather than at each caller: step(), DeviceFactory::stepValidation() and the
+    // bootloader updaters all reach validation through this one function, and a device told never to
+    // validate must mean it for all of them. FAILURE rather than TIMEOUT for the same reason a closed
+    // port returns it -- nothing was asked, and retrying will not change that.
+    if (doNotValidate)
+        return ASYNC_STATE__FAILURE;
+
     if (!isConnected())
         return ASYNC_STATE__FAILURE;
 
--- a/src/ISDevice.h
+++ b/src/ISDevice.h
@@ -898,6 +898,27 @@ public:
     /** @brief Non-blocking counterpart to validate(): advances one step of the query cycle per call. @return an AsyncState value indicating progress/result. */
     int validateAsync(uint32_t timeout = 1000);
 
+    /**
+     * Stops this device ever attempting to validate.
+     *
+     * For a device that is known not to speak ISB at all -- a port bound without validation, or one
+     * carrying a directly attached third-party receiver -- validation cannot succeed, and every
+     * attempt writes probe traffic to the link and consumes what comes back. On a port whose data
+     * belongs to something else, that is not merely wasted effort: it takes the bytes that other
+     * reader was waiting for.
+     *
+     * Distinct from COMM_PORT_FLAG__EXPLICIT_READ, which says another reader owns the port rather
+     * than that this device should give up on identifying itself. The flag happens to suppress the
+     * validation attempt in step(), but says nothing to validate()/validateAsync() called from
+     * anywhere else, and conflating the two leaves the intent unrecorded.
+     *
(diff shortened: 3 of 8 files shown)


tmcd249 (approved):

Hephaestus-Maximus (approved):
In-depth review of all 5 changes (8 files, +150/-9). This touches the exact `fwUpdate` flow-control/pacing machinery from sdk#1270/#1282 that I reviewed extensively a few days ago, so I traced each change against that code rather than reading it fresh.

**1. `fwUpdate_requestSlowdown()`** — verified it's correctly scoped: the new "real chunk id + REASON_TOO_FAST" branch sits inside the same `if (reason == REASON_TOO_FAST) { ...; return true; }` block as the existing `CHUNK_ID_PAUSE` branch, so both sub-forms bypass the give-up-guard accounting below (`last_resent_chunk`/`resent_chunkid_count`) exactly as the comment claims — confirmed by reading the control flow, not just trusting the comment. The device-side split (`pause_requested` vs `slowdown_requested` vs falling through to `REASON_WRITE_ERROR`) is symmetric with the existing pause mechanism and correctly reset in `fwUpdate_resetEngine()`.

**2. `nextChunkSend` arming moved to the top of `fwUpdate_writeToWire()`** — read the full function before and after: previously the line sat after the `deviceUpdater != nullptr` early return, so it only ever ran for the remote-device path. Moving it to the top means it now runs unconditionally, fixing exactly the described bug (pacing had zero effect on any in-process/local updater) with no behavioral change to the remote path — the only difference there is `current_timeMs()` being sampled a few instructions earlier, not something that matters at this timescale. The companion fix in `fwUpdate_handleResendChunk()` (`nextChunkSend = current_ms + chunkDelay; return true;` instead of calling `fwUpdate_sendNextChunk()` immediately) is correctly scoped to only the "real chunk id" TOO_FAST branch — verified the `CHUNK_ID_PAUSE` branch above it still returns early via its own path, unaffected. Also checked the `CHUNK_DELAY_MAX_MS` 50→1000ms change: 512 bytes at 115200 baud with 8N1 framing (10 bits/byte) is 5120/115200 = 44.4ms, matching the PR's own arithmetic for why 50ms left no room for a device that genuinely needs to ask twice.

**3. GPX-1.0-only alternate MD5** — confirmed `isGpx10` requires `target_devInfo` truthy before reading `remoteDevInfo`, so an unreadable version now correctly falls through to standard MD5 (the opposite of the old default). Confirmed the broadening from exact `2.0.0` to "≤2.0" (dropping the `firmwareVer[2]==0` check) is deliberate and matches the PR's own stated threshold framing, not an oversight. Confirmed this is the only site that sets `IMG_FLAG_useAlternateMD5`, so the fix isn't partial.

**4. `doNotValidate`/`disableValidation()`** — both `validate()` and `validateAsync()` guarded at the top, returning `false`/`ASYNC_STATE__FAILURE` respectively (nice continuity with the `ASYNC_STATE__FAILURE` naming from sdk#1282, now put to real use). The `InertialSense::OpenPorts()` wiring has a real concurrency detail I verified rather than assumed: it collects ports from `portManager.locked_range()` into a `std::vector` in one loop, then calls `registerNewDevice()` in a *separate* loop after that range (and its lock) has gone out of scope. I confirmed `locked_range()` genuinely holds a mutex for its duration and `registerNewDevice(port_handle_t, dev_info_t)` genuinely takes `DeviceManager`'s own `std::recursive_mutex` — so this two-pass structure is a real, correctly-reasoned avoidance of a lock-order inversion, not defensive-comment theater.

**5. `ptype_names[]` completion** — confirmed `_PTYPE_SIZE` is 12 (`_PTYPE_SEPTENTRIO_REPLY(11) + 1`) and the array now has exactly 12 entries in matching order.

CI is green. **One recurring gap worth naming**: none of these five changes have unit test coverage, and this is now the third PR in a row touching this exact pacing/flow-control subsystem (sdk#1270, #1282, this one) where the validation story is entirely hardware-dependent (the paired imx/is-imx PRs) rather than `is-common` unit tests. The bug fixed in change #2 here — pacing silently having zero effect for in-process updaters — is exactly the kind of regression a targeted test against the existing `ISFirmwareUpdateTestHost`/`ISFirmwareUpdateTestDev` harness (already used elsewhere in `tests/test_ISFirmwareUpdate.cpp`) would catch cheaply and immediately, rather than requiring another round of hardware measurement to notice if it regresses again. Not blocking — the fix itself is correct and well-verified by hand — just flagging the pattern since it's now recurred three times in this exact area.

Approving.

waltjohnson (approved):

PR #1291, opened 2026-09-02
SN-8567/SN-8568: fix the ISv2 stall waiting for ISbl, and report relay device occupancy

kylemallory:
## Problem

An ISv2 firmware update reset the device into its bootloader and then waited for it to
appear there — re-issuing an APP-mode reset every 10 s at a device already sitting in
ISbl — until the 60 s phase budget expired and the update failed. Affects every
transport; reproduced over both the relay and plain USB serial.

## Root cause

A bootloader entered from APP is **unsynchronized**: it needs its `'U'` autobaud burst
once before it answers anything. Measured on SN62913 (ISbl v6i):

```
bare version query  ->  (no response)
after a 'U' burst   ->  aa 55 06 69 01 01 00 c1 f5 00 00 2e 0d 0a
bare query again    ->  answers          (sync persists once established)
```

The ISbl wait left that discovery to `validate()`/`validateAsync()`, which hand
`queryDeviceInfoISbl()` ~250 ms as one step of a round-robin that also tries NMEA and
ISB. `queryIsblVersionFrame()` derives its schedule from that budget and truncates the
autobaud burst to whatever remains, so the bootloader never synchronized — and an
unsynchronized bootloader is invisible to the host and to the relay alike, the relay
continuing to report the device's last known APP state.

## Fix

`ISBFirmwareUpdater::fwUpdate_step()` probes for ISbl directly during the phase that is
waiting for exactly that, with a 1500 ms budget that fits a full burst, throttled to
once every 2 s because the probe blocks.

## Validation

| Transport | Before | After |
|---|---|---|
| Local USB (SN62913) | wait never completed | `Rediscovered ... ISbl (v6i)` 208 ms after the first wait message; PASSED in 36.7 s |
| Relay (SN60246) | ran to the 60 s budget -> `ERR_TIMEOUT` | `Rediscovered ... ISbl (v6j)` 251 ms later; uploaded in 119.9 s; PASSED |

A device deliberately left in an unsynchronized ISbl is now recovered by the ci_hdw
fixture in 33 s, after which the full suite runs 42/45 passing.

SDK unit tests: 589 passed, 4 pre-existing skips.

## Relay device occupancy (SN-8568)

`RelayPortFactory::parseDeviceJson()` discarded `has_tcp_client` and `listening`.
`DeviceRecord` now carries both, and `getRelayDevices()` exposes the per-device view so
a consumer can tell a busy slot from an absent device.

Occupancy must come from the relay's report rather than a connection attempt: the
bridgeboard accepts a second client and then closes it, so `portOpen()` succeeds and the
refusal only surfaces later as `ECONNRESET`, and the attempt would claim the slot being
tested. The accessor serves the cached poll/SSE state, so it performs no I/O and opens no
port. `listening` defaults true so an older bridgeboard that omits the field still reads
as offering the slot.

The ci_hdw consumer is in the matching is-common PR.


aabogdanov (approved):

Hephaestus-Maximus (approved):
## Review summary

Cloned the branch and diffed against `3.1.0-rc` (final diff is small and clean: 78 lines added across 4 files, nothing removed — matches the PR's own account of reverting the broader `devInfoConfirmedMs` approach once the narrower autobaud-burst root cause was demonstrated directly). Traced both fixes through the surrounding code rather than taking the description at face value.

### SN-8567 — ISbl probe budget (`ISBFirmwareUpdater.cpp`/`.h`)
- Confirmed the math in the new doc comment: `BOOTLOADER_HANDSHAKE_COUNT` (50) × `BOOTLOADER_HANDSHAKE_DELAY` (10 ms) = 500 ms, comfortably inside the new 1500 ms probe budget — the old ~250 ms round-robin slice genuinely couldn't fit a full burst.
- The new probe is gated behind `if (!device->hasDeviceInfo() && (device->validateAsync() != 1)) return true;`. For the actual failure scenario (a device that *was* known while in APP mode before the reboot, so `hasDeviceInfo()` stays true on stale cached info) that early return is skipped and the probe runs — correctly targets the reported bug.
- `queryDeviceInfoISbl()` itself (untouched by this PR) sets `devInfo.hdwRunState = HDW_STATE_BOOTLOADER` and `firmwareVer[]` on a successful frame, which is exactly what the following `if (device->devInfo.hdwRunState == HDW_STATE_BOOTLOADER)` block and its log message consume — the new call is wired into state the rest of the function already expects.
- A blocking `queryDeviceInfoISbl()` call from inside `fwUpdate_step()` isn't a new pattern — `fetch_device_info_and_signature()` a few hundred lines down already does the same thing with a 1000 ms budget, so this isn't introducing a new blocking-call architecture, just applying the existing one to the wait phase that needed it.

### SN-8568 — relay device occupancy (`RelayPortFactory.cpp`/`.h`)
- `getRelayDevices()` mirrors the existing `getRelayHosts()` exactly (same mutex, same enabled-host filter, same copy-out pattern) — no new locking risk.
- `parseDeviceJson()`'s new `dev.value("has_tcp_client", false)` / `dev.value("listening", true)` reads are defensive (defaulted), and the `listening` default of `true` matches the stated intent of treating an older bridgeboard that omits the field as still offering the slot.
- No new unit test lands in *this* repo for `getRelayDevices()` itself, but I verified the consumer: imx#1680's SDK submodule pointer (`8da4c4ca1a2`) is an exact match for this PR's head commit, and that PR's validation table shows the occupancy accessor exercised against real relay hardware (contention detected/held/released, wedged-device handling). So the feature is exercised end-to-end even though the SDK-side unit suite doesn't cover it directly.

### CI
Green on the current head (`8da4c4ca1a2`) — Linux and Windows both pass.

Nothing found that needs changing. Appreciate the commit history keeping the reverted `devInfoConfirmedMs` attempt visible with the reasoning for backing it out rather than squashing it away — made this much easier to verify.

tmcd249 (approved):

PR #1292, opened 2026-09-03
SN-8608: render platform type and IMX io-config in device descriptions

kylemallory:
Jira: [SN-8608](https://inertialsense.atlassian.net/browse/SN-8608)

A device description named the module but not the board it was on, so nothing in a log or test result said which hardware was targeted. On a bench carrying an IMX-5+GPX-1, an IMX-6+u-blox-X20 on a RUG-4 and an IMX-6+Septentrio on a BRK, those lines differed only by serial number.

Rendered from the five boards on my bench, through this code path:

```
SN63736     (IMX-5.0, IG1.1-G2) fw3.1.0 [G1G2=off]
SN942757193 (IMX-6.0, RUG4-X20) fw3.1.0
SN519465    (IMX-5.0, IG2)      fw3.1.0
SN942753848 (IMX-6.0, BRK-SG5)  fw3.1.0 [G1G2=CAN]
SN63809     (IMX-5.0, IG2)      fw3.1.0
```

Three stock boards are silent; the two that differ say how.

## What it adds

Four renderers in `utils`, plus `SHOW_PLATFORM` (0x0020) / `SHOW_IO_CONFIG` (0x0040) / `CONFIG_VERBOSE` in `ISDevice::DevInfoFormatFlags`:

```cpp
const char* platformTypeName(uint8_t platformType);
std::string platformDescription(uint32_t platformConfig, int flags = 0);
std::string ioConfigDescription(uint32_t ioConfig, uint8_t ioConfig2, uint32_t platformConfig, int flags = 0);
std::string imxConfigDescription(const nvm_flash_cfg_t& cfg, int flags = 0);
```

Both flags **default off**, so no existing description output changes.

## Design points worth a reviewer's attention

**Flash config, not OTP.** Both groups read only `nvm_flash_cfg_t` (`platformConfig`, `ioConfig`/`ioConfig2`), because flash config is what the firmware runs on and can differ from manufacturing OTP. OTP remains the ci_hdw fixture's business, for provenance and recovery.

**Deviation-only rendering.** The io stanza shows only fields differing from what the platform implies, at group granularity, so a correctly configured board renders nothing and anything printed is worth reading.

**The baseline must be seeded with `IO_CONFIG_DEFAULT`** before the overlay, matching `nvm_flash_cfg_defaults()`:

```cpp
uint32_t expected = IO_CONFIG_DEFAULT;  uint8_t expected2 = 0;
imxPlatformConfigToFlashCfgIoConfig(&expected, &expected2, platformConfig);
```

`imxPlatformConfigToFlashCfgIoConfig()` is an overlay, not a complete generator — it sets only the platform-derived GNSS fields. Without the seed, the pin-function fields of every stock board read as deviations. This was found by diffing the mapper against real boards, and `StockBoardsRenderNothing` pins it.

**Deliberately not rendered:** the carrier preset id (it only selects an io configuration; the resulting `ioConfig` is the observable) and the RUG I/O-expander field (an input only when no preset is selected, and `RUG.cpp` overrides `s_ioBits` from the preset table without writing back, so nothing keeps the field in step with an active preset).

**Static overloads.** The platform is rendered by the instance `getName()`, which is what produces the parens. The static `dev_info_t` overloads cannot render either group and now say so at their declarations.

## Incidental fix

`ISDevice::getDescription(const dev_info_t&, int)` was **declared in the header but never defined** — a latent link error for any caller. Nothing called it, so it went unnoticed; the tests here are the first caller. Now implemented as name + firmware (a `dev_info_t` has no port, platform or io config).

## Testing

620 tests, **616 passed, 4 skipped (pre-existing), 0 failed**. 14 are new.

Five test vectors are captured from real hardware, covering u-blox / GPX / Septentrio types, Ser0 and Ser2 sourcing, mixed sourcing (GNSS1@Ser0 + GNSS2@Ser2), PPS1=G15, a non-zero `ioConfig2` (`0xC0` = PPS2 on G13, matching the BRK-2 routing the enum documents), CAN on G1G2, G1G2 disabled, and both OTP-locked and unlocked platforms. **No hardware required.**

Not covered by real hardware: `BRK_GPX` (1), `BRK_2_X20` (2), any RUG-3, an SPI-enabled board, a GPX-1 as its own device.

## Reviewer notes

- **Pin detail is `VERBOSE`-only** by design, which means a deviation confined to `G15` / `G5G8` / strobe edge / QDEC won't appear by default. It never happens on the five real boards, but it is a blind spot; happy to include those groups when they deviate if reviewers prefer.
- `tests/CMakeLists.txt` globs `test_*.cpp` without `CONFIGURE_DEPENDS`, so a newly added test is silently skipped until CMake reconfigures. Not changed here, but it cost me a confusing "0 tests" run.
- Related: SN-8469 (removing the `RUG4_GPX` 4.0 special case in `rug.cpp` — the reason preset 7 maps to different expander bits on Rugged-4.0 vs 4.1).

[SN-8608]: https://inertialsense.atlassian.net/browse/SN-8608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ISDevice.cpp
+++ b/src/ISDevice.cpp
@@ -797,7 +797,17 @@ std::string ISDevice::getName(const dev_info_t &devInfo, int flags) {
 }
 
 std::string ISDevice::getName(int flags) const {
-    return getName(devInfo, flags);
+    std::string out = getName(devInfo, flags);
+
+    // The carrier board belongs with the hardware it qualifies, so it goes inside the same parens.
+    // Only an instance can render it: the platform is held in flash config, not devInfo, and an
+    // unsynchronized flash config has not been read from the device and so describes nothing.
+    if ((flags & SHOW_PLATFORM) && (imxFlashCfg.checksum != 0xFFFFFFFF)) {
+        std::string platform = utils::platformDescription(imxFlashCfg.platformConfig, flags);
+        if (!platform.empty() && !out.empty() && (out.back() == ')'))
+            out.insert(out.size() - 1, ", " + platform);
+    }
+    return out;
 }
 
 /**
@@ -828,6 +838,15 @@ std::string ISDevice::getFirmwareInfo(int flags) const {
     return getFirmwareInfo(devInfo, flags);
 }
 
+std::string ISDevice::getDescription(const dev_info_t& devInfo, int flags) {
+    // No port, platform or io configuration here: a dev_info_t carries none of them, so this is the
+    // name and firmware only. The instance overload adds the rest.
+    std::string desc = getName(devInfo, flags);
+    if (!(flags & OMIT_FIRMWARE_VERSION))
+        desc += " " + getFirmwareInfo(devInfo, flags);
+    return desc;
+}
+
 std::string ISDevice::getDescription(int flags) const {
     std::string desc = getName(flags);
     if (!(flags & OMIT_FIRMWARE_VERSION)) {
@@ -835,6 +854,16 @@ std::string ISDevice::getDescription(int flags) const {
     }
     if (!(flags & OMIT_PORT_NAME) && portIsValid(port))
         desc += ", " + getPortName() + (isConnected() ? "" : " (Closed)");
+
+    // After the port, so the port and its connection state stay adjacent. Renders nothing unless
+    // the configuration differs from what this device's platform implies, which keeps it absent on
+    // a device configured as its carrier intends. A GPX has no ioConfig, and its imxFlashCfg is
+    // never synchronized, so the checksum gate covers that too.
+    if ((flags & SHOW_IO_CONFIG) && (imxFlashCfg.checksum != 0xFFFFFFFF)) {
+        std::string io = utils::ioConfigDescription(imxFlashCfg.ioConfig, imxFlashCfg.ioConfig2, imxFlashCfg.platformConfig, flags);
+        if (!io.empty())
+            desc += " [" + io + "]";
+    }
     return desc;
 }
 
--- a/src/ISDevice.h
+++ b/src/ISDevice.h
@@ -95,6 +95,11 @@ public:
         COMPACT_HARDWARE_VER     = 0x0004,      //!< forces hiding digits 3 & 4 of the hardware version number, digits 1 & 2 are always shown
         COMPACT_SERIALNO         = 0x0008,      //!< disables zero-padding of the serial number
         COMPACT_BUILD_TYPE       = utils::FWI_COMPACT_BUILD_TYPE, //!< formats the build-type (when the firmware version is show) as a single character
+        SHOW_PLATFORM            = 0x0020,      //!< renders the carrier board, from flash config, inside the hardware parens: "SN60246 (IMX-5.0.4, RUG4-X20)". Instance methods only -- see the note on the static overloads
+        SHOW_IO_CONFIG           = 0x0040,      //!< renders the IMX io configuration after the port, bracketed, showing only what deviates from the platform's own configuration. Instance methods only
+
+        // Config Options -- aliases of the utils bits, so the two definitions cannot drift
+        CONFIG_VERBOSE           = utils::CFGI_VERBOSE,           //!< with SHOW_PLATFORM/SHOW_IO_CONFIG, renders every field rather than only deviations, and includes the pin-level detail
 
         // Version Options -- aliases of the utils bits, so the two definitions cannot drift
         OMIT_COMMIT_HASH         = utils::FWI_OMIT_COMMIT_HASH,   //!< suppresses the output of the commit hash/dirty status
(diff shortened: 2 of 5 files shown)


shefft (approved):

Hephaestus-Maximus (approved):
## Review summary

Cloned the branch, read through both commits in full, and — rather than trusting the reported test numbers — built the SDK unit-test project from source myself and ran the suite:

```
[==========] 620 tests from 70 test suites ran. (78390 ms total)
[  PASSED  ] 616 tests.
[  SKIPPED ] 4 tests, listed below:
[  SKIPPED ] LogReaderLiveFixture.HdwIdMatchesImx5_0FromCltoolCapture
[  SKIPPED ] protocol_nmea.zda_gps_time_skip
[  SKIPPED ] TimeResolverLiveFixture.RealCltoolCaptureSmoke
[  SKIPPED ] V21EndToEnd.PseudoRandomMultiDidWithTimelessPortMonitor
```
Exact match for the PR's own claim (620/616/4/0), including all 14 new `platform_ioconfig.*` tests passing — `StockBoardsRenderNothing` and `DeviatingBoards` in particular, which are the ones that actually exercise the seeding-order correctness this PR's design hinges on.

### `raw_hexdump()` fix (commit 2) — verified correct by tracing the mechanism
The old code's bug is real and exactly as described: `ptr += SNPRINTF(ptr, ptrEnd - ptr, ...)` advances by `snprintf`'s return value, which is the length it *would* have written, not what it did — once the 2048-byte buffer filled, `ptr` walked past `ptrEnd`, `ptrEnd - ptr` went negative, and passing that as the unsigned `size_t` count to the next `snprintf` call is exactly the mechanism that overflows the stack buffer. The rewrite sidesteps the whole class of bug by giving every `SNPRINTF` call into `hex[4]` its own full, fixed 4-byte bound ("ff " + NUL, always exactly enough for a `%02x`) and accumulating into the `std::string` instead of walking a raw pointer — there's no shrinking "remaining space" counter left to go negative. I also checked the two "adjacent defects" claims: the old `linelen = bytesLen % bytesPerLine` on the last line does drop a full final line when `bytesLen` is an exact multiple of `bytesPerLine` (renders a blank line instead) — the new `(bytesLen - i) < bytesPerLine ? (bytesLen - i) : bytesPerLine` correctly handles that boundary. And the removed `DISPLAY_DELTA_TIME` block did reference undeclared identifiers (`wheel`, an out-of-scope `i`), confirmed dead/non-compilable code.

One gap: there's no dedicated unit test for `raw_hexdump()` itself (`grep -rl raw_hexdump tests/` turns up nothing). The commit message's verification (1/31/32/33/64/600/700/1024/4096 bytes, matching line/character counts) is convincing but was evidently done ad hoc and isn't captured as a regression test — worth adding given this was a stack-smashing abort that destroyed failure evidence on a real ci_hdw run. Not blocking; the fix is structurally sound.

### Platform/io-config renderers (commit 1) — the design claims check out
- **No bit collision**: walked the full `DevInfoFormatFlags`/`devConfigInfoFlags_e` bit layout by hand — 0x0001/0x0002/0x0004/0x0008/0x0010 were already taken, then it jumps to 0x0100+ for the `OMIT_BUILD_*` family, so 0x0020/0x0040/0x0080 for the three new flags were genuinely free.
- **The seeding claim** ("baseline must be `IO_CONFIG_DEFAULT` before the overlay") checks out against `imxPlatformConfigToFlashCfgIoConfig()` in `imx_defaults.c`: it's a true overlay — every branch only calls `SET_IO_CFG_GNSS*_*` macros, never touching the pin-function bits — so seeding with `IO_CONFIG_DEFAULT` first is exactly what makes a stock board's baseline match its actual defaults instead of reading every pin field as a deviation. `StockBoardsRenderNothing` passing is the empirical proof.
- **`getName()`'s paren-insertion** (`out.back() == ')'` before inserting `", " + platform`) is a defensive check that's actually always satisfied — the static `getName(devInfo, flags)` unconditionally appends `")"` at the end of every branch, so there's no code path where `SHOW_PLATFORM` data would be silently dropped.
- **The `checksum == 0xFFFFFFFF` "unsynchronized" gate** used in three places (`getName`, `getDescription`, `imxConfigDescription`) is a genuine pre-existing sentinel (set in `ISDevice.h`'s reset/construction paths for both `imxFlashCfg` and `gpxFlashCfg`), not a new convention invented for this PR — correctly reused, and it's also what makes the GPX case safe (a GPX's `imxFlashCfg` is never synchronized, so `SHOW_IO_CONFIG` naturally renders nothing for it without needing a separate hardware-type check).
- The latent `getDescription(const dev_info_t&, int)` link error (declared, never defined) is now implemented, matches its own doc comment's stated scope (name + firmware only, since a bare `dev_info_t` has no port/platform/io-config), and `StaticOverloadsIgnoreTheFlags` passing confirms the static overloads correctly no-op on the new flags rather than doing something wrong with data they don't have.

Nothing else stood out — the deviation-only rendering, per-group granularity (SPI suppressing G6G7, GNSS groups rendering whole), and the pin-detail VERBOSE-only gating all matched their doc comments and passed their dedicated tests. CI is green on both Linux and Windows.

PR #1298, opened 2026-09-05
SN-8626: .dat (LOGTYPE_DAT) reader support in ISLogReader

waltjohnson:
Extends ISLogReader to read .dat segments (cDeviceLogSerial's chunk-header- framed, already-parsed p_data_hdr_t/payload format) alongside .raw, via internal format-branching rather than a subclass -- ISDeviceLog stores ISLogReader by value in a std::vector, and ISLogReader has a private default ctor + non-virtual dtor, so a literal derived reader class doesn't fit. Per D0082 (Kyle's directive): cDeviceLogSerial/DataChunk.h are consulted only for the on-disk byte layout, never as an interface template.

- New SegmentFormat enum + format() accessor on ISLogReader, recognized by extension at construct() time.
- buildIndexFromScanDat(): walks sChunkHeader-framed chunks of p_data_hdr_t/payload pairs directly (no protocol parsing needed, unlike .raw's is_comm_parse_byte scan). Reuses the .idx v2 schema verbatim -- its on-disk layout has no .raw-specific coupling.
- recordEndOffset() branches for .dat: each record is self-delimiting (p_data_hdr_t.size), computed directly rather than inferred from the next record's offset (a .raw chunk-input artifact).
- deriveDeviceIdDat(): unlike .raw, a .dat record's .idx offset IS its p_data_hdr_t start (no wire framing in between), so this walks the already-built records_ instead of re-parsing chunk headers from scratch -- simpler than the .raw path. Extracted the shared filename-fallback logic (parseSerialFromFilenameStem) to avoid duplicating it a third time.
- ISDeviceLog::fromSegments rejects mixed .raw/.dat segments in one composed device log (same pattern as its existing device-id-mismatch check).
- Fixed ISLogger.h's eLogType enum comment, which had LOGTYPE_DAT/ LOGTYPE_RAW's descriptions backwards relative to DeviceLogSerial.h/ DeviceLogRaw.h's authoritative file docs.
- New tests/test_log_reader_dat.cpp (7 tests): open/header/DID-list parity, sidecar rebuild-and-reuse, mid-chunk truncation, multi-segment rollover, and a record-for-record equivalence test against .raw (decodes the same synthetic wire messages into parsed (hdr, payload) pairs, feeds both writers, asserts both readers agree record-for-record on DID + timestamp).

Full existing suite (627 tests) stays green -- zero .raw regressions from the shared-function branching in construct()/recordEndOffset()/ deriveDeviceId().

RawSeriesBuilder's consumer-side follow-up (bytes() for .dat is the raw on-disk payload, not synthesized wire framing -- 4 call sites there plus derivation_helpers.h's walkPayloads<T> need a format-aware branch to skip the comm-reparse) is SN-8627's scope, not this story.

### DIFF
--- a/src/DeviceLogSerial.cpp
+++ b/src/DeviceLogSerial.cpp
@@ -63,13 +63,33 @@ void cDeviceLogSerial::InitDeviceForReading()
 }
 
 bool cDeviceLogSerial::CloseAllFiles() {
-    cDeviceLog::CloseAllFiles();
-
-    // Write remaining data to file
+    // D-119/SN-8626 (mirrors cDeviceLogRaw's SN-8328 fix): flush the buffered
+    // chunk to disk BEFORE the base class writes/finalizes the .idx.
+    // WriteChunkToFile() lazily creates the real segment file via
+    // OpenNewSaveFile(), which is the only place m_fileName is assigned. If the
+    // base's CloseAllFiles() ran first (as this did before), a log small enough
+    // that no chunk was ever flushed during logging would still have an empty
+    // m_fileName at finalize time -- writeIndexChunk()/finalizeIndex() would
+    // land on an orphan "./.idx", and the subsequent lazy OpenNewSaveFile()
+    // would reset the index state and re-emit an empty, non-finalized
+    // <segment>.idx (0 records).
     FlushToFile();
+
+    // Flush any remaining buffered index records and finalize the .idx header
+    // against the now-existing segment file.
+    cDeviceLog::CloseAllFiles();
 
     // Close file
     CloseISLogFile(m_pFile);
+
+    // D-119/SN-8626: this segment is finalized. Reset the physical-offset
+    // accounting so the NEXT segment's records index from 0 -- otherwise
+    // SaveData()'s m_lastIndexOffset computation (below) would use this
+    // now-closed file's stale size for the first record(s) of the next
+    // segment. (The base's lazy OpenNewSaveFile() also zeroes m_fileSize, but
+    // that fires only at the next chunk flush -- too late for records indexed
+    // in between.)
+    m_fileSize = 0;
 
     return true;
 }
@@ -88,6 +108,33 @@ bool cDeviceLogSerial::FlushToFile() {
 
 
 bool cDeviceLogSerial::SaveData(p_data_hdr_t *dataHdr, const uint8_t *dataBuf, protocol_type_t ptype) {
+    // D-119/SN-8626: ensure this record's chunk fits BEFORE indexing it (moved up
+    // from below cDeviceLog::SaveData()'s call, which stamps the .idx offset from
+    // m_lastIndexOffset). Indexing must see the post-flush/post-rotation chunk
+    // state, or a record that triggers a flush gets stamped with the PRE-flush
+    // offset -- wrong by a full chunk. Mirrors cDeviceLogRaw's SN-8328 ordering.
+    int32_t dataBytes = sizeof(p_data_hdr_t) + dataHdr->size;
+    int32_t buffFree = m_chunk.GetBuffFree();
+    if (dataBytes > buffFree) {
+        // Save chunk to file and clear
+        if (!WriteChunkToFile()) {
+            return false;
+        } else if (m_fileSize >= m_maxFileSize) {
+            // Close existing file
+            CloseAllFiles();
+        }
+    }
+
+    // D-119/SN-8626: stamp this record's TRUE physical .dat file offset before
+    // indexing. Unlike .raw (whose on-disk chunks carry no header), .dat's
+    // chunks ARE header-prefixed on disk (WriteChunkToFile's default
+    // writeHeader=true) -- the still-buffered m_chunk will itself be preceded by
+    // a not-yet-written sChunkHeader once flushed, so that header's size counts
+    // toward this record's eventual on-disk position even though it hasn't been
+    // written yet.
+    m_lastIndexOffset = static_cast<uint64_t>(m_fileSize) + sizeof(sChunkHeader)
+                       + static_cast<uint64_t>(m_chunk.GetDataSize());
+
     cDeviceLog::SaveData(dataHdr, dataBuf, ptype);
 
     dev_info_t tmpInfo = {};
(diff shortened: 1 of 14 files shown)


Hephaestus-Maximus (changes requested):
## Review summary

Cloned the branch, traced the core `.dat` parsing/indexing logic by hand, and built + ran the full SDK unit-test suite myself. The design and logic are solid — but this PR breaks the Windows CI build, and the exact same codebase already has an established fix pattern for the reason it breaks.

### Blocking: `test_log_reader_dat.cpp` breaks the Windows build
CI's `General Software Windows` job fails with:
```
test_log_reader_dat.cpp(42): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory
```
The new test file `#include <unistd.h>` for `::getpid()` (line 56, building a unique temp-file hint) — a POSIX-only API with no Windows equivalent. `tests/CMakeLists.txt`'s `WIN32` block already excludes several sibling test files for exactly this reason (`test_device_log.cpp`, `test_log.cpp`, `test_log_index.cpp`, `test_log_reader.cpp`, `test_log_writer.cpp`, `test_truncated_log.cpp` — the comment there literally says "POSIX-only tests: use ::getpid(), /tmp/ paths, or other POSIX APIs that have no Windows equivalent"), but `test_log_reader_dat.cpp` was never added to that list. Fix is a one-line addition:
```cmake
list(FILTER TESTS_SOURCES EXCLUDE REGEX "test_log_reader_dat.cpp")
```
alongside the existing entries in that same `if(WIN32)` block.

### Everything else checks out

**Core parsing logic (`buildIndexFromScanDat`)** — traced the chunk/record validation against `cDataChunk::ReadFromFile`'s actual checks in `DataChunk.cpp`: the `hdr.dataSize != ~hdr.invDataSize` complementary check and `marker != DATA_CHUNK_MARKER` check are exact matches, not approximations. Bounds checks (`total - pos < sizeof(sChunkHeader)`, `chunkDataStart + chunkDataSize > total`, `payloadStart + recHdr.size > bodyEnd`) are all correctly ordered before each read. `p_data_hdr_t::size` is `uint16_t`, so there's no overflow risk in the `size_t` arithmetic even on a corrupted/adversarial file.

**`recordEndOffset()`'s `.dat` branch** — re-reads the record's own header at its stored offset to recompute the end, rather than inferring it from a neighboring record's offset (the `.raw` approach). This is the right call given the comment's explanation that `.raw` offsets are per-arrival-chunk and not naturally ordered — `.dat`'s self-delimiting records don't need that inference and shouldn't inherit its complexity.

**`deriveDeviceIdDat()`** — the redundant `payloadStart + hdr.size > total` bounds check (already validated once by `buildIndexFromScanDat()`) is legitimately defensive, not dead code: `records_` can also come from a trusted on-disk `.idx` sidecar that wasn't re-validated against the current file content this run, so a stale/truncated `.dat` with a stale `.idx` is exactly the case this second check protects against.

**`ISDeviceLog::fromSegments`'s mixed-format rejection** — placed after the existing empty-list guard and device-id-mismatch loop, so `readers.front()` is safely non-empty by the time it runs.

### Test results
Built and ran the full suite myself (not just trusting the PR's reported numbers): **629 tests, 625 passed, 4 pre-existing skips, 0 failed** — includes all 9 new tests (5 in `LogReaderDatTest`, 4 in `LogReaderDat`, covering open/header parity, sidecar rebuild, mid-chunk truncation, multi-segment rollover, and the `.raw`/`.dat` record-for-record equivalence check), all passing individually too. No regressions in the existing suite.

(Aside, unrelated to this PR: I hit a 7+ minute stall in `ComManager.Evb2DataForwardTest` — pre-existing, untouched code — when running it via an overly-broad `--gtest_filter` in isolation. It passed normally as part of the full-suite run, so this looks like a test-order/isolation artifact in that specific test rather than anything to do with this PR; flagging only in case it's a known quirk worth a separate look.)

Once the Windows exclusion is added, this looks ready to merge.

kylemallory (approved):

tmcd249 (approved):

Hephaestus-Maximus (approved):
## Follow-up review

Confirming my previous request-changes concern is resolved, and reviewing the substantial new work added since (writer offset fix, `.dat`-native sync-point scanning, and a new Option A/B self-anchoring feature). Rebuilt from scratch and ran the full suite myself rather than trusting the reported numbers: **647 tests, 643 passed, 4 pre-existing skips, 0 failed**, and CI is green on the current head (`774fd163`).

### Original Windows build break — fixed, and better than my suggestion
I'd suggested adding `test_log_reader_dat.cpp` to the Windows exclusion list. Instead, the fix ports `uniqueTempDir()` to `std::filesystem::temp_directory_path()`, so the test now actually **runs** on Windows CI instead of being skipped there — a better outcome than what I proposed. Verified: no `unistd.h`/`::getpid()` remains in the file, and CI's `General Software Windows` job is green.

### Writer offset bug (`cDeviceLogSerial::SaveData()`/`CloseAllFiles()`)
Traced the new `m_lastIndexOffset = m_fileSize + sizeof(sChunkHeader) + m_chunk.GetDataSize()` computation by hand: it correctly predicts the record's *future* on-disk position — bytes already flushed (`m_fileSize`) + the not-yet-written header of the current in-memory chunk + whatever's already buffered ahead of this record in that chunk. Cross-checked against `cDeviceLogRaw`'s existing `m_rawIndexCursor` pattern (the thing this explicitly mirrors) and confirmed both use the identical "override the generic `m_lastIndexOffset` tally with a real physical cursor right before indexing" convention.

### `.dat`-native sync-point scanning + Option A/B self-anchoring
- `scanSegmentForSyncsDat()` is a clean structural mirror of the `.raw` path's `scanSegmentForSyncs()`, substituting `reader.allRecords()` for the byte-level comm scan — checked field-by-field, including the later "Option A" `DID_SYS_PARAMS` self-anchor block, which I initially found *missing* from an earlier commit's `.dat` path before realizing a later commit (`bf67addf67`) added it to both paths in lockstep. Confirmed the current head has it in both, byte-for-byte identical logic.
- `daysFromCivil()` (Option B's filename-timestamp parser) is Howard Hinnant's public-domain `days_from_civil` algorithm — compared line by line against the canonical reference and it's an exact, unmodified port.
- The regex (`(\d{4})(\d{2})(\d{2})_(\d{2})(\d{2})(\d{2})`) correctly matches cISLogger's own `LOG_SN<n>_YYYYMMDD_HHMMSS_<seq>` naming convention, and the range checks (year/month/day/hour/min/sec) are a reasonable sanity gate for a fallback anchor that's explicitly documented as coarse.

Nothing else to add — this is now a clean, well-tested, and thoroughly cross-checked change. Approving.

PR #1300, opened 2026-09-08
README: document air-gapped / USB-less build option (SN-8286)

Hephaestus-Maximus:
## Summary
- Documents the `ExampleProjects/Minimal_ISDevice` build pattern as the supported path for air-gapped machines or USB-less embedded (UART/serial-only) targets.
- Addresses both blockers reported in SN-8286: the unconditional `libusb`/`libudev` dependency in the default build, and the `pip install`-against-PyPI step in `repositoryInfo.h` generation (neither of which the `Minimal_ISDevice` build path invokes).

## Verification
- Checked out `origin/Airgap` (commit `e887439`) and compiled `ExampleProjects/Minimal_ISDevice` in isolation: `cmake -S . -B build-minimal` configures and `cmake --build build-minimal -j` links cleanly, with no libusb/libudev dependency and no `repositoryInfo.h`/pip step triggered (confirmed `include_is_sdk_find_library.cmake` never calls `update_version_files`).
- Confirmed the example's signal handling (SIGINT/SIGTERM) and raw `test_log` logging additions are present and compile cleanly.
- Have not tested against physical hardware in this environment — build-level verification only.

## Test plan
- [x] Confirm this README addition reads correctly alongside whatever `IS_SDK_ENABLE_USB`-style option (if any) eventually lands for the full/default build, so the docs don't go stale relative to that work.
- [x] Have the SN-8286 reporter (or Novatel) confirm the `Minimal_ISDevice` pattern actually resolves their air-gapped build end-to-end on their SYSROOT.

Ref SN-8286.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,19 @@ The [Inertial Sense software development kit (SDK)](https://github.com/inertials
  * [Software Releases](https://github.com/inertialsense/inertial-sense-sdk/releases) - uINS, uAHRS, uIMU, and EVB-2 firmware and application installers.
  * [SDK & CLTool Source Code](https://github.com/inertialsense/inertial-sense-sdk) - Open source SDK repository with command line tool and example C/C++ source code.
 
+### Building on Air-Gapped or USB-less Systems
+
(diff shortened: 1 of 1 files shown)


shefft (approved):

jbergstrom88 (approved):
The changes to the Readme document are accurate.

Issue #1283, opened 2026-08-26
Cannot connect to device when using symlinked port on Linux


shakeling:
I have an IMX-6 set up to symlink to `/dev/imx6` with this udev rule:

```
ACTION=="add", SUBSYSTEM=="tty", ATTRS{idVendor}=="0483", ATTRS{idProduct}=="5740", ATTRS{product}=="IMX-6", SYMLINK+="imx6"
```

When using the latest commit of the main branch (`879eda29bf36208f69c9e3f33ce388e2c134162b`), I am unable to connect when passing in "/dev/imx6" as the port to the `InertialSense::Open(const char* port, int baudRate, bool disableBroadcastsOnClose, uint16_t filterHdwType)` function like so:

```cpp
InertialSense imu;
imu.Open("/dev/imx6", 921600);
```

When I just do /dev/ttyACM0 it works as expected.
```cpp
InertialSense imu;
imu.Open("/dev/ttyACM0", 921600);
```

I've traced it down to the input "pattern" (`/dev/imx6`) not matching any of the discovered ports in the `./src/PortFactory.cpp` `SerialPortFactor::locatePorts(...)` function.

This worked has worked on older versions of this SDK.

jbergstrom88:
Hi,
1. Do you know the exact SDK version numbers? You mentioned "a few versions back" and updating to the latest.
2. Do you have an error message /log text from the SDK?
3. What is the priority to you?
Thank you

PR #1302, opened 2026-09-09
SN-8575: Fix to resolve symlinked serial port names on Linux

shefft:
Fixes [SN-8575]. Related: inertialsense/inertial-sense-sdk#1283.  Reported by IMX-6 user, but is hardware-independent SDK issue and was reproduced.

`InertialSense::Open()` failed for any serial port reached through a udev alias on Linux — a rule
carrying `SYMLINK+="imx5"` gives `/dev/imx5 -> ttyACM0`, and opening `/dev/imx5` returned `false`
while `/dev/ttyACM0` on the same hardware worked. This resolves such an alias to its canonical
device path before name matching and before the sysfs lookup.

Rebased onto `develop` (`b31483b5`, i.e. after SN-8473's enumeration unification). One commit,
3 files, +304/-3.

## What we found

The report was filed from a source read and had never been run on hardware. It reproduced exactly
as described, and investigation turned up two things the ticket did not have.

**1. There are two independent blockers, not one.** The ticket treated
`validate_port__linux()` as a "separately, would also need" concern. It is a hard second gate,
confirmed failing on its own:

| Gate | Why an alias fails | Reached by |
| --- | --- | --- |
| `SerialPortFactory::locatePorts()` | regex-matches the caller's literal string against names enumerated from `/sys/class/tty`, which are canonical kernel names. An alias can never equal one. | `Open()`, `CorrectionService` |
| `SerialPortFactory::validate_port__linux()` | builds `/sys/class/tty/<basename>` for the driver lookup. `basename("/dev/imx5")` is `imx5`, and `/sys/class/tty/imx5` does not exist. | `bindPort()` |

Fixing only the regex gate would have left `bindPort()` broken — and `bindPort()` is public and
takes a caller-supplied name, which is exactly what `ExampleProjects/Minimal_ISDevice` does with
`argv`. Both gates were measured failing independently on hardware.

Worth noting *why* the second one is easy to miss: `validate_port__linux()`'s first check
(`stat()` + `S_ISCHR`) **passes** for an alias, because `stat()` follows symlinks. Only the sysfs
lookup fails.

**2. `CorrectionService` was a second broken entry point, not mentioned in the ticket.** Its
convenience constructor calls `locatePorts()` with the caller's port name directly and then
`bindPort()`, never touching `Open()`. Verified on hardware: constructing with the alias threw
`std::invalid_argument`, with the canonical name it bound. This is what decided *where* the fix
goes — see below.

## Why this fix

**One new function, `SerialPortFactory::resolvePortName()`, called from the two gates above.**

The alternative of resolving higher up in `InertialSense::OpenPorts()` was ruled out by evidence,
not preference: `CorrectionService` and `bindPort()` both bypass it. `locatePorts()` and
`validate_port__linux()` are where every affected entry point converges.

**Gated on the filesystem, not on the shape of the string.** Callers pass regex *patterns* through
the same argument as literal device paths (`*`, `(.+)`, `/dev/tty(ACM|USB)[0-9]+`). The function
resolves only what `lstat()` reports as an actual symlink, so a pattern — which never names a file —
passes through byte-identical. This is what keeps pattern-based discovery unaffected, and it is
structural rather than heuristic. Everything else is also returned unchanged: canonical paths,
nonexistent names, and dangling links (left for the caller's own existence checks to reject).

**`realpath()` rather than `readlink()`, and hardware drove that choice.** udev writes a *relative*
target — the observed link was `/dev/imx5 -> ttyACM0`, not `-> /dev/ttyACM0`. `readlink()` alone
yields a bare name needing a `/dev/` prefix guessed back onto it, which then breaks on any alias that
*is* absolute, producing `/dev//dev/ttyACM0`. `realpath()` also collapses multi-hop chains and always
returns an absolute path — the form `/sys/class/tty` enumeration produces, and therefore the form we
need to match against.

**On `SerialPortFactory`, not the `PortFactory` base.** `PortFactory` is a pure interface (every
member `virtual` + `= 0` apart from the `onPortAlias()` default), the semantics are
filesystem-specific and only this factory has filesystem names — the others match `host:port` URLs
and mDNS/relay specifiers — and there is no polymorphic call site: everything needing resolution is
already inside `SerialPortFactory`. If TCP hostname or mDNS-instance canonicalization ever becomes a
requirement, promoting this static to a `virtual` with an identity default on the base is a
mechanical refactor. Doing it now would spend an interface decision on speculation and put a new
virtual in front of all five factories plus any third-party ones, which the base class comment
explicitly invites.

**It is `static` (rather than file-local) so the resolution logic is unit-testable in CI without
hardware.** That is what tier 1 below relies on.

## Linux / Windows / other platforms

The whole POSIX body is gated:

```cpp
#if !PLATFORM_IS_LINUX
    return pName;      // no-op
#else
    ... lstat / realpath ...
#endif
```

| Platform | Behavior | Verified |
| --- | --- | --- |
| Linux | resolves aliases | **yes** — hardware + full suite |
| Windows | no-op. COM-port aliasing goes through `QueryDosDevice`, not the filesystem, so there is nothing to resolve. | no Windows host available |
| Apple / embedded | same no-op | not built here |

The guard is deliberately **not** `#if PLATFORM_IS_WINDOWS / #else`, which is how it was first
written. `ISConstants.h` defines three host platforms and `PLATFORM_IS_APPLE` does not imply
`PLATFORM_IS_LINUX`, so "not Windows" would have compiled `lstat`/`realpath`/`PATH_MAX` on Apple and
embedded where those headers are not included — turning an existing weakness into a hard build
error. macOS is not a current target and no work was done to support it; this just avoids *newly*
breaking it. (`validatePort()` and `getComPorts()` in this file are already
`#if WINDOWS / #elif LINUX / #endif` with no `#else`, so they have no return statement on Apple —
a pre-existing condition this PR neither fixes nor worsens.) An in-line comment records why the
guard is shaped this way so it is not "simplified" back later.

`<stdlib.h>`/`<limits.h>` are added to the existing `PLATFORM_IS_LINUX` include guard. They do
currently arrive transitively, but `realpath`'s path in is `<functional>` → `bits/stl_algo.h` →
`<cstdlib>` — a libstdc++ internal that no standard guarantees. This file already includes both
explicitly further down for the same `realpath`/`PATH_MAX` usage, so this follows its own convention.

## How it was tested

**Hardware** — IMX-5.0.6, `SN000525402`, fw `3.1.0-rc.146`, `/dev/ttyACM0` @ 921600, aliased to
`/dev/imx5` by a real udev `SYMLINK+=` rule (not a hand-made link). Every previously failing check
now passes, in the same run, on the same board:

| Check | Before | After |
| --- | --- | --- |
| `locatePorts("/dev/imx5")` | 0 ports | 1 port → `/dev/ttyACM0` |
| `validatePort("/dev/imx5")` | false | true |
| `Open("/dev/imx5", 921600)` | FALSE | TRUE — device identified |
| `CorrectionService("/dev/imx5")` | threw | bound → `/dev/ttyACM0` |

Controls held throughout: the canonical-name open still succeeds, and `open(2)` on the alias still
succeeds — so the failure was never the hardware, the baud rate, or permissions. Ports are still
reported under their canonical names, so enumeration identity is unchanged.

**Unit tests** — 10 new tests in `tests/test_PortFactory.cpp`, three tiers:

1. **Resolution logic, hermetic** (7) — relative link (the form udev writes), absolute link, chained
   link, non-symlink passthrough, dangling link, nonexistent path, and a set of regex patterns
   (`*`, `(.+)`, `.*`, `/dev/tty(ACM|USB)[0-9]+`, comma-list, empty) asserted to come back unmangled.
   Temp-dir symlinks; no device, no root, no network.
2. **Pattern non-regression** (2) — the wildcard reaches `locatePorts()` two ways and both must still
   match everything: verbatim `*` from `OpenPorts()` (an *invalid* regex, handled by the
   `regex_error` fallback) and `.*` from cltool via `utils::globToRegex()`. Plus a matches-nothing
   pattern still yielding nothing.
3. **End-to-end** (1) — alias a real tty, assert both previously failing gates accept it.
   `GTEST_SKIP()` when the host has no serial ports, so it passes on hardware and skips in CI.

| | Baseline (pre-change, same branch) | After |
| --- | --- | --- |
| Tests | 647 | 658 |
| Passed | 643 | 654 |
| Skipped | 4 | 4 (same four, all pre-existing opt-ins) |
| **Failed** | **0** | **0** |

**The tests were confirmed to discriminate.** A test that passes with and without the fix proves
nothing, so this was checked directly: with the two call sites reverted and `resolvePortName()` left
in place, tier 3 failed on **both** assertions (`validatePort` false, `locatePorts` 0 ports — the
original bug exactly) while tiers 1 and 2 passed. The tests fail for the right reason and pass for
the right reason.

## Known limitation (documented, not fixed)

Two conventions reach `locatePorts()`: `OpenPorts()` forwards the caller's string verbatim (the path
this ticket reports — handled), while cltool's `-ufpkg` re-discovery loop passes it through
`utils::globToRegex()` first, which escapes regex metacharacters. An alias containing one —
`/dev/imx5.0` arriving as `/dev/imx5\.0` — is no longer a real path, so the `lstat()` gate declines
it. Effect: during a cltool firmware-update run against a dotted alias, the periodic re-discovery
would not re-find the port.

Not fixed here on purpose. Un-escaping a regex back into a path is guesswork, and the real issue is
that these two paths disagree about whether a port specifier is a literal or a pattern. That wants
addressing on its own terms rather than being papered over inside `resolvePortName()`. Recorded as an
in-line `KNOWN LIMITATION` comment at the call site.

## Out of scope / follow-ups

- **The glob-vs-literal port specifier inconsistency** described under Known limitation above. Two
  call paths disagree about whether a port specifier is a literal path or a regex pattern, which is
  the root of that limitation and the only thing blocking metacharacter-bearing aliases.
- **Re-`Open()` in one process returns `TRUE` while `IsOpen()` reports `false`** and the device
  renders as `(Closed)`. Confirmed independent of this change by pointing both opens at the same
  path (first `IsOpen=true`, second `false`) — a singleton `PortManager` state artifact, visible now
  only because the first open no longer fails. Returning success for a port that is not open looks
  like a genuine inconsistency worth its own ticket.
- **IMX-6 is unverified.** The original report is IMX-6; only IMX-5 was available. The fix is
  entirely host-side path handling and cannot see the device — and after resolution the alias case
  becomes the canonical case, which the reporter states already works on their IMX-6. Their udev rule
  also uses the same `0483:5740` VID:PID as the IMX-5 tested here. Confirmation on IMX-6 would be
  good to have on the record.

## Notes for reviewers

Reproducing the original bug needs a udev alias, and two details in the ticket's recipe will waste
your time:

- Its rule matches `ATTRS{product}=="IMX-6"`. The IMX-5 here reports `"STM32 Virtual ComPort"` with
  the same VID:PID, so that rule matches nothing on this unit. Read the attributes off the device
  with `udevadm info -a -n /dev/ttyACM0` instead of hardcoding a product match.
- The rule is `ACTION=="add"`, but `udevadm trigger` defaults to `--action=change` and so silently
  does nothing. Use `udevadm trigger --action=add`, or replug the board.
  `sudo udevadm test $(udevadm info -q path -n /dev/ttyACM0) 2>&1 | grep -i symlink` distinguishes
  "rule is wrong" from "event never delivered".

Tier 3 of the new tests covers the same ground in-tree without any of that, using whatever tty the
host has.

[SN-8575]: https://inertialsense.atlassian.net/browse/SN-8575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/PortFactory.h
+++ b/src/PortFactory.h
@@ -126,6 +126,25 @@ public:
     bool validatePort(const std::string& pName, uint16_t pType = 0) override;
 
     /**
+     * Resolves a caller-supplied port name to the canonical device path the OS enumerates it under.
+     *
+     * On Linux a serial port is commonly reached through a udev alias -- a rule carrying
+     * SYMLINK+="imx5" yields /dev/imx5 -> ttyACM0 -- but port discovery enumerates canonical kernel
+     * names from /sys/class/tty, and validation keys its sysfs lookup on that same canonical name.
+     * Either one, given an alias, rejects a device that is otherwise perfectly usable (SN-8575).
+     *
+     * Only an existing symlink is resolved. Everything else is returned unchanged: a canonical path,
+     * a name that does not exist, a dangling link, and -- importantly -- a regex pattern such as
+     * "*", "(.+)" or "/dev/tty(ACM|USB)[0-9]+", none of which name a file on disk. Pattern-based
+     * discovery is therefore unaffected. No-op on Windows, where COM-port aliasing is handled by
+     * QueryDosDevice rather than the filesystem.
+     *
+     * @param pName the port name, path, or pattern as supplied by the caller
+     * @return the canonical absolute device path if @p pName is a resolvable symlink, else @p pName
+     */
+    static std::string resolvePortName(const std::string& pName);
+
+    /**
      * Allocates and initializes a serial_port_t for the given port name. Does NOT open the device.
      * @param pName the serial device name/path to bind
      * @param pType additional port-type flags OR'd with PORT_TYPE__UART | PORT_TYPE__COMM
--- a/src/PortFactory.cpp
+++ b/src/PortFactory.cpp
@@ -108,10 +108,10 @@ std::string SerialPortFactory::resolvePortName(const std::string& pName) {
 
     // realpath() rather than readlink(): udev writes a RELATIVE target (a SYMLINK+="imx5" rule
     // produces /dev/imx5 -> ttyACM0, not -> /dev/ttyACM0), so readlink() alone yields a bare name
-    // that needs a "/dev/" prefix guessed back onto it -- which is how the equivalent block in
-    // InertialSense::BootloadFile() ends up producing "/dev//dev/ttyACM0" for an absolute target.
-    // realpath() also collapses multi-hop chains and always returns an absolute path, which is the
-    // form /sys/class/tty enumeration produces and therefore the form we need to match against.
+    // needing a "/dev/" prefix guessed back onto it -- which then breaks on any alias that IS
+    // absolute, producing "/dev//dev/ttyACM0". realpath() also collapses multi-hop chains and
+    // always returns an absolute path, which is the form /sys/class/tty enumeration produces and
+    // therefore the form we need to match against.
     char resolved[PATH_MAX] = {};
     if (realpath(pName.c_str(), resolved) == nullptr)
         return pName;   // dangling link; leave it for the caller's own existence checks to reject
--- a/src/PortFactory.cpp
+++ b/src/PortFactory.cpp
@@ -15,6 +15,8 @@
 #if PLATFORM_IS_LINUX
 #   include <dirent.h>
 #   include <sys/stat.h>
+#   include <stdlib.h>      // realpath()
+#   include <limits.h>      // PATH_MAX
 #endif
 
 // #define REMOTE_SOCAT_PORTS      // only does anything on linux
@@ -86,7 +88,54 @@ bool SerialPortFactory::validatePort(const std::string& pName, uint16_t pType) {
 #endif
 }
 
+std::string SerialPortFactory::resolvePortName(const std::string& pName) {
+#if !PLATFORM_IS_LINUX
+    // Windows: COM-port aliasing goes through QueryDosDevice, not the filesystem, so there is
+    // nothing here to resolve. Every other platform (Apple, embedded) falls through to the same
+    // no-op -- symlink resolution is only wired up for Linux, which is where SN-8575 was reported
+    // and the only host platform this is verified on. Note this is deliberately NOT "#if WINDOWS /
+    // #else": the POSIX branch below needs headers that are only included for Linux, so treating
+    // "not Windows" as "has a filesystem" would break the build everywhere else.
+    return pName;
+#else
+    // Resolve only something that is actually a symlink on disk. Callers reach here with regex
+    // PATTERNS as well as literal device paths ("*", "(.+)", "/dev/tty(ACM|USB)[0-9]+"); none of
+    // those name a file, so lstat() fails and they take this early return untouched. Gating on the
+    // filesystem rather than on the shape of the string is what keeps pattern matching unaffected.
+    struct stat st = {};
+    if (lstat(pName.c_str(), &st) != 0 || !S_ISLNK(st.st_mode))
+        return pName;
+
+    // realpath() rather than readlink(): udev writes a RELATIVE target (a SYMLINK+="imx5" rule
+    // produces /dev/imx5 -> ttyACM0, not -> /dev/ttyACM0), so readlink() alone yields a bare name
+    // that needs a "/dev/" prefix guessed back onto it -- which is how the equivalent block in
+    // InertialSense::BootloadFile() ends up producing "/dev//dev/ttyACM0" for an absolute target.
+    // realpath() also collapses multi-hop chains and always returns an absolute path, which is the
+    // form /sys/class/tty enumeration produces and therefore the form we need to match against.
+    char resolved[PATH_MAX] = {};
+    if (realpath(pName.c_str(), resolved) == nullptr)
+        return pName;   // dangling link; leave it for the caller's own existence checks to reject
+
+    return std::string(resolved);
+#endif
+}
+
 void SerialPortFactory::locatePorts(std::function<void(PortFactory*, uint16_t, std::string)> portCallback, const std::string& pattern, uint16_t pType) {
+    // A caller-supplied path may be a udev alias (/dev/imx5 -> ttyACM0) while getComPorts() below
+    // enumerates canonical /sys/class/tty names only, so an unresolved alias matches nothing no
+    // matter how the regex is written. Resolve before matching. Non-symlinks, patterns included,
+    // pass through untouched. (SN-8575)
+    //
+    // KNOWN LIMITATION: two conventions reach this function. OpenPorts() forwards the caller's
+    // string verbatim (the path SN-8575 reports, and the one this handles), but cltool's -ufpkg
+    // re-discovery loop passes it through utils::globToRegex() first, which escapes regex
+    // metacharacters. An alias containing one -- "/dev/imx5.0" arriving here as "/dev/imx5\.0" --
+    // is no longer a real path, so the lstat() gate above declines to resolve it. Un-escaping a
+    // regex back into a path would be guesswork; the underlying problem is that the two paths
+    // disagree about whether a port specifier is a literal or a pattern, which wants fixing on its
+    // own terms rather than being papered over here.
+    const std::string resolvedPattern = resolvePortName(pattern);
+
     // An unusable pattern must not abort a port scan by throwing out of it. std::regex's constructor
     // throws std::regex_error on any invalid expression, and callers reach this with strings they think
     // of as port SPECIFIERS rather than regexes -- cltool's default "*" (its all-ports token) is a valid
@@ -96,7 +145,7 @@ void SerialPortFactory::locatePorts(std::function<void(PortFactory*, uint16_t, s
     // so loudly enough to be fixed at the call site.
     std::regex matchPattern;
     try {
-        matchPattern.assign(pattern);
+        matchPattern.assign(resolvedPattern);
     } catch (const std::regex_error& e) {
         log_error(IS_LOG_PORT_FACTORY, "locatePorts(): pattern '%s' is not a valid regular expression (%s); "
                                        "matching all ports instead.", pattern.c_str(), e.what());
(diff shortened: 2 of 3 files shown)


Hephaestus-Maximus (approved):
## Review — verified independently, not just read

Cloned the branch, built the SDK from scratch (`InertialSenseSDK` static lib, clean build, no warnings), built `tests/build` against it, and ran the full suite plus the new tests specifically:

```
[==========] 658 tests from 74 test suites ran. (59399 ms total)
[  PASSED  ] 653 tests.
[  SKIPPED ] 5 tests  (test_PortFactory.symlinkedPort_isValidatedAndLocated + 4 pre-existing hardware/fixture opt-ins)
```

All 12 runnable `test_PortFactory.*` tests pass. `symlinkedPort_isValidatedAndLocated` (tier 3) skips in this container because there's no real tty here — exactly the documented `GTEST_SKIP()` behavior, not a failure.

### Code review against the PR's own claims
Went through `PortFactory.cpp`/`.h` line by line and checked the specific technical claims in the description rather than taking them at face value:

- **The two-gate claim holds up.** `locatePorts()` resolves `pattern` before `matchPattern.assign(...)`, and `validate_port__linux()` resolves `pName` before building `/sys/class/tty/%s` from `basename()`. Both are real, independent call sites — fixing only one would leave `bindPort()` (which reaches `validate_port__linux()` directly, bypassing `locatePorts()`) still broken for an alias. Confirmed by reading the call graph, not assumed.
- **`realpath()` vs `readlink()` reasoning is correct.** `readlink()` on a relative udev target (`/dev/imx5 -> ttyACM0`) returns the bare relative string; `InertialSense.cpp`'s existing `BootloadFile()` block does exactly this and produces `/dev//dev/ttyACM0` on an absolute target — I read that code to confirm the claimed pre-existing defect is real and that this PR doesn't touch/reuse that path.
- **The `#if !PLATFORM_IS_LINUX / #else` guard shape is deliberately correct**, not just stylistically chosen: `ISConstants.h` defines `PLATFORM_IS_WINDOWS`/`PLATFORM_IS_LINUX`/`PLATFORM_IS_APPLE` as independent flags, so `#if PLATFORM_IS_WINDOWS / #else` would compile `lstat`/`realpath`/`PATH_MAX` on Apple/embedded targets that don't include those headers. The chosen guard avoids that. Checked `ISConstants.h` directly rather than trusting the description.
- **`static` on `SerialPortFactory` rather than a new virtual on `PortFactory`** is reasonable — I checked `PortFactory.h` and confirmed it's a pure interface with no other implementor needing filesystem semantics (TCP/relay/mDNS factories use URL-shaped names, not paths).
- **The `globToRegex()` known-limitation claim is accurate** — `test_utils.globToRegex_escapes_dot` (pre-existing test, still passing) confirms `globToRegex()` escapes `.`, which is exactly the mechanism the documented KNOWN LIMITATION comment describes for the cltool `-ufpkg` re-discovery path.
- Pattern-passthrough safety (the main regression risk — patterns riding through the same argument as literal paths) is covered hermetically in tier 1 (`resolvePortName_leavesRegexPatternsUnchanged`) and I confirmed it: `lstat()` on a regex string like `(.+)` or `/dev/tty(ACM|USB)[0-9]+` fails (ENOENT, not a valid path), so the early-return-unchanged path is taken. No shape-based string sniffing involved.
- `PATH_MAX`-sized stack buffer for `realpath()` output is the standard, correct usage pattern (no overflow risk); `lstat` before `realpath` correctly filters out non-symlinks and dangling links without the function needing to distinguish those cases in the `realpath` call itself.

### What I could not verify
- The hardware-specific claims (IMX-5 board, real udev `SYMLINK+=` rule, before/after `Open()`/`CorrectionService` behavior) — no hardware available in this environment. The hermetic tier-1/tier-2 tests substitute reasonably well since they exercise the exact same `resolvePortName()` logic path with real filesystem symlinks, which is what I ran and confirmed above.
- CI (`Continuous Integration (SDK)`) is still `in_progress` on the head commit as of this review.

### Nits (non-blocking)
- None worth raising — this is unusually well-documented and the documentation matches the code exactly everywhere I checked.

Approving. Clean build, full test suite green, and every specific technical claim in the description checked out against the actual code rather than assumed.

tmcd249 (approved):

src/PortFactory.h
```diff
@@ -125,6 +125,25 @@ class SerialPortFactory : public PortFactory {
      */
     bool validatePort(const std::string& pName, uint16_t pType = 0) override;
 
+    /**
```
kylemallory, line 128: Minor: This should probably be a private/internal function.

kylemallory (approved):

waltjohnson (approved):

src/PortFactory.h
shefft: Deliberate, but happy to change it. It's public so the resolution logic is unit-testable without a device attached — 7 of the 10 new tests call it directly, covering relative/absolute/chained/dangling links and asserting regex patterns pass through unmodified. Private, those all go away and the only coverage left is the tier-3 test, which skips in CI, so the part most at risk of regressing becomes untestable there.

Precedent: getComPorts() in this same class is public for the same reason as of SN-8473, and test_PortFactory.cpp calls it directly.
 If keeping it off SerialPortFactory's surface matters, I'd suggest moving it to utils:: in util/util.h rather than making it private — it's a generic path helper with no serial-port knowledge, it sits naturally beside globToRegex(), and it stays testable. Say which you prefer and I'll make the change.

shefft: Has been decided not to move it into utils.  Let's merge since it is approved, and we'll go back and have Claude look into making it private and adjusting the tests as needed, for a possible later improvement.

PR #1306, opened 2026-09-11
Loginspector fix for the heading residual plot crash

aabogdanov:
Fixes the bug that causes logInspectoir crash

### DIFF
--- a/python/inertialsense/logInspector/logPlotter.py
+++ b/python/inertialsense/logInspector/logPlotter.py
@@ -1277,6 +1277,8 @@ class logPlot:
     def angle_unwrap(self, angle):
         unwrap = 0.0
         result = np.empty_like(angle)
+        if np.shape(angle)[0] == 0:
+            return result
         anglePrev = angle[0]
         for i in range(np.shape(angle)[0]):
             result[i] = angle[i] + unwrap
@@ -1428,7 +1430,8 @@ class logPlot:
                         delta = self.vec3_wrap(intEuler - refEuler)
                         sumDelta += delta
                         sumCount += 1
-                refEuler += sumDelta / sumCount
+                if sumDelta is not None:
+                    refEuler += sumDelta / sumCount
 
         for d in self.active_devs:
             qn2b = self.getData(d, DID_INS_2, 'qn2b')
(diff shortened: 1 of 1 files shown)


kylemallory (approved):

Hephaestus-Maximus (approved):
## Review — reproduced the actual crash on the base branch before trusting the fix

The PR description is thin ("Fixes the bug that causes logInspectoir crash"), so rather than take that on faith I reproduced the crash mechanism directly against the pre-fix code.

### Root cause, demonstrated
Extracted `angle_unwrap()` and the RTK-heading reference-computation block from `heading()` verbatim from `origin/3.1.0-rc` (unmodified) and ran them against a synthetic 2-device scenario: device A has real RTK-compassing data, device B has none (e.g. a unit without an RTK antenna baseline, or just not fixed yet in this log — a completely ordinary mixed-fleet case). Result:

```
IndexError: index 0 is out of bounds for axis 0 with size 0
```

Mechanism: in the old code, once the *first* device sets `refRtkTime`/`refRtkHdg`, every *subsequent* device unconditionally calls `self.angle_unwrap(gnssHdg)` on its own data with no emptiness check. If that device has no RTK data, `gnssHdg` is a zero-length array, and `angle_unwrap()`'s `anglePrev = angle[0]` indexes into it and raises. (If the *first* device happens to be the one lacking data, the old code doesn't crash immediately — it silently sets the whole reference to an empty array instead, which is its own latent problem but not what's being fixed here.)

### The fix, verified against the same reproduction
Ran the same synthetic scenario through this PR's patched code: `heading()`'s RTK block now `continue`s past any device with empty `gnssTime`/`gnssHdg` before ever calling `angle_unwrap()` on it, and `angle_unwrap()` itself gained a `if np.shape(angle)[0] == 0: return result` guard as defense in depth. Re-ran the identical repro — no crash, `refRtkTime`/`refRtkHdg` correctly computed from just the device that has data.

### Checked the rest of the diff for consistency
- The INS-reference block (`refInsTime`/`refEuler`) gets the exact same treatment (early `continue` on `len(time)==0 or len(qn2b)==0`, paired with `if sumDelta is not None:` before the final `+=`) — same bug class, same fix shape, correctly paired (the early-continue and the final-None-guard have to travel together, since skipping every device is exactly what makes `sumDelta` stay `None` under the new code, whereas the old code could never leave it `None` as long as at least one device existed — I checked this pairing is symmetric between the two blocks, not just copy-pasted incompletely).
- The added `and refRtkHdg is not None` on the per-device residual-plotting `if gnssTime.any():` check is correct but likely redundant given the first loop guarantees `refRtkHdg` is set whenever any device's `gnssTime.any()` is true — harmless defensive addition, not a concern.
- The pre-existing `if not (refInsTime is None) and ...:` guard for the INS residual column (third plot) was already there before this PR and is unaffected.
- Confirmed `python -m py_compile` passes on the patched file (no syntax issues), and CI (`Continuous Integration (SDK)`) is green on the current head.

### Verdict
Approving. Small, correctly-scoped fix; I reproduced the actual crash on the unpatched base branch and confirmed this diff resolves it rather than trusting the description.

shefft (approved):

PR #1318, opened 2026-09-20
SN-8697: broaden Windows serial device-loss detection past a two-code allowlist

Hephaestus-Maximus:
## Summary

Fixes the mechanism behind SN-8697 (Windows EvalTool doesn't drop/rediscover an IMX-6's COM port after an FPKG reboot, so the firmware-update dialog hangs; Linux is unaffected).

- Adds `win32ErrorIndicatesDeviceLost()` — a broadened, standard set of Win32 "device gone" error codes (`ERROR_GEN_FAILURE`, `ERROR_DEVICE_NOT_CONNECTED`, `ERROR_OPERATION_ABORTED`, `ERROR_SEM_TIMEOUT`, `ERROR_INVALID_HANDLE`, plus the two pre-existing codes), replacing the previous two-code allowlist in `serialPortWritePlatform()`'s Windows branch — that list was self-acknowledged as incomplete in its own comment.
- Fixes an `errno`/`GetLastError()` mismatch: the Windows write path was storing `errno` (a CRT global, meaningless on Windows) into `serialPort->errorCode` instead of the real Win32 error it had already fetched.
- Fixes `serialPortReadTimeoutPlatformWindows()`, which previously had **no error-propagation path at all** — every `ReadFile()`/`GetOverlappedResult()` hard failure was silently discarded. It now records the real error and returns `-1`, matching the negative-on-failure contract `serialPortReadTimeoutPlatformLinux()` already uses; the generic wrapper is adjusted per-platform so it doesn't immediately clobber that with a bogus `errno` read.

## Why this explains the symptom

Traced through the code, not reproduced on hardware (see Test plan): a rebooting IMX-6/GPX-1 doesn't re-enumerate on Windows, so the COM port's OS node persists throughout the reboot (`ISFirmwareUpdater.cpp` already has a comment about this for the legacy ISbl path). For the FPKG/ISv2 path, nothing calls `disconnect(true)` around a reset — `portInvalidate()` is only ever reached via a live I/O operation failing. Between the write path's narrow 2-code allowlist and the read path's total silence, a typical Windows device-loss failure during that reboot window had no path to actually trip port invalidation, so the stale device/port was never rediscovered. Linux doesn't have this problem — its error propagation is broad (multiple `errno` values, both directions) rather than a short allowlist.

## Test plan
- [x] `InertialSenseSDK` and every host tool that links it (`ci_hdw_tests`) rebuild clean on Linux.
- [x] Full `IS-SDK_unit-tests` suite: 664/669 pass, 5 pre-existing environment-gated skips, 0 failures.
- [x] Confirmed the Linux branch of every touched function is byte-identical to before — all new logic is behind `#if PLATFORM_IS_WINDOWS`.
- [ ] **Not yet verified: real Windows + IMX-6 + GPX-1 hardware test, performing a full forced IMX+GPX upgrade** (exactly what the ticket calls for). I have no Windows build toolchain or runtime available to me. @kylemallory mentioned having a Windows VM with hardware — this needs that validation before merge.

### DIFF
--- a/ExampleProjects/Bootloader/ISBootloaderExample.cpp
+++ b/ExampleProjects/Bootloader/ISBootloaderExample.cpp
@@ -21,7 +21,7 @@ THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 #include "../../src/serialPortPlatform.h"
 #include "../../src/ISBootloaderThread.h"
 #include "../../src/ISBootloaderBase.h"
-#include "../../src/ISSerialPort.h"
+#include "../../src/PortFactory.h"
 
 using namespace ISBootloader;
 using namespace std;
@@ -116,7 +116,7 @@ int main(int argc, char* argv[])
 
     // For now, we will use all present devices.
     std::vector<std::string> portStrings;
-    cISSerialPort::GetComPorts(portStrings);
+    SerialPortFactory::getComPorts(portStrings);
 
     // Set all files the same, the bootloader logic will identify the file and only put it onto the appropriate devices.
     firmwares_t files;
@@ -132,7 +132,7 @@ int main(int argc, char* argv[])
     vector<string> all_ports;                   // List of ports connected
 
     // For now, we will use all present devices.
-    cISSerialPort::GetComPorts(all_ports);
+    SerialPortFactory::getComPorts(all_ports);
 
     // Update the firmware on any port that was open
     std::vector<cISBootloaderThread::confirm_bootload_t> confirm_device_list;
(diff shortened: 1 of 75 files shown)


shefft (approved):

Hephaestus-Maximus:
Rebuilt this branch directly off \`3.1.1-rc\` (cherry-picked just the fix commit, force-pushed) -- the original branch was rooted in \`develop\`, so retargeting the base alone was pulling in ~40 unrelated \`develop\` commits into the diff. This now shows a clean single-commit, single-file diff against \`3.1.1-rc\`.

Re-verified after the rebuild: SDK library rebuilds clean, full \`IS-SDK_unit-tests\` suite passes (589/593, 4 pre-existing environment-gated skips, 0 failures) against this older base.

https://github.com/inertialsense/inertial-sense-sdk/pull/1318

aabogdanov (approved):

kylemallory (approved):

tmcd249 (approved):

